PR #1, opened 2016-11-22
Updated Python samples to use MQTT

Ilias-Tsigkogiannis:


zolvarga:
HI Ilias,

We already changed the default values to MQTT. Please take a look at our latest release:
https://github.com/Azure/azure-iot-sdk-python/releases

Thanks,
Zoltan

PR #12, opened 2017-01-27
Set message properties

briglx:
The sample code has the option to set a message property but it is never assigned to the message. This code will assign the prop_map to the message properties.


zolvarga:
Hi briglx,

Thank you for your contribution!

Zoltan

PR #47, opened 2017-06-23
Updated Linux and ARM release tags

digimaun:
# Checklist
- [X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-python/blob/master/.github/CONTRIBUTING.md).
_I have rerun the release scripts. No tests failed._

- I submitted this PR against the correct branch: _I am not sure what the current dev branch is!_
  - [] This pull-request is submitted against the `master` branch. 

# Description of the problem
Existing platform architecture tags are not compliant with PEP513. Also there is no specifier for ARM like for Raspberry Pi.

# Description of the solution
Small tweak to the Release.sh script with correct identifiers as specified by
https://www.python.org/dev/peps/pep-0513/  (for Linux)

and future state (for arm)
https://github.com/pypa/warehouse/issues/2003

Presently Pypi accepts the manylinux1_....


zolvarga:
Hi Paymaun,

Please remove the service/src/iothub_service_client_python.cpp changes from the PR.
Our release process will maintain the version number of the Python SDK. We should not change it directly.

Thanks,
Zoltan

digimaun:
Might there be a better branch to submit a pull request against? Like a current dev branch? The .cpp version change was actually from a different author - not from the changes I would like to submit. 

Thank You,
Paymaun

PR #52, opened 2017-07-12
Release.sh changes for .whl tags to support breadth of Linux platforms

digimaun:
# Checklist
- [X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-python/blob/master/.github/CONTRIBUTING.md).
  - [X] This pull-request is submitted against the `master` branch. 

# Description of the problem
Existing platform architecture tags are not compliant with PEP513. Also there added specifier for ARMv7l like for Raspberry Pi 3.

# Description of the solution
Small tweak to the Release.sh script with correct identifiers as specified by
https://www.python.org/dev/peps/pep-0513/ (for Linux)

and future state (for arm)
pypa/warehouse#2003


zolvarga:
@digimaun 
Thanks for contributing!

PR #55, opened 2017-07-17
Fix Python library install

markrad:
# Checklist
- [x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-python/blob/master/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down). _No additional test requried_
  - [x] This pull-request is submitted against the `master` branch. 

# Description of the problem
Currently the build system does not enable the installation of the two Python shared objects: iothub_client.so and iothub_service_client.so. 

# Description of the solution
Modified to two cmake files so that when a customer uses _make install_ the two shared objects will be installed into the standard Linux location (or a user specified location if CMAKE_INSTALL_PREFIX is used).

### DIFF
--- a/device/iothub_client_python/src/CMakeLists.txt
+++ b/device/iothub_client_python/src/CMakeLists.txt
@@ -78,3 +78,14 @@ endif()
 
 
 linkSharedUtil(iothub_client_python)
+
+
+set(install_libs
+	iothub_client_python
+)
+
+install(TARGETS ${install_libs} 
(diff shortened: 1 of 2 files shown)


zolvarga:
:shipit:

zolvarga:
@markrad 

I signed of on the PR.

Thanks,
Zoltan

PR #172, opened 2018-08-14
Create Python 3 release packages

zolvarga:
# Checklist
- [ ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-python/blob/master/.github/CONTRIBUTING.md).
  - [ ] This pull-request is submitted against the `master` branch. 
  - [ ] I have merged the latest `master` branch prior to submission and re-merged as needed after I took any feedback.
  - [ ] I have squashed my changes into one with a clear description of the change.

  
# Description of the problem

# Description of the solution

### DIFF
--- a/build_all/mac/release.sh
+++ b/build_all/mac/release.sh
@@ -91,22 +91,35 @@ build_c_sdk()
 
 copy_binaries()
 {
-    [ $? -eq 0 ] || exit $?
     cd $build_root
 
     echo copy iothub_client library to samples folder
     cp $build_folder/python/src/iothub_client.dylib ./device/samples/iothub_client.so
+    [ $? -eq 0 ] || exit $?
+
     echo copy iothub_client_mock library to tests folder
     cp $build_folder/python/test/iothub_client_mock.dylib ./device/tests/iothub_client_mock.so
+    [ $? -eq 0 ] || exit $?
+
     cp $build_folder/python/src/iothub_client.dylib ./device/tests/iothub_client.so
+    [ $? -eq 0 ] || exit $?
+
     cp $build_folder/python_service_client/src/iothub_service_client.dylib ./device/tests/iothub_service_client.so
+    [ $? -eq 0 ] || exit $?
 
     echo copy iothub_service_client library to samples folder
     cp $build_folder/python_service_client/src/iothub_service_client.dylib ./service/samples/iothub_service_client.so
+    [ $? -eq 0 ] || exit $?
+
     echo copy iothub_service_client_mock library to tests folder
     cp $build_folder/python_service_client/tests/iothub_service_client_mock.dylib ./service/tests/iothub_service_client_mock.so
+    [ $? -eq 0 ] || exit $?
+
     cp $build_folder/python_service_client/src/iothub_service_client.dylib ./service/tests/iothub_service_client.so
+    [ $? -eq 0 ] || exit $?
+
     cp $build_folder/python/src/iothub_client.dylib ./service/tests/iothub_client.so
+    [ $? -eq 0 ] || exit $?
 }
 
 run_unit_tests()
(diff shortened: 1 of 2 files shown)


build_all/mac/release.sh
```diff
@@ -7,9 +7,15 @@ build_folder=$build_root"/c/cmake/iotsdk_mac"
...
 
-PYTHON_VERSION=2.7
+# build with Python 2 and Python 3 by default 
+PYTHON_VERSION=all
+
+# don't upload by default
 UPLOAD_PIP=0
 
+# processor architecture is any for Mac packages
+PLAT_ARCH="any"
```
pierreca: i'm still concerned by this. i'm afraid this wheel is going to be installable on systems that aren't macos and aren't covered by our other wheels. 
[This page](https://github.com/MacPython/wiki/wiki/Spinning-wheels) which is linked [from the official python docs](https://packaging.python.org/guides/packaging-binary-extensions/#binary-extensions-for-macos) suggests the platform tag to be macosx_<sdk_version>_intel (the page says it can be obtained by running`python -c "import distutils.util; print(distutils.util.get_platform())"`)

build_all/mac/release.sh
```diff
@@ -45,96 +51,138 @@ process_args()
...
+    echo "python${PYTHON_VERSION}" iothub_service_client_ut.py
+    "python${PYTHON_VERSION}" iothub_service_client_ut.py
+    [ $? -eq 0 ] || exit $?
+    echo "python${PYTHON_VERSION}" iothub_service_client_map_test.py
+    "python${PYTHON_VERSION}" iothub_service_client_map_test.py
+    [ $? -eq 0 ] || exit $?
+    cd $build_root
+}
 
+create_packages()
```
pierreca: this needs error checking

build_all/mac/release.sh
```diff
@@ -45,96 +51,138 @@ process_args()
...
+
+    cd ./build_all/mac/release_service_client
+    cp $build_folder/python_service_client/src/iothub_service_client.dylib iothub_service_client/iothub_service_client.so
+    "python${PYTHON_VERSION}" setup.py bdist_wheel --plat-name $PLAT_ARCH
+    cd $build_root
+}
 
-[ $? -eq 0 ] || exit $?
-cd $build_root
+upload_packages()
```
pierreca: same here, needs error checking

build_all/mac/release.sh
```diff
@@ -45,96 +51,138 @@ process_args()
...
+
+    make --jobs=$CORES
+    ctest -j $CORES -C "Debug" --output-on-failure
+
+    popd
+}
 
-make --jobs=$CORES
-ctest -j $CORES -C "Debug" --output-on-failure
+copy_binaries()
```
pierreca: needs error checking

build_all/mac/release.sh
zolvarga: There a 2 operations in this function (copy and bdist_wheel) and both will output errors if any.
When we manually run the release script we always have to check the output for errors.

build_all/mac/release.sh
zolvarga: When we manually run the release script we always have to check the output for errors. 
The upload will output an error if any. We don't want tot go to the 3rd party toll's output parsing business.

build_all/mac/release.sh
zolvarga: I don't agree but respect your design choice. I will update the PR.

build_all/mac/release.sh
zolvarga: When we manually run the release script we always have to check the output for errors.

pierreca: just to clarify i'm asking for more of this: `[ $? -eq 0 ] || exit $?`
it's already used in some parts of the scripts, but not all of them - why the inconsistency?
should pretty much be everywhere, right?

having a human check console output is error prone.

zolvarga: Added error handling

Issue #230, opened 2018-11-07
No to_python (by-value) converter found for C++ type: void*


k2nsloos:
Hi,

I noticed that the azure-iot-python-sdk knocks it self out whenever I receive an incoming device method. The callback is set by set_device_method_callback_ex(), instead of the regular version.

 I intended to register a callback, which allowed to respond to the incoming call asynchronously. That is not returning the response in  the same call, but using device_method_response() member function, to set the actual response at some point in the future.

However boost tosses a TypeError: No to_python (by-value) converter found for C++ type: void*
I'm not sure how to fix this.

Kind regards,
Jaap

k2nsloos:
The issue arises on line 1228 of  in device/iothub_client_python/src/iothub_client_python.cpp

Where you pass the void pointer to the python routine.

boost::python::object user_response_obj = deviceMethodCallback(method_name_std_string, payload_std_string, method_id, userContext);

pierreca:
I will - please provide a sample code in python that shows the issue as well as the console log, and versions of the OS and Python interpreter used. thanks!

k2nsloos:
The hw is a raspberry pi 1b+ (armhf / armv6l), running rasbian and python 2.7.
I'll try to create an python example script today.

k2nsloos:
[example.zip](https://github.com/Azure/azure-iot-sdk-python/files/2571513/example.zip)

If I run the attached python file, and invoke a direct method using the HTTPS Api, then I get the following log output. The example requires twisted to run. (That's in fact what I want: integrate the client with existing code written for the twisted framework.)

`
2018-11-12T10:19:16+0000 [__main__#info] Using connection string HostName=testjaap.azure-devices.net;DeviceId=iot_test_device;SharedAccessKey=MF72+cjhmOo2orIgRcC/yOgwJRe+ZVrw/C233Vl6i18=
-> 10:19:16 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: testjaap.azure-devices.net/iot_test_device/?api-version=2017-11-08-preview&DeviceClientType=iothubclient%2f1.2.9%20(native%3b%20Linux%3b%20armv7l) | PWD: XXXX | CLEAN: 0
<- 10:19:16 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
[...]

PR #234, opened 2018-11-15
Fix for #230 - method callbacks

pierreca:
The method_id parameter wasn't properly marshalled between the boost wrapper and the python code.
Also, the boost wrapper was not unpacking the return value of the callback with the appropriate tag.

### DIFF
--- a/device/iothub_client_python/src/iothub_client_python.cpp
+++ b/device/iothub_client_python/src/iothub_client_python.cpp
@@ -1225,7 +1225,7 @@ InboundDeviceMethodCallback(
     ScopedGILAcquire acquire;
     try
     {
-        boost::python::object user_response_obj = deviceMethodCallback(method_name_std_string, payload_std_string, (unsigned long int)method_id, userContext);
+        boost::python::object user_response_obj = deviceMethodCallback(method_name_std_string, payload_std_string, (unsigned long long)method_id, userContext);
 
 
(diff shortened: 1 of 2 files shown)


device/iothub_client_python/src/iothub_client_python.cpp
```diff
@@ -1225,11 +1225,10 @@ InboundDeviceMethodCallback(
     ScopedGILAcquire acquire;
     try
     {
-        boost::python::object user_response_obj = deviceMethodCallback(method_name_std_string, payload_std_string, method_id, userContext);
+        boost::python::object user_response_obj = deviceMethodCallback(method_name_std_string, payload_std_string, (unsigned long long)method_id, userContext);
 
-        DeviceMethodReturnValue user_response_value = boost::python::extract<DeviceMethodReturnValue>(user_response_obj)();
 
-        retVal = user_response_value.status;
+        retVal = boost::python::extract<int>(user_response_obj)();
```
zolvarga: >  retVal = boost::python::extract<int>(user_response_obj)(); [](start = 6, length = 60)

Use the DeviceMethodReturnValue instead as in the DeviceMethodCallback. The code what you replaced should work.

pierreca: this method only expects the callback to return an integer and the response itself will be sent with IoTHubClient.device_method_response()

PR #239, opened 2018-11-26
DPS Update

cartertinney:
# Checklist
- [x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-python/blob/master/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
  - [x] This pull-request is submitted against the `master` branch. 
  - [x] I have merged the latest `master` branch prior to submission and re-merged as needed after I took any feedback.
  - [ ] I have squashed my changes into one with a clear description of the change.

  
# Description of the problem

Added new DPS features without breaking existing API

# Description of the solution

Used new generated code from the preview SDK, while adjusting the old wrapper to work with it


provisioning_service_client/samples/enrollment_group_sample.py
```diff
@@ -28,7 +28,7 @@ def main():
     six.print_(eg)
 
     #make a Provisioning Service query
-    qs = QuerySpecification("*")
+    qs = QuerySpecification(query="*")
```
pierreca: >query [](start = 28, length = 5)

isn't switching to keyword args a breaking change?

cartertinney: Switching to **enforced** keyword args is a breaking change. However, changing the syntax of a sample to favor keyword args (for consistency with the **new** objects which **do** use enforced keyword args) is not breaking since it supports both.

PR #249, opened 2019-01-02
Add python SDK version in product_info

pierreca:


device/iothub_client_python/src/iothub_client_python.cpp
```diff
@@ -1513,6 +1514,17 @@ class IoTHubClient
         {
             throw IoTHubClientError(__func__, IOTHUB_CLIENT_ERROR);
         }
+        else
+        {
+            if (client_interface_type == CLIENT_INTERFACE_DEVICE)
+            {
+                IoTHubDeviceClient_SetOption(iotHubClientHandle, "product_info", PYTHON_PRODUCT_INFO);
```
jspaith: >"product_info" [](start = 65, length = 14)

Please use iothub_client_options.h, OPTION_PRODUCT_INFO (and in line below also).

device/iothub_client_python/src/iothub_client_python.cpp
```diff
@@ -53,6 +53,7 @@
 #endif
 
 #define VERSION_STRING "1.4.5"
+#define PYTHON_PRODUCT_INFO "python/" VERSION_STRING
```
jspaith: >VERSION_STRING [](start = 38, length = 14)

Is there any other info that would be useful to get here?  The C SDK gives you a bunch of stuff free (see e.g. [https://github.com/Azure/azure-c-shared-utility/blob/1d622902d7842f94193fc394987f2b4e978bb700/adapters/platform_win32.c#L62](https://github.com/Azure/azure-c-shared-utility/blob/1d622902d7842f94193fc394987f2b4e978bb700/adapters/platform_win32.c#L62) for Win32).  One obvious would be which version of Python you're running (2 vs 3) and maybe there's other no-PII Python specific settings that may be useful??

device/iothub_client_python/src/iothub_client_python.cpp
```diff
@@ -1513,6 +1514,17 @@ class IoTHubClient
         {
             throw IoTHubClientError(__func__, IOTHUB_CLIENT_ERROR);
         }
+        else
+        {
+            if (client_interface_type == CLIENT_INTERFACE_DEVICE)
```
jspaith: >if (client_interface_type == CLIENT_INTERFACE_DEVICE) [](start = 12, length = 53)

Nit - but I'd recommend having a common function like IoTHubClient::SetProductInfo() that all the other CreateHandle's call into.  This way if there ever turns out to be other ways of having a CLIENT_INTERFACE_MODULE, you don't sprinkle this logic all over the place but instead just have it work from get-go.

device/iothub_client_python/src/iothub_client_python.cpp
jspaith: >PYTHON_PRODUCT_INFO [](start = 8, length = 19)

Just in general - is there any desire for PY callers themselves to be able to set this value?  Same way C SDK lets you append stuff, should there be a python_object.set_option("Product_info") that appends 3rd party logic onto what you're building up here?

pierreca: i'm fine with this :) 

---
In reply to: [244873976](https://github.com/Azure/azure-iot-sdk-python/pull/249#discussion_r244873976) [](ancestors = 244873976)

PR #250, opened 2019-01-02
xcode fix for C submodule

jspaith:
Update C SDK submodule, including addressing C break for xcode warnings.


pierreca:
taken care of with #248

PR #269, opened 2019-04-05
Pin boost version to 1.58 to fix Azure Devops build agent

pierreca:
Looks like the msft-hosted build agents for Ubuntu 16.04 now have Boost 1.69 installed and this version isn't building with our SDK so we need to pin the version to 1.58


pierreca:
unnecessary - can be dealt with using the BOOST_ROOT env variable.

PR #285, opened 2019-08-29
create random registration names

olivakar:
sometimes enrollments are being left out because we are cancelling builds and that is causing issues for the next build to run.

have added random uuids to the names of the registration names so that consecutive builds can run even with getting cancelled.

also added time out.


cartertinney:
As I mentioned offline, there's still an issue in that none of these devices get cleared. So even with adding random elements to the generated names, there still may be collisions over a long span of time. I would like to see a PR that would clear out any existing provisionings before making anything new

Issue #148, opened 2018-06-15
Python Interpreter shutdown when using multithreading


Cnidarias:
- **OS and version used:** Ubuntu 16.04.3

- **Python runtime used:** Python 3.5.2

- **SDK version used:** azure-iothub-device-client==1.3.5

# Description of the issue:
When sending data to an IoT-Hub using the azure-iothub-device-client lib the python interpreter crashes after some undefined period of time when using multithreading.
This behaviour is 100% reproducible but the time it takes for it to happen is varied. When using more threads and sending more frequently it seems that it will happen faster. When using 50 threads and sending every second it crashed after about 30 minutes. 

The error one gets is simply:
Fatal Python error: GC object already tracked

Sometimes this comes with a SIGABRT but not always.

Since I ran python with gdb I have attached a backtrace will all the info 
I think the most important parts are at the top - I also included a py-bt of all threads which most of them are just sleeping
[gdb.txt](https://github.com/Azure/azure-iot-sdk-python/files/2105347/gdb.txt)

# Code sample exhibiting the issue:
```
import threading
import json
import time
import random

from iothub_client import IoTHubClient, IoTHubTransportProvider, IoTHubMessage, IoTHubError


class AzureThreadingTest(threading.Thread):
    def __init__(self, azure_config_file, name):
        super().__init__(name=name)
        self.MSG_COUNT = 0
        self.SEND_CALLBACKS = 0

        self.sleep_time = 1
        self.message_print_time = 30
        self.name = name

        with open(azure_config_file) as json_data:
            self.azure_config = json.load(json_data)

        self.iothub_client = self._iothub_client_init()

    def _iothub_client_init(self):
        connection_string = self.azure_config["iot_hub"]["ConnectionString"]
        protocol = self.azure_config["iot_hub"]["Options"]["PROTOCOL"]
        if protocol == "HTTP":
            protocol_class = IoTHubTransportProvider.HTTP
        elif protocol == "AMQP":
            protocol_class = IoTHubTransportProvider.AMQP
        elif protocol == "MQTT":
            protocol_class = IoTHubTransportProvider.MQTT
        else:
            protocol_class = IoTHubTransportProvider.HTTP
        client = IoTHubClient(connection_string, protocol_class)
        client.set_option("messageTimeout", self.azure_config["iot_hub"]["Options"]["MESSAGE_TIMEOUT"])

        if client.protocol == IoTHubTransportProvider.HTTP:
            timeout = 241000
[...]

PR #293, opened 2019-09-05
reduce noise at level=logging.INFO by moving messages to logging.DEBUG

BertKleewein:
Based on feedback left by @LouanDuToitS3 at https://github.com/Azure/azure-iot-sdk-python/issues/148#issuecomment-527037904

### DIFF
--- a/azure-iot-device/azure/iot/device/common/mqtt_transport.py
+++ b/azure-iot-device/azure/iot/device/common/mqtt_transport.py
@@ -133,7 +133,7 @@ class MQTTTransport(object):
                         logger.error("Unexpected error calling on_mqtt_connection_failure_handler")
                         logger.error(traceback.format_exc())
                 else:
-                    logger.info(
+                    logger.warning(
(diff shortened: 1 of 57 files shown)


cartertinney:
Good idea, this was a change we were definitely due for.

LouanDuToitS3:
Hi, @BertKleewein 
I've tried the latest release, preview 13 ( 01caaf7). But I still find the logs to be very verbose under normal circumstances. Did this change make it into the release?

BertKleewein:
Hi @LouanDuToitS3,

This fix did make it into the release.  We reduced the number of messages, but it sounds like we didn't reduce it enough for your scenario.

Would it is help to limit our messages to just errors?
```
 logging.getLogger("azure.iot.device").setLevel(level=logging.ERROR)
```
To reduce it even further, you could even set it to logging.CRITICAL, which would remove all messages from our component.

LouanDuToitS3:
Hi @BertKleewein 

That is perfect, I did not realise that you used a named logger for the logging. So all good, thanks.

PR #299, opened 2019-09-07
[REFACTOR] Removed rogue print statement

cartertinney:
There was a print statement executing upon sending a twin patch. No longer!


cartertinney:
Fixed in #292

PR #331, opened 2019-09-27
feature(azure-iot-device): mqtt websockets on device client

YoDaMa:


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -44,13 +44,14 @@ def __init__(self, iothub_pipeline):
         self._edge_pipeline = None
 
     @classmethod
-    def create_from_connection_string(cls, connection_string, ca_cert=None):
+    def create_from_connection_string(cls, connection_string, ca_cert=None, transport=None):
```
BertKleewein: >None [](start = 86, length = 4)

If the default really is mqtt, then you should have "transport=mqtt" in the function definition.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -113,14 +113,20 @@ def _create_mqtt_client(self):
 
         # Instantiate client
         mqtt_client = mqtt.Client(
-            client_id=self._client_id, clean_session=False, protocol=mqtt.MQTTv311
+            client_id=self._client_id, 
+            clean_session=False, 
+            protocol=mqtt.MQTTv311,
+            transport="websockets"
```
BertKleewein: >transport="websockets" [](start = 12, length = 22)

This file will obviously change later.  Specifically, it will send transport="tcp" if not using websockets and transport="websockets" only if using websockets.  The call to ws_set_options is only necessary if you're using websockets, and tcp and websockets both use different ports.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -44,13 +44,15 @@ def __init__(self, iothub_pipeline):
...
     @classmethod
-    def create_from_connection_string(cls, connection_string, ca_cert=None):
+    def create_from_connection_string(cls, connection_string, ca_cert=None, **kwargs):
         """
         Instantiate the client from a IoTHub device or module connection string.
 
         :param str connection_string: The connection string for the IoTHub you wish to connect to.
         :param str ca_cert: The trusted certificate chain. Only necessary when using a
             connection string with a GatewayHostName parameter.
+        :param str transport: The transport for IoTHub you wish to use. Defaults to MQTT.
```
YoDaMa: need to remove this line.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub.py
```diff
@@ -5,6 +5,27 @@
 # --------------------------------------------------------------------------
 from azure.iot.device.common.pipeline import PipelineOperation
 
+class SetTransportOperation(PipelinOperation):
```
BertKleewein: >SetTransportOperation [](start = 6, length = 21)

This is a dead object, right?

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -21,15 +21,25 @@
...
 class IoTHubPipeline(object):
-    def __init__(self, auth_provider):
+    def __init__(self, auth_provider, **kwargs):
         """
         Constructor for instantiating a pipeline adapter object
         :param auth_provider: The authentication provider
+        :param protocol: The protocol for connecting to IoT Hub
         """
+        try:
+            self._config = ConfigProvider(**kwargs)
```
BertKleewein: >self._config = ConfigProvider(**kwargs) [](start = 12, length = 39)

Nice!  
I like how you used **kwargs on the public interface, but then used explicit keyword args in the ConfigProvider object.  This is a great balance between the freedom that **kwargs gives us and the checking we get with explicit argument names

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,15 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+class ConfigProvider(object):
```
BertKleewein: >ConfigProvider [](start = 6, length = 14)

ConfigProvider isn't quite the right name.  "Provider" is too generic -- I don't know what it mean to provide.  How about IoTHubPipelineOptions?  
Also, don't forget that we'll need to do something similar for DPS.  At this layer, it's OK to call it IoTHubPipelineOptions, but in the pipeline, I would just call it "PipelineOptions". I think we want the freedom to add IoTHub-only options in the future, but for now, the only option we have applies to both.  So the question becomes: do we make 2 parallel option structures for the top layer or do we have a single option structure and worry about diverging the options later?

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -101,12 +111,12 @@ def _on_disconnected():
 
         if isinstance(auth_provider, X509AuthenticationProvider):
             op = pipeline_ops_iothub.SetX509AuthProviderOperation(
-                auth_provider=auth_provider, callback=callback
-            )
+                auth_provider=auth_provider, websockets=self._config.websockets, 
+                callback=callback)
```
BertKleewein: >a [](start = 26, length = 1)

You have the pipeline extracting the "websockets" value at the top and then passing it down.  This is good, but it means you have to update the pipeline internals to pass it from the top to the bottom.  Think about adding more pipeline options in the future -- do you want to add more lines  to pipeline_stages_iothub for every single option you add.   Instead, I might suggest adding the entire structure as a member of PipelineRootStage.  This way, all stages can get to it via self.pipeline_root and you can acces the attribute only from the stage that actually cares about the value.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub.py
YoDaMa: yes, need to remove.

azure-iot-device/azure/iot/device/iothub/config.py
YoDaMa: We were trying to shy away from calling it Options since we already have Operations shortened to Ops, and the usual shortening of Options is Ops as well... so it's a bit of a naming confusion.

YoDaMa: How's `IoTHubPipelineConfig`?

azure-iot-device/azure/iot/device/common/config.py
```diff
@@ -0,0 +1,13 @@
...
+
+logger = logging.getLogger(__name__)
+
+class BasePipelineConfig(object):
+    """A base class for storing all configurations/options shared across the Azure IoT Python Device Client Library.
+    More specific configurations such as those that only apply to the IoT Hub Client will be found in the respective 
+    config files.
+    """
+
+    __init__(self):
```
olivakar: >_init__(self) [](start = 5, length = 13)

The syntax looks a bit off..I think it is `def __init__(self):`

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -80,7 +80,7 @@ class MQTTTransport(object):
     :type on_mqtt_connection_failure_handler: Function
     """
 
-    def __init__(self, client_id, hostname, username, ca_cert=None, x509_cert=None):
+    def __init__(self, client_id, hostname, username, ca_cert=None, x509_cert=None, websockets=False):
```
olivakar: >websockets=False [](start = 84, length = 16)

as an param in the doc string portion

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -181,12 +182,15 @@ class PipelineRootStage(PipelineStage):
     :type on_disconnected_handler: Function
     """
 
-    def __init__(self):
+    def __init__(self, configObj=None):
```
olivakar: >configObj [](start = 23, length = 9)

I dont think python has camel case..it may work but generally we name our variables as underscored separations of words...

for example client id is `client_id` but not 'clientId`

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub.py
```diff
@@ -90,6 +89,7 @@ def __init__(
         :param X509 client_cert: (Optional) The x509 object containing a client certificate and key used to connect
           to the service
         :param str sas_token: The token string which will be used to authenticate with the service
+        :param str transport: The transport string which will be used to connect to IoT Hub.
```
olivakar: >transport [](start = 19, length = 9)

delete doc line

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -181,12 +182,15 @@ class PipelineRootStage(PipelineStage):
...
         super(PipelineRootStage, self).__init__()
         self.on_pipeline_event_handler = None
         self.on_connected_handler = None
         self.on_disconnected_handler = None
         self.connected = False
+        if (issubclass(configObj.__class__, PipelineConfig)):
```
olivakar: >(issubclass(configObj.__class__, PipelineConfig)) [](start = 11, length = 49)

The brackets are redundant

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_mqtt.py
```diff
@@ -45,6 +46,7 @@ def __init__(
         self.ca_cert = ca_cert
         self.client_cert = client_cert
         self.sas_token = sas_token
+        self.websockets = websockets
```
olivakar: >self.websockets = websockets [](start = 8, length = 28)

Is this getting used anywhere ? I think you are directly passing this into MQTTTransport from the pipeline root

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -65,6 +65,7 @@ def _execute_op(self, op):
                 username=self.username,
                 ca_cert=self.ca_cert,
                 x509_cert=self.client_cert,
+                websockets=self.pipeline_root.configObj.websockets
```
olivakar: >websockets=self.pipeline_root.configObj.websockets [](start = 16, length = 50)

I thought the op already has the variable websockets when we initialized the op..

YoDaMa: The op will have access to the variable websockets via the pipeline_root. That's not to say that the MQTT_Transport has access to the websockets parameter, hence why we're passing it in here as a value.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_mqtt.py
YoDaMa: good catch

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
YoDaMa: this has been done now by adding a configurations object to the PipelineRootStage initialization as a parameter.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -181,12 +182,15 @@ class PipelineRootStage(PipelineStage):
...
     """
 
-    def __init__(self):
+    def __init__(self, pipeline_configurations=None):
         super(PipelineRootStage, self).__init__()
         self.on_pipeline_event_handler = None
         self.on_connected_handler = None
         self.on_disconnected_handler = None
         self.connected = False
+        if issubclass(pipeline_configurations.__class__, PipelineConfig):
```
BertKleewein: >issubclass [](start = 11, length = 10)

The more typical way is `if isinstance(pipeline_configuration, PipelineConfig):`

BertKleewein: >issubclass [](start = 11, length = 10)

And even more pythonic is to skip the type check entirely.  You don't care that it's a subclass.  You only care that it has the right attributes.  If it doesn't, something will raise when it tries to access them if it's really important  This is the EAFP vs. LBYL conversation that always comes up with new Python programmers.  google it or ask Carter.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -44,13 +44,14 @@ def __init__(self, iothub_pipeline):
         self._edge_pipeline = None
 
     @classmethod
-    def create_from_connection_string(cls, connection_string, ca_cert=None):
+    def create_from_connection_string(cls, connection_string, ca_cert=None, **kwargs):
```
BertKleewein: >**kwargs [](start = 76, length = 8)

Question for the future: how are we going to document these kwargs?  We want them somewhere and this seems like the right place..

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,15 @@
+import logging
+from azure.iot.device.common.config import PipelineConfig
+logger = logging.getLogger(__name__)
+
+class IoTHubPipelineConfig(PipelineConfig):
```
BertKleewein: >PipelineConfig [](start = 27, length = 14)

Not sure if we want the base class or not.  I like the idea of a place to store options that are common between DPS and IoTHub, but I'm not sure if a class hierarchy is overkill or not.  Right now, we have a single option that applies to both, and the other options (like keepalive interval, sas renewal intervals, and retry policy parameters) all apply to both.  So maybe we only need a base class and not the derived classes.  We can always add them later if we need them.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub.py
```diff
@@ -5,7 +5,6 @@
 # --------------------------------------------------------------------------
 from azure.iot.device.common.pipeline import PipelineOperation
 
-
 # TODO: Combine SetAuthProviderOperation and SetX509AuthProviderOperation once
```
BertKleewein: Make sure you have pre-commit installed.  I'm pretty sure this deleted line would cause a pre-commit failure.

azure-iot-device/azure/iot/device/iothub/config.py
YoDaMa: ok that sounds fair

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -84,21 +84,25 @@ class MQTTTransport(object):
...
+        self, client_id, hostname, username, ca_cert=None, x509_cert=None, websockets=False
+    ):
         """
         Constructor to instantiate an MQTT protocol wrapper.
         :param str client_id: The id of the client connecting to the broker.
         :param str hostname: Hostname or IP address of the remote broker.
         :param str username: Username for login to the remote broker.
         :param str ca_cert: Certificate which can be used to validate a server-side TLS connection (optional).
         :param x509_cert: Certificate which can be used to authenticate connection to a server in lieu of a password (optional).
+        :param websockets: Boolean value to determine whether or not to enable a websockets connection in the Transport.
```
cartertinney: >Boolean valu [](start = 27, length = 12)

You can identify the type in the markup with :param bool websockets: rather than having to specify it in the description

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -14,6 +14,7 @@
 from . import operation_flow
 from . import pipeline_thread
 from azure.iot.device.common import handle_exceptions
+from azure.iot.device.common.config import BasePipelineConfig
```
cartertinney: This import is totally unnecessary

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -15,6 +15,7 @@
     pipeline_thread,
     pipeline_exceptions,
 )
+
 from azure.iot.device.common.mqtt_transport import MQTTTransport
```
cartertinney: Remove the blank line added above

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Yes, API-wise I think we need to discuss as a whole team what we want this experience to be like.

---
In reply to: [330704305](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r330704305) [](ancestors = 330704305)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -44,13 +44,14 @@ def __init__(self, iothub_pipeline):
...
         """
         Instantiate the client from a IoTHub device or module connection string.
 
         :param str connection_string: The connection string for the IoTHub you wish to connect to.
         :param str ca_cert: The trusted certificate chain. Only necessary when using a
             connection string with a GatewayHostName parameter.
+        :**kwargs config: Set of keyword arguments for setting options.
```
cartertinney: >:**kwargs config: Set of keyword arguments for setting options. [](start = 8, length = 63)

Right now the correct documentation (at least as best as the doc team has) is to list the optional arguments here, rather than using **kwargs. Which makes sense - **kwargs doesn't tell us anything useful.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -21,15 +21,26 @@
...
-    def __init__(self, auth_provider):
+    def __init__(self, auth_provider, **kwargs):
         """
         Constructor for instantiating a pipeline adapter object
         :param auth_provider: The authentication provider
+        :param protocol: The protocol for connecting to IoT Hub
         """
+        try:
+            # Question: should this be kept in self? Is that necessary?
+            self._config = IoTHubPipelineConfig(**kwargs)
```
cartertinney: >self._config = IoTHubPipelineConfig(**kwargs) [](start = 12, length = 45)

Per your question, no it shouldn't. Not necessary to keep a reference to it hanging around here. At least for now.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -21,15 +21,26 @@
...
+        except TypeError as error:
```
cartertinney: >except TypeError as error: [](start = 8, length = 26)

What would cause the TypeError?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -181,12 +182,13 @@ class PipelineRootStage(PipelineStage):
     :type on_disconnected_handler: Function
     """
 
-    def __init__(self):
+    def __init__(self, pipeline_configurations=None):
```
cartertinney: >pipeline_configurations=None [](start = 23, length = 28)

This shouldn't be optional, since other parts of the code directly access the pipeline_configuration. There's no case where we wouldn't have this.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -76,8 +76,8 @@ def _execute_op(self, op):
                     username=username,
                     ca_cert=op.ca_cert,
                     client_cert=op.client_cert,
-                    sas_token=op.sas_token,
-                ),
+                    sas_token=op.sas_token
+                )
```
cartertinney: Why did you change this?

YoDaMa: these commas seemed unnecessary.

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,16 @@
+import logging
+from azure.iot.device.common.config import BasePipelineConfig
+
+logger = logging.getLogger(__name__)
+
+
+class IoTHubPipelineConfig(BasePipelineConfig):
+    """A BasePipelineConfig object which contains all possible options for the iothub client.
```
olivakar: >BasePipelineConfig [](start = 9, length = 18)

A iothub specific config object...

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -115,10 +119,22 @@ def _create_mqtt_client(self):
...
+        # Instaniate the client
+        if self.websockets:
+            # MQTT Over Websockets
+            mqtt_client = mqtt.Client(
+                client_id=self._client_id,
+                clean_session=False,
+                protocol=mqtt.MQTTv311,
+                transport="websockets",
+            )
+            mqtt_client.ws_set_options(path="/$iothub/websocket")
```
olivakar: >$iothub [](start = 46, length = 7)

is this value for all iothub and dps ?

YoDaMa: good question. I'm not sure on the DPS path, but it's likely different. Still, this PR will be scoped to just Device Client.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
YoDaMa: The TypeError would be from kwargs not mapping properly, yes. I'm adding this for more verbose error logging in case the user messes up the options.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
YoDaMa: kk

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -61,24 +62,26 @@ def create_from_connection_string(cls, connection_string, ca_cert=None):
         # in order to differentiate types of connection strings.
         authentication_provider = auth.SymmetricKeyAuthenticationProvider.parse(connection_string)
         authentication_provider.ca_cert = ca_cert  # TODO: make this part of the instantiation
-        iothub_pipeline = pipeline.IoTHubPipeline(authentication_provider)
+        iothub_pipeline = pipeline.IoTHubPipeline(authentication_provider, **kwargs)
```
cartertinney: >**kwargs [](start = 75, length = 8)

I actually think options should be instantiated in this scope rather than passing the kwargs down. I think we should convert kwargs into a concrete object as soon as possible.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: After doing some reading, I'm on board with the use of **kwargs

---
In reply to: [331164870](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r331164870) [](ancestors = 331164870,330704305)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Since we have now agreed upon usage of **kwargs, I would want to see the individual options documented here

---
In reply to: [331168121](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r331168121) [](ancestors = 331168121)

azure-iot-device/azure/iot/device/common/config.py
```diff
@@ -0,0 +1,17 @@
+import logging
```
cartertinney: Missing Microsoft license header

cartertinney: This file should arguably be moved to the pipeline subpackage, but I suppose it's not crucial to do right this minute.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -83,7 +83,7 @@ def apply_fake_pipeline_thread(fake_pipeline_thread):
 @pytest.fixture
 def stage(mocker):
     stage = pipeline_stages_mqtt.MQTTTransportStage()
-    root = pipeline_stages_base.PipelineRootStage()
+    root = pipeline_stages_base.PipelineRootStage("fakeConfiguration")
```
cartertinney: >"fakeConfiguration" [](start = 50, length = 19)

Use a mocker.MagicMock() instead of a string

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -1030,7 +1030,7 @@ def test_fails_on_invalid_feature_name(
 
 @pytest.fixture
 def add_pipeline_root(stage, mocker):
-    root = pipeline_stages_base.PipelineRootStage()
+    root = pipeline_stages_base.PipelineRootStage("fakeConfiguration")
```
cartertinney: >"fakeConfiguration" [](start = 50, length = 19)

Also here use a MagicMock()

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning_mqtt.py
```diff
@@ -351,7 +351,7 @@ def test_converts_feature_name_to_topic(
 
 @pytest.fixture
 def add_pipeline_root(mock_stage, mocker):
-    root = pipeline_stages_base.PipelineRootStage()
+    root = pipeline_stages_base.PipelineRootStage("fakeConfiguration")
```
cartertinney: >"fakeConfiguration" [](start = 50, length = 19)

Ditto

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
YoDaMa: done.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
YoDaMa: I think actually this should be optional at this point. Not every root pipeline stage has the need for options to be implemented right now, and it adds a large amount of test overhead to make this default.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -25,11 +25,13 @@
 
 
 class IoTHubPipeline(object):
-    def __init__(self, auth_provider):
+    def __init__(self, auth_provider, pipelineConfiguration):
```
olivakar: >pipelineConfiguration [](start = 38, length = 21)

rename to  `pipeline_configuration`

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -148,6 +148,32 @@ def test_instantiates_mqtt_client(self, mocker):
             client_id=fake_device_id, clean_session=False, protocol=mqtt.MQTTv311
         )
 
+    @pytest.mark.it(
+        "Creates an instance of the Paho MQTT Client using Websockets, if websockets parameter is True"
```
olivakar: >if websockets parameter is True [](start = 71, length = 31)

with websockets to be True

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -157,6 +159,7 @@ def callback(*arg, **argv):
 TestPipelineRootStagePipelineThreading.test_pipeline_root_runs_on_event_received_in_callback_thread = (
     _test_pipeline_root_runs_on_event_received_in_callback_thread
 )
+"""
```
olivakar: are these to be removed ?

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -230,12 +239,14 @@ def create_from_edge_environment(cls):
                 new_err = OSError("Unexpected failure in IoTEdge")
                 new_err.__cause__ = e
                 raise new_err
-        iothub_pipeline = pipeline.IoTHubPipeline(authentication_provider)
+
+        pipelineConfiguration = config.BasePipelineConfig(**kwargs)
```
cartertinney: >pipelineConfiguration [](start = 8, length = 21)

Here and everywhere - camelCase indicates a class definition in Python. Please always use snake_case for everything else.

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -76,11 +76,17 @@ class SharedClientCreateFromConnectionStringTests(object):
...
-            pytest.param(None, id="No CA certificate"),
-            pytest.param("some-certificate", id="With CA certificate"),
+            pytest.param(None, id=" No CA certificate "),
+            pytest.param("some-certificate", id=" With CA certificate "),
         ],
     )
-    def test_auth_provider_creation(self, mocker, client_class, connection_string, ca_cert):
+    @pytest.mark.parametrize(
+        "websockets",
+        [pytest.param(None, id=" No Websockets "), pytest.param(True, id=" With Websockets ")],
```
olivakar: > [pytest.param(None, id=" No Websockets "),  [](start = 7, length = 44)

is None equivalent to  False since False goes to default else block ?

YoDaMa: Yes. None is equivalent to False and also exercises that the default is used.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
YoDaMa: fixed.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
YoDaMa: no. good catch.

cartertinney:
> class TestMQTTProviderRunOpWithSetConnectionArgs(RunOpTests):

We need a test that shows that websockets is passed to the transport based on the value in the configuration object

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py:201 in 7b6384f. [](commit_id = 7b6384fd71502ce0414ac0e61389d8d631f91ceb, deletion_comment = False)

cartertinney:
>     def test_passes_right_params(self, stage, transport, mocker, op_set_connection_args):

We need a test that shows that websockets is passed to the transport based on the value in the configuration object. This might require slight restructure of the contents of this test and/or making a new test

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py:210 in 7b6384f. [](commit_id = 7b6384fd71502ce0414ac0e61389d8d631f91ceb, deletion_comment = False)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: There should also be a category of False. You want to test:

1) Default - No Websockets
2) Websockets == True
3) Websockets == False

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -99,59 +107,82 @@ def test_auth_provider_creation(self, mocker, client_class, connection_string, c
...
+    )
     def test_pipeline_creation(
-        self, mocker, client_class, connection_string, ca_cert, mock_pipeline_init
+        self, mocker, client_class, connection_string, ca_cert, websockets, mock_pipeline_init
     ):
         mock_auth = mocker.patch(
             "azure.iot.device.iothub.auth.SymmetricKeyAuthenticationProvider"
         ).parse.return_value
 
+        mock_config = mocker.patch(
```
cartertinney: Here and everywhere - Need to have a test showing that the config is set up correctly, i.e. is called with the correct args

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -99,59 +107,82 @@ def test_auth_provider_creation(self, mocker, client_class, connection_string, c
     @pytest.mark.parametrize(
         "ca_cert",
         [
-            pytest.param(None, id="No CA certificate"),
+            pytest.param(None, id=" No CA certificate "),
             pytest.param("some-certificate", id="With CA certificate"),
         ],
     )
+    @pytest.mark.parametrize(
```
cartertinney: Also, with the way these parameters are stacking, we may need to at some point reconsider how these tests are structured now that there's much more variable input.

But don't worry about this now.

azure-iot-device/tests/iothub/test_sync_clients.py
YoDaMa: done.

azure-iot-device/tests/iothub/test_sync_clients.py
YoDaMa: yeah we'll have to think about how to improve these tests without making them an ultimate labyrinth....

azure-iot-device/tests/iothub/test_sync_clients.py
YoDaMa: done.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -185,12 +185,13 @@ class PipelineRootStage(PipelineStage):
...
         super(PipelineRootStage, self).__init__()
         self.on_pipeline_event_handler = None
         self.on_connected_handler = None
         self.on_disconnected_handler = None
         self.connected = False
+        self.pipeline_configurations = pipeline_configurations
```
BertKleewein: >pipeline_configurations [](start = 39, length = 23)

nit: "configurations" with an s implies that this is possibly more than one configuration object.  My first thought is that this holds some set of possible configurations and there must be some way to choose between them.

YoDaMa: done

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -76,11 +76,21 @@ class SharedClientCreateFromConnectionStringTests(object):
     @pytest.mark.parametrize(
         "ca_cert",
         [
-            pytest.param(None, id="No CA certificate"),
-            pytest.param("some-certificate", id="With CA certificate"),
+            pytest.param(None, id=" No CA certificate"),
+            pytest.param("some-certificate", id=" With CA certificate"),
```
BertKleewein: was the extra space intentional?  Why?

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -99,59 +111,98 @@ def test_auth_provider_creation(self, mocker, client_class, connection_string, c
...
         kwargs = {}
         if ca_cert:
             kwargs["ca_cert"] = ca_cert
+        if websockets:
+            kwargs["websockets"] = websockets
         client_class.create_from_connection_string(*args, **kwargs)
-
+        assert mock_config_init.call_count == 1
+        if websockets:
+            assert mock_config_init.call_args == ({"websockets": websockets},)
```
BertKleewein: >assert mock_config_init.call_args == ({"websockets": websockets},) [](start = 12, length = 66)

I'm not sure if this is significant or not, but I think it might be more appropriate to use mocker.call(), as in: `assert mock_config_init.call_args == mocker.call(websockets=websockets)`

azure-iot-device/tests/common/pipeline/helpers.py
```diff
@@ -78,7 +78,10 @@ def stage_execute_op(self, op):
         else:
             assert False
 
-    first_stage = stage_to_make()
+    if stage_to_make == pipeline_stages_base.PipelineRootStage:
+        first_stage = stage_to_make(None)
```
BertKleewein: >first_stage = stage_to_make(None) [](start = 8, length = 33)

Can you mark this line with `# BKTODO: make this more generic` please?  I have a better fix for this and I don't want to lose the spot.

BertKleewein: And I thought this hack was in 2 places but I can't find the other one now.  If there is another place, can you mark it also please?

---
In reply to: [335153070](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r335153070) [](ancestors = 335153070)

YoDaMa: found the other points in the workaround and added this.

azure-iot-device/tests/iothub/test_sync_clients.py
YoDaMa: so, I think mocker call actually changes things? I looked this up, because originally I did `mocker.call()` but the test failed. Looked up the documentation and they explained it nicely:

```
>>> mock = Mock(return_value=None)
>>> print(mock.call_args)
None
>>> mock()
>>> mock.call_args
call()
>>> mock.call_args == ()
True
>>> mock(3, 4)
>>> mock.call_args
call(3, 4)
>>> mock.call_args == ((3, 4),)
True
>>> mock.call_args.args
(3, 4)
>>> mock.call_args.kwargs
{}
>>> mock(3, 4, 5, key='fish', next='w00t!')
>>> mock.call_args
call(3, 4, 5, key='fish', next='w00t!')
>>> mock.call_args.args
(3, 4, 5)
>>> mock.call_args.kwargs
{'key': 'fish', 'next': 'w00t!'}
```

azure-iot-device/samples/advanced-hub-scenarios/send_message_over_websockets.py
```diff
@@ -0,0 +1,37 @@
...
+
+async def main():
+    # Fetch the connection string from an enviornment variable
+    conn_str = os.getenv("IOTHUB_DEVICE_CONNECTION_STRING")
+
+    # Create instance of the device client using the connection string
+    device_client = IoTHubDeviceClient.create_from_connection_string(conn_str, websockets=True)
+
+    # Connect the device client.
+    await device_client.connect()
```
cartertinney: This isn't necessary, remove (I know some samples have it, but we're trying to avoid it)

azure-iot-device/samples/advanced-hub-scenarios/send_message_over_websockets.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Send a single message
+    print("Sending message...")
+    await device_client.send_message("This is a message that is being sent")
+    print("Message successfully sent!")
+
+    # finally, disconnect
+    await device_client.disconnect()
```
cartertinney: Also not necessary

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -59,13 +59,28 @@ def _execute_op(self, op):
...
+                    username=self.username,
+                    ca_cert=self.ca_cert,
+                    x509_cert=self.client_cert,
+                    websockets=self.pipeline_root.pipeline_configuration.websockets,
+                )
+            else:
+                # TODO:
+                # While MQTTWS in provisioning client is not implemented, this is defaulted
+                # to being false for websockets.
+                # When this is implemented there should be no IF here.
```
cartertinney: This could be solved by always having the configuration in provisioning.

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: +1

---
In reply to: [335151026](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r335151026) [](ancestors = 335151026)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
YoDaMa: yes, so when provisioning is implemented, it would always have the configuration and this if would not be necessary.

---
In reply to: [335713568](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r335713568) [](ancestors = 335713568)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: Yes, please remove

---
In reply to: [335149871](https://github.com/Azure/azure-iot-sdk-python/pull/331#discussion_r335149871) [](ancestors = 335149871)

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -107,21 +128,41 @@ class SharedClientCreateFromConnectionStringTests(object):
...
         kwargs = {}
         if ca_cert:
             kwargs["ca_cert"] = ca_cert
+        if websockets:
+            kwargs["websockets"] = websockets
         client_class.create_from_connection_string(*args, **kwargs)
 
+        assert mock_config_init.call_count == 1
+        if websockets:
+            assert mock_config_init.call_args == ({"websockets": websockets},)
```
cartertinney: >assert mock_config_init.call_args == ({"websockets": websockets},) [](start = 12, length = 66)

Make the change from sync clients here as well

PR #334, opened 2019-09-30
Fix memory leak that was causing instability

BertKleewein:
This looks minor, but I think this is the root of at least some of our random disconnect bugs.  This fixes a problem that was indirectly preventing Paho from cleaning up after itself.


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -80,6 +80,12 @@ class MQTTTransport(object):
     :type on_mqtt_connection_failure_handler: Function
     """
 
+    # Horton uses this variable to make sure we're cleaning up correctly.
+    # This object in particuar is checked because it holds a reference to Paho
+    # and Paho holds references to network resources and we want to be extra sure
+    # that we clean those up correctly.
+    _object_count = 0
```
cartertinney: I don't quite understand the purpose, since the only thing done with this value is increment and decrement. There seem to be no implementation changes based on the value, so doesn't that mean that nothing has changed?

azure-iot-device/azure/iot/device/iothub/auth/base_renewable_token_authentication_provider.py
```diff
@@ -160,9 +171,12 @@ def _schedule_token_update(self, seconds_until_update):
             seconds_until_update,
         )
 
+        self_weakref = weakref.ref(self)
```
cartertinney: >self_weakref = weakref.ref(self) [](start = 8, length = 32)

Could you provide an explanatory comment outlining the scenario in which this is relevant? Does the Auth Provider ever get deleted? My understanding was that it hangs around indefinitely in the transport stage that handles authentication.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: Yes, Horton uses (or will use) this variable to make sure we're cleaning up correctly.  The Python glue checks this as part of test teardown to make sure no objects were leaked.  It's kludgy but I need to know I'm on stable ground for my dropped connection tests. 

---
In reply to: [329764771](https://github.com/Azure/azure-iot-sdk-python/pull/334#discussion_r329764771) [](ancestors = 329764771)

azure-iot-device/azure/iot/device/iothub/auth/base_renewable_token_authentication_provider.py
BertKleewein: The auth provider should get cleaned up when the pipeline gets cleaned up, which happens when the client gets cleaned up.  Usually this is when you do a clean shutdown, but it can also happen if the user deletes their client object and creates a new one.

Because of various circular references (like callbacks, but also next and previous pointers), if the auth provider doesn't get collected, then the entire pipeline doesn't get collected.

I'm also adding a comment  to the code in a new revision.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
olivakar: >_object_count [](start = 4, length = 13)

can u include a line saying only for test or perhaps make this variable to be `horton_test_references`  or  `test_object_count` ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -66,10 +67,18 @@ def _execute_op(self, op):
                 ca_cert=self.ca_cert,
                 x509_cert=self.client_cert,
             )
-            self.transport.on_mqtt_connected_handler = self._on_mqtt_connected
-            self.transport.on_mqtt_connection_failure_handler = self._on_mqtt_connection_failure
-            self.transport.on_mqtt_disconnected_handler = self._on_mqtt_disconnected
-            self.transport.on_mqtt_message_received_handler = self._on_mqtt_message_received
+            self.transport.on_mqtt_connected_handler = CallableWeakMethod(
+                self, "_on_mqtt_connected"
+            )
```
cartertinney: Could this be applied as a decorator?

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -330,7 +335,9 @@ def __init__(self, iothub_pipeline):
         :type iothub_pipeline: :class:`azure.iot.device.iothub.pipeline.IoTHubPipeline`
         """
         super(IoTHubDeviceClient, self).__init__(iothub_pipeline=iothub_pipeline)
-        self._iothub_pipeline.on_c2d_message_received = self._inbox_manager.route_c2d_message
+        self._iothub_pipeline.on_c2d_message_received = CallableWeakMethod(
+            self._inbox_manager, "route_c2d_message"
+        )
```
cartertinney: Is this actually necessary here? Not that it's a problem as is, but I don't think there can be a circular dependency with inbox manager and client, because no reference to the client (or any other part of the code) is held by the inbox manager.

azure-iot-device/azure/iot/device/common/callable_weak_method.py
```diff
@@ -0,0 +1,78 @@
...
+
+import weakref
+
+
+class CallableWeakMethod(object):
+    """
+    Object which makes a weak reference to a method call.  Similar to weakref.WeakMethod,
+    but works on Python 2.7 and returns an object which is callable.
+
+    This objet is used primarily for callbacks and it prevents circular references in the
```
cartertinney: >callbacks [](start = 37, length = 9)

and handlers

azure-iot-device/tests/iothub/auth/test_base_renewable_token_authentication_provider.py
```diff
@@ -49,16 +49,12 @@ def parse(source):
 
 @pytest.fixture(scope="function")
 def device_auth_provider():
-    auth_provider = FakeAuthProvider(fake_hostname, fake_device_id, None)
-    yield auth_provider
-    auth_provider.disconnect()
```
cartertinney: >auth_provider.disconnect() [](start = 4, length = 26)

Took a look at the source, and it doesn't look like removing this would cause any problems, but if that's true, was the inclusion of it here unneeded before?

BertKleewein: disconnect used to be a cleanup function on auth provider.  I added this long ago, before I knew that daemon threads were a thing.  now, instead of having an explicit cleanup via disconnect, I have an implicit cleanup via finalizer. so the disconnect is no longer necessary.

---
In reply to: [332259239](https://github.com/Azure/azure-iot-sdk-python/pull/334#discussion_r332259239) [](ancestors = 332259239)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: No, but that would be nice.  Well, maybe possible, but that would involve some unusual gymnastics.

To briefly explain: Notice the way the method name is a string rather than a function object.  I did it this way because it's easy to make a weak reference to an object, but it's hard to make a weak reference to a method.  I investigated this far enough to dismiss this idea, but not far enough to explain.   

(The lifetime semantics of a bound method are somewhat somewhat unique because of the binding mechanism.  (3.7 doc for weakref.WeakMethod() calls it  "ephemeral")   In 3.4+, there is a weakref.WeakMethod() object that makes it possible to make a weak reference to a method, but this doesn't exist in 2.7.  I've seem WeakMethod implementations for 2.7, but they are old enough and differently-licensed, so i don't want to dig into them.)


---
In reply to: [332250484](https://github.com/Azure/azure-iot-sdk-python/pull/334#discussion_r332250484) [](ancestors = 332250484)

cartertinney: That makes sense, I didn't even think of that, but yeah, best to just avoid. Dealing with the differences in method/function binding across Py 2 vs 3 is a pain.

---
In reply to: [332527083](https://github.com/Azure/azure-iot-sdk-python/pull/334#discussion_r332527083) [](ancestors = 332527083,332250484)

PR #335, opened 2019-09-30
Add DEFAULT_KEEPALIVE to speed up dropped connection test runs

BertKleewein:
This lets me speed up my dropped connection testing by at least an order or magnitude.  We should probably have an officially documented way to set this value, but I don't want that to get in the way.

### DIFF
--- a/azure-iot-device/tests/common/test_mqtt_transport.py
+++ b/azure-iot-device/tests/common/test_mqtt_transport.py
@@ -4,6 +4,7 @@
 # license information.
 # --------------------------------------------------------------------------
 
+import azure.iot.device.common.mqtt_transport as mqtt_transport
 from azure.iot.device.common.mqtt_transport import MQTTTransport, OperationManager
 from azure.iot.device.common.models.x509 import X509
 from azure.iot.device.common import transport_exceptions as errors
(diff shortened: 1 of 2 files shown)


azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -292,10 +294,34 @@ def test_use_no_password(self, mocker, mock_mqtt_client, transport):
...
     def test_calls_paho_connect(self, mocker, mock_mqtt_client, transport, password):
+
+        mqtt_transport.DEFAULT_KEEPALIVE = fake_keepalive
+
         transport.connect(password)
 
         assert mock_mqtt_client.connect.call_count == 1
-        assert mock_mqtt_client.connect.call_args == mocker.call(host=fake_hostname, port=8883)
+        assert mock_mqtt_client.connect.call_args == mocker.call(
+            host=fake_hostname, port=8883, keepalive=mocker.ANY
```
olivakar: >keepalive=mocker.ANY [](start = 43, length = 20)

should this be fake_keepalive ?

BertKleewein: That depends on the order the tests run.  If this test runs after the one that changes it to fake_keepalive, then you're right.  If it runs before that test, then you're wrong.  This test doesn't really care what the value is.

PR #347, opened 2019-10-10
refactor operation_flow as part of retry changes

BertKleewein:
This is part of the retry changes factored out as it's own PR.  I did this for 2 reasons:
1. it made retry easier and cleaner
2. these things have been bothering me

This has 3 major changes:
1. I moved the error from an operation attribute into a callback parameter.  This is because the idea of an operation failing is becoming more transient with retry.  An error is not an integral part of an operation.  Rather, it's a temporary status that represents what happened to the operation most recently.  
2. The operation_flow.py methods are not part of the PipelineStage objects, which is sort-of what they were before.  I also renamed these to (I hope) make more sense.  
3. I added the idea that an op could be marked as "completed" and separated the concepts of `complete_op()` and `send_completed_op_up()`.  This way, we don't have code that completes an op and then completes it again.  The `_send_op_down_and_intercept_return` function was also added for stages that need to keep track of operation (like `EnsureConnectionStage `and `TimeoutStage`)

### DIFF
--- a/azure-iot-device/azure/iot/device/common/chainable_exception.py
+++ b/azure-iot-device/azure/iot/device/common/chainable_exception.py
@@ -14,3 +14,8 @@ class ChainableException(Exception):
         # while still being able to operate in Python 2.
         self.__cause__ = cause
         super(ChainableException, self).__init__(message)
+
+    def __repr__(self):
+        return "{} caused by {}".format(
+            super(ChainableException, self).__repr__(), self.__cause__.__repr__()
+        )
--- a/azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
@@ -15,124 +15,145 @@ from six.moves import queue
 logger = logging.getLogger(__name__)
 
 
-@pipeline_thread.runs_on_pipeline_thread
-def delegate_to_different_op(stage, original_op, new_op):
-    """
-    Continue an operation using a new operation.  This means that the new operation
-    will be passed down the pipeline (starting at the next stage). When that new
-    operation completes, the original operation will also complete.  In this way,
-    a stage can accept one type of operation and, effectively, change that operation
-    into a different type of operation before passing it to the next stage.
+class PipelineFlow(object):
+    @pipeline_thread.runs_on_pipeline_thread
+    def _send_worker_op_down(self, worker_op, op):
+        """
+        Continue an operation using a new operation.  This means that the new operation
+        will be passed down the pipeline (starting at the next stage). When that new
+        operation completes, the original operation will be completed.  In this way,
+        a stage can accept one type of operation and, effectively, change that operation
+        into a different type of operation before passing it to the next stage.
 
-    This is useful when a generic operation (such as "enable feature") needs to be
-    converted into a more specific operation (such as "subscribe to mqtt topic").
-    In that case, a stage's _execute_op function would call this function passing in
-    the original "enable feature" op and the new "subscribe to mqtt topic"
-    op.  This function will pass the "subscribe" down. When the "subscribe" op
-    is completed, this function will cause the original op to complete.
+        This is useful when a generic operation (such as "enable feature") needs to be
+        converted into a more specific operation (such as "subscribe to mqtt topic").
+        In that case, a stage's _execute_op function would call this function passing in
+        the original "enable feature" op and the new "subscribe to mqtt topic"
+        op.  This function will pass the "subscribe" down. When the "subscribe" op
+        is completed, this function will cause the original op to complete.
 
-    This function is only really useful if there is no data returned in the
-    new_op that that needs to be copied back into the original_op before
-    completing it.  If data needs to be copied this way, some other method needs
-    to be used.  (or a "copy data back" function needs to be added to this function
-    as an optional parameter.)
+        This function is only really useful if there is no data returned in the
+        new_op that that needs to be copied back into the original_op before
+        completing it.  If data needs to be copied this way, some other method needs
+        to be used.  (or a "copy data back" function needs to be added to this function
+        as an optional parameter.)
 
-    :param PipelineStage stage: stage to delegate the operation to
-    :param PipelineOperation original_op: Operation that is being continued using a
-      different op.  This is most likely the operation that is currently being handled
-      by the stage.  This operation is not actually continued, in that it is not
-      actually passed down the pipeline.  Instead, the original_op operation is
-      effectively paused while we wait for the new_op operation to complete.  When
-      the new_op operation completes, the original_op operation will also be completed.
-    :param PipelineOperation new_op: Operation that is being passed down the pipeline
-      to effectively continue the work represented by original_op.  This is most likely
-      a different type of operation that is able to accomplish the intention of the
-      original_op in a way that is more specific than the original_op.
-    """
+        :param PipelineOperation op: Operation that is being continued using a
+          different op.  This is most likely the operation that is currently being handled
+          by the stage.  This operation is not actually continued, in that it is not
+          actually passed down the pipeline.  Instead, the original_op operation is
+          effectively paused while we wait for the new_op operation to complete.  When
+          the new_op operation completes, the original_op operation will also be completed.
+        :param PipelineOperation worker_op: Operation that is being passed down the pipeline
+          to effectively continue the work represented by the original op.  This is most likely
+          a different type of operation that is able to accomplish the intention of the
+          original op in a way that is more specific than the original op.
+        """
 
-    logger.debug("{}({}): continuing with {} op".format(stage.name, original_op.name, new_op.name))
+        logger.debug("{}({}): continuing with {} op".format(self.name, op.name, worker_op.name))
+
+        @pipeline_thread.runs_on_pipeline_thread
+        def worker_op_complete(new_op, error):
+            logger.debug(
+                "{}({}): completing with result from {}".format(self.name, op.name, new_op.name)
+            )
+            self._complete_op(op, error=error)
+
+        worker_op.callback = worker_op_complete
+        self._send_op_down(worker_op)
 
     @pipeline_thread.runs_on_pipeline_thread
-    def new_op_complete(op):
-        logger.debug(
-            "{}({}): completing with result from {}".format(
-                stage.name, original_op.name, new_op.name
+    def _send_op_down(self, op):
+        """
+        Helper function to continue a given operation by passing it to the next stage
+        in the pipeline.  If there is no next stage in the pipeline, this function
+        will fail the operation and call _complete_op to return the failure back up the
+        pipeline.
+
+        :param PipelineOperation op: Operation which is being passed on
+        """
+        if not self.next:
+            logger.error("{}({}): no next stage.  completing with error".format(self.name, op.name))
+            error = PipelineError(
+                "{} not handled after {} stage with no next stage".format(op.name, self.name)
             )
-        )
-        original_op.error = new_op.error
-        complete_op(stage, original_op)
+            self._complete_op(op, error=error)
+        else:
+            logger.debug("{}({}): passing to next stage.".format(self.name, op.name))
+            self.next.run_op(op)
 
-    new_op.callback = new_op_complete
-    pass_op_to_next_stage(stage, new_op)
+    @pipeline_thread.runs_on_pipeline_thread
+    def _complete_op(self, op, error=None):
+        """
+        Helper function to complete an operation by calling its callback function thus
+        returning the result of the operation back up the pipeline.  This is perferred to
+        calling the operation's callback directly as it provides several layers of protection
+        (such as a try/except wrapper) which are strongly advised.
+        """
+        if error:
+            logger.error("{}({}): completing with error {}".format(self.name, op.name, error))
+        else:
+            logger.debug("{}({}): completing without error".format(self.name, op.name))
 
+        if op.completed:
+            logger.error(
+                "{}({}): completing op that has already been completed!".format(self.name, op.name)
+            )
+            e = PipelineError(
+                "Internal pipeline error: attempting to complete an already-completed operation: {}({})".format(
+                    self.name, op.name
+                )
+            )
+            handle_exceptions.handle_background_exception(e)
+        else:
+            op.completed = True
+            self._send_completed_op_up(op, error)
 
-@pipeline_thread.runs_on_pipeline_thread
-def pass_op_to_next_stage(stage, op):
-    """
-    Helper function to continue a given operation by passing it to the next stage
-    in the pipeline.  If there is no next stage in the pipeline, this function
-    will fail the operation and call complete_op to return the failure back up the
-    pipeline.  If the operation is already in an error state, this function will
-    complete the operation in order to return that error to the caller.
+    @pipeline_thread.runs_on_pipeline_thread
+    def _send_completed_op_up(self, op, error=None):
+        if not op.completed:
+            raise PipelineError(
+                "Internal pipeline error: attempting to send an incomplete {} op up".format(op.name)
+            )
 
-    :param PipelineStage stage: stage that the operation is being passed from
-    :param PipelineOperation op: Operation which is being passed on
-    """
-    if op.error:
-        logger.error("{}({}): op has error.  completing.".format(stage.name, op.name))
-        complete_op(stage, op)
-    elif not stage.next:
-        logger.error("{}({}): no next stage.  completing with error".format(stage.name, op.name))
-        op.error = PipelineError(
-            "{} not handled after {} stage with no next stage".format(op.name, stage.name)
-        )
-        complete_op(stage, op)
-    else:
-        logger.debug("{}({}): passing to next stage.".format(stage.name, op.name))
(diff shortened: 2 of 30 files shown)


azure-iot-device/azure/iot/device/common/chainable_exception.py
```diff
@@ -14,3 +14,8 @@ def __init__(self, message=None, cause=None):
         # while still being able to operate in Python 2.
         self.__cause__ = cause
         super(ChainableException, self).__init__(message)
+
+    def __repr__(self):
+        return "{} caused by {}".format(
+            super(ChainableException, self).__repr__(), self.__cause__.__repr__()
+        )
```
cartertinney: Interesting. Don't know when this would actually be used (since the `__str___` inherited from Exception would take precedence), but for any situation where `__repr__` was invoked this would be helpful.

olivakar:
> # --------------------------------------------------------------------------------------------

should this file be renamed ? I understand its not necessary in python

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py:1 in ee1fa90. [](commit_id = ee1fa90dbfc5fe46d19bcbc687208abf965f313e, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -15,124 +15,145 @@
...
+        else:
+            op.completed = True
+            self._send_completed_op_up(op, error)
 
     @pipeline_thread.runs_on_pipeline_thread
-    def new_op_complete(op):
-        logger.debug(
-            "{}({}): completing with result from {}".format(
-                stage.name, original_op.name, new_op.name
+    def _send_completed_op_up(self, op, error=None):
```
cartertinney: Can I get doc for this function?

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -15,124 +15,145 @@
...
+            self.previous.handle_pipeline_event(event)
+        else:
+            logger.error("{}({}): Error: unhandled event".format(self.name, event.name))
+            error = PipelineError(
+                "{} unhandled at {} stage with no previous stage".format(event.name, self.name)
+            )
+            handle_exceptions.handle_background_exception(error)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _send_op_down_and_intercept_return(self, op, intercept_return):
```
cartertinney: Also doc for this

cartertinney:
> # --------------------------------------------------------------------------------------------

+1

---
In reply to: [540736851](https://github.com/Azure/azure-iot-sdk-python/pull/347#issuecomment-540736851) [](ancestors = 540736851)

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py:1 in ee1fa90. [](commit_id = ee1fa90dbfc5fe46d19bcbc687208abf965f313e, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -15,124 +15,145 @@
...
+        :param PipelineOperation worker_op: Operation that is being passed down the pipeline
+          to effectively continue the work represented by the original op.  This is most likely
+          a different type of operation that is able to accomplish the intention of the
+          original op in a way that is more specific than the original op.
+        """
+
+        logger.debug("{}({}): continuing with {} op".format(self.name, op.name, worker_op.name))
+
+        @pipeline_thread.runs_on_pipeline_thread
+        def worker_op_complete(new_op, error):
```
cartertinney: I think for consistency we should refer to this variable as worker_op

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -45,7 +46,20 @@ def __init__(self, callback=None):
         self.name = self.__class__.__name__
         self.callback = callback
         self.needs_connection = False
-        self.error = None
+        self.completed = False
+
+    if "pytest" in sys.modules:
```
cartertinney: >if "pytest" in sys.modules: [](start = 4, length = 27)

My understanding is that this is only necessary until unittests are fixed?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -129,7 +143,7 @@ class DisableFeatureOperation(PipelineOperation):
     Even though this is an base operation, it will most likely be handled by a more specific stage (such as an IoTHub or MQTT stage).
     """
 
-    def __init__(self, feature_name, callback=None):
+    def __init__(self, feature_name, callback):
```
cartertinney: >callback [](start = 37, length = 8)

Good call. Optional callbacks make testing a pain, and there's no reason for it. This is a pattern I introduced that I regret.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -7,19 +7,24 @@
...
 from . import pipeline_thread
 from azure.iot.device.common import handle_exceptions
+from azure.iot.device.common.callable_weak_method import CallableWeakMethod
 
 logger = logging.getLogger(__name__)
 
 
 @six.add_metaclass(abc.ABCMeta)
-class PipelineStage(object):
+class PipelineStage(PipelineFlow):
```
cartertinney: >PipelineFlow [](start = 20, length = 12)

This is a very interesting design choice that I had never really considered.

I don't have an opinion on it (yet), but this is some really out of the box thinking here.

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
cartertinney: Also I'm not sure if "error" is the right semantics here for this arg. We've discussed this before, but I think it's more relevant now - are these truly errors? After all, especially as we get more and more robust in the pipeline, we're definitely throwing exceptions around that are recoverable. Retry for instance.

That being said, "error" is referring to the status of the operation itself, so, I'm in favor of there being no change here, provided that an exception (even a recoverable one) that is attached to an op, always indicates failure of an op.

That is to say:

1) An exception occurs which can be recovered from (e.g. can't connect, will retry)
2) This fails the current operation
3) Another operation is spawned in order to recover (e.g. retry the connect again)

If that's always true, then I still consider this to be an error in the context of an operation. However if an operation can have an "error" set, that is recoverable without failure completion of operation, then I think the semantics are wrong.

That all said, it seems to me that everything is okay, and the semantics ARE correct. Just wanted to double check that my interpretation of how this is supposed to work is all correct.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -346,43 +347,40 @@ def _execute_op(self, op):
...
                     "{}({}): unblock is complete.  completing op that caused unblock".format(
                         self.name, op.name
                     )
                 )
-                operation_flow.complete_op(stage=self, op=op)
+                self._send_completed_op_up(op, error)
 
-            op.callback = on_operation_complete
-            operation_flow.pass_op_to_next_stage(stage=self, op=op)
+            self._send_op_down_and_intercept_return(op, intercept_return=on_operation_complete)
```
cartertinney: Nit: using mixed positional/keyword args implies that the keyword args are optional. In this case they're both required, so they should either both be positional or both be keyword (I'm partial to keyword here for clarity).

This is not that big a deal though.

azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -110,9 +110,6 @@ def _on_disconnected():
 
         self._pipeline.run_op(op)
         callback.wait_for_completion()
-        if op.error:
-            logger.error("{} failed: {}".format(op.name, op.error))
-            raise op.error
```
cartertinney: Good catch, this is totally dead code.

cartertinney:
>         This callback should have one parameter, which will contain the requested twin when called.

two

---
Refers to: azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py:230 in ee1fa90. [](commit_id = ee1fa90dbfc5fe46d19bcbc687208abf965f313e, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -40,34 +35,34 @@ def _execute_op(self, op):
...
-                new_op=pipeline_ops_iothub.SetIoTHubConnectionArgsOperation(
+            self._send_worker_op_down(
+                worker_op=pipeline_ops_iothub.SetIoTHubConnectionArgsOperation(
                     device_id=self.auth_provider.device_id,
                     module_id=getattr(self.auth_provider, "module_id", None),
                     hostname=self.auth_provider.hostname,
                     gateway_hostname=getattr(self.auth_provider, "gateway_hostname", None),
                     ca_cert=getattr(self.auth_provider, "ca_cert", None),
                     sas_token=self.auth_provider.get_current_sas_token(),
+                    callback=op.callback,
```
cartertinney: >callback=op.callback, [](start = 20, length = 21)

This is kind of confusing, because this callback you're setting immediately gets overridden in the implementation of the `_send_worker_op_down` method.

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
```diff
@@ -15,124 +15,145 @@
...
-      effectively paused while we wait for the new_op operation to complete.  When
-      the new_op operation completes, the original_op operation will also be completed.
-    :param PipelineOperation new_op: Operation that is being passed down the pipeline
-      to effectively continue the work represented by original_op.  This is most likely
-      a different type of operation that is able to accomplish the intention of the
-      original_op in a way that is more specific than the original_op.
-    """
-
-    logger.debug("{}({}): continuing with {} op".format(stage.name, original_op.name, new_op.name))
+class PipelineFlow(object):
```
cartertinney: Not sure this is an ideal name - not really a noun.

azure-iot-device/tests/common/pipeline/test_operation_flow.py
```diff
@@ -50,11 +45,11 @@ def stage(mocker, arbitrary_exception, arbitrary_base_exception):
     )
 
 
-@pytest.mark.describe("delegate_to_different_op()")
+@pytest.mark.describe("_send_worker_op_down()")
```
cartertinney: As methods, these should have the class in the description

`OperationFlow - .send_worker_op_down()`

azure-iot-device/tests/common/pipeline/test_operation_flow.py
```diff
@@ -50,11 +45,11 @@ def stage(mocker, arbitrary_exception, arbitrary_base_exception):
...
 class TestContineWithDifferntOp(object):
```
cartertinney: test class needs to be renamed

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Really?  OK.  This idea came from you.  You threw this at me before operation_flow.py even got pulled out of pipeline_stages_base.py.  I thinking you were talking about _inheritance_ at the time, and I think about this in terms of _composition_, and maybe that's why you don't recognize it.  

Also, this is the same pattern (sort-of) as you use in SharedClientInstantiationTests.  I've moved to using a similar pattern in Horton, only to a much larger extreme.  My objects for modules and devices are just compositions of smaller objects for features like twins, methods, telemetry, etc.  I do this for tests themselves, for rest wrappers, and for glue.  It gives me huge flexibility for sharing code between clients.   (See here for the best example: [https://github.com/Azure/iot-sdks-e2e-fx/blob/af15021fdf560096f326a563052308aa1d6d3bb0/test-runner/adapters/rest/rest_iothub_apis.py#L327](https://github.com/Azure/iot-sdks-e2e-fx/blob/af15021fdf560096f326a563052308aa1d6d3bb0/test-runner/adapters/rest/rest_iothub_apis.py#L327)) 

---
In reply to: [333702205](https://github.com/Azure/azure-iot-sdk-python/pull/347#discussion_r333702205) [](ancestors = 333702205)

cartertinney: Hm. Maybe. I don't recall. 

---
In reply to: [333730282](https://github.com/Azure/azure-iot-sdk-python/pull/347#discussion_r333730282) [](ancestors = 333730282,333702205)

azure-iot-device/tests/common/pipeline/test_operation_flow.py
```diff
@@ -63,58 +58,48 @@ def test_completes_original_op_after_new_op_completes(self, stage, op, new_op, c
...
-        assert_callback_failed(op=op, error=arbitrary_exception)
-        assert stage.next.run_op.call_count == 0
-
     @pytest.mark.it("Fails the op if there is no next stage")
     def test_fails_op_when_no_next_stage(self, stage, op, callback):
         op.callback = callback
         stage.next = None
-        pass_op_to_next_stage(stage, op)
+        stage._send_op_down(op)
         assert_callback_failed(op=op)
```
cartertinney: This should also assert the correct type of error (in this case `PipelineError`)

cartertinney:
> # license information.

No tests for `._send_event_up()` or `._send_op_down_and_intercept_return()`?

---
Refers to: azure-iot-device/tests/common/pipeline/test_operation_flow.py:4 in ee1fa90. [](commit_id = ee1fa90dbfc5fe46d19bcbc687208abf965f313e, deletion_comment = False)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -46,6 +46,10 @@ def apply_fake_pipeline_thread(fake_pipeline_thread):
     pass
 
 
+class FakeOperation(pipeline_ops_base.PipelineOperation):
+    pass
```
cartertinney: Use the one defined in the conftest.py instead. The fixture is "op" (maybe should be renamed)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -248,9 +253,11 @@ def test_sets_pending_operation(self, stage, create_transport, op_connect):
     @pytest.mark.parametrize(
         "pending_connection_op",
         [
-            pytest.param(pipeline_ops_base.ConnectOperation(), id="Pending ConnectOperation"),
-            pytest.param(pipeline_ops_base.ReconnectOperation(), id="Pending ReconnectOperation"),
-            pytest.param(pipeline_ops_base.DisconnectOperation(), id="Pending DisconnectOperation"),
+            pytest.param(pipeline_ops_base.ConnectOperation(1), id="Pending ConnectOperation"),
+            pytest.param(pipeline_ops_base.ReconnectOperation(1), id="Pending ReconnectOperation"),
+            pytest.param(
+                pipeline_ops_base.DisconnectOperation(1), id="Pending DisconnectOperation"
```
cartertinney: >1 [](start = 54, length = 1)

This is an unpleasant side effect of the callback being required - and we can't use fixtures because of the parametrization (WHY PYTEST WHY?).

I'd probably prefer use of None to use of integer 1, but eh, neither is great.

cartertinney:
Signed off - There's some slight design and syntactical things I'd like to see (plus coverage for those missing tests with `PipelineFlow`) but otherwise this is ready to be merged.

azure-iot-device/azure/iot/device/common/chainable_exception.py
BertKleewein: The default __str__ calls __repr__.  I think I did this because some of Logging() code uses __repr__ and I wanted to see this in logger.error() output.

---
In reply to: [333675697](https://github.com/Azure/azure-iot-sdk-python/pull/347#discussion_r333675697) [](ancestors = 333675697)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
BertKleewein: You know what, you're right.  I was using this to make sure everything was changed, and now that all the code is using the new pattern, I'm going to remove this.  

---
In reply to: [333700461](https://github.com/Azure/azure-iot-sdk-python/pull/347#discussion_r333700461) [](ancestors = 333700461)

azure-iot-device/azure/iot/device/common/pipeline/operation_flow.py
BertKleewein: Noun, schmoun.  Flow is a noun, as in "this is how the flow through the pipeline happens".

---
In reply to: [333724482](https://github.com/Azure/azure-iot-sdk-python/pull/347#discussion_r333724482) [](ancestors = 333724482)

PR #349, opened 2019-10-14
docs: Further changes for correct API doc generation

cartertinney:
* Added missing elements of docstrings needed for correct API docs to IoTHub and DPS clients
* Correct exception docs in DPS will be added when exception surfacing is added.
* Removed `protocol_choice` parameter from DPS clients

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/sync_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/sync_clients.py
@@ -51,8 +51,10 @@ class GenericIoTHubClient(AbstractIoTHubClient):
         This initializer should not be called directly.
         Instead, use one of the 'create_from_' classmethods to instantiate
 
-        TODO: How to document kwargs?
-        Possible values: iothub_pipeline, edge_pipeline
+        :param iothub_pipeline: The IoTHubPipeline used for the client
+        :type iothub_pipeline: :class:`azure.iot.device.iothub.pipeline.IoTHubPipeline`
+        :param edge_pipeline: The EdgePipeline used for the client
+        :type edge_pipeline: :class:`azure.iot.device.iothub.pipeline.EdgePipeline`
         """
         # Depending on the subclass calling this __init__, there could be different arguments,
         # and the super() call could call a different class, due to the different MROs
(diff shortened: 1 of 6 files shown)


azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -51,8 +51,10 @@ def __init__(self, **kwargs):
         This initializer should not be called directly.
         Instead, use one of the 'create_from_' classmethods to instantiate
 
-        TODO: How to document kwargs?
-        Possible values: iothub_pipeline, edge_pipeline
+        :param iothub_pipeline: The IoTHubPipeline used for the client
+        :type iothub_pipeline: :class:`azure.iot.device.iothub.pipeline.IoTHubPipeline`
+        :param edge_pipeline: The EdgePipeline used for the client
```
BertKleewein: >edge_pipeline [](start = 15, length = 13)

Not part of this review, but I really dislike this name.  Having the 2 pipeline args makes it look like you chose one pipeline if you're connecting to iothub and a different pipeline if you're connecting to edgehub.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -58,6 +61,10 @@ def register(self):
 
     def cancel(self):
```
BertKleewein: >cancel [](start = 8, length = 6)

Out of scope, but we should consider the need to include this API before we GA.  I know that we have this in the Node SDK, but, AFAIK, this is the only operation that we have in the entire Python SDK that is cancelable.  I would question the utility of this function.  If we do need it, we need to make sure we know why, and we need to make sure it's bulletproof

azure-iot-device/azure/iot/device/iothub/sync_clients.py
cartertinney: I also dislike this name, but have yet to come up with something better. Open to suggestion.

---
In reply to: [334942864](https://github.com/Azure/azure-iot-sdk-python/pull/349#discussion_r334942864) [](ancestors = 334942864)

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
cartertinney: I agree

---
In reply to: [334947310](https://github.com/Azure/azure-iot-sdk-python/pull/349#discussion_r334947310) [](ancestors = 334947310)

PR #350, opened 2019-10-15
chore: Clean up pipeline fixtures

BertKleewein:


olivakar:
> def _test_pipeline_root_runs_callback_in_callback_thread(self, stage, mocker):

I never noticed these..any reason these names start with an underscore ?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:65 in b095722. [](commit_id = b095722b90b77782d78aacc3614df3c1bc8a5581, deletion_comment = False)

BertKleewein:
> def _test_pipeline_root_runs_callback_in_callback_thread(self, stage, mocker):

It's a bad pattern that I used and I hope to remove.  These are actually methods, but I'm defining them as functions (with an underscore so they don't run), then assigning the underscored function as a method on line 147 so it can run in the context of that class.  Horrible idea.  

---
In reply to: [542369476](https://github.com/Azure/azure-iot-sdk-python/pull/350#issuecomment-542369476) [](ancestors = 542369476)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:65 in b095722. [](commit_id = b095722b90b77782d78aacc3614df3c1bc8a5581, deletion_comment = False)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -68,9 +68,9 @@ def add_unknown_ops_tests(cls, module, all_ops, handled_ops):
     @pytest.mark.describe("{} - .run_op() -- unknown and unhandled operations".format(cls.__name__))
     class LocalTestObject(object):
         @pytest.fixture
-        def op(self, op_cls, callback):
+        def op(self, op_cls, mocker):
             op = make_mock_op_or_event(op_cls)
-            op.callback = callback
+            op.callback = mocker.MagicMock()
```
cartertinney: >mocker.MagicMock() [](start = 26, length = 18)

Good change. This has been something necessary for a while.

azure-iot-device/tests/common/pipeline/test_pipeline_flow.py
```diff
@@ -18,6 +18,7 @@
     assert_callback_failed,
     assert_callback_succeeded,
 )
+from .fixtures import FakeOperation
```
cartertinney: >FakeOperation [](start = 22, length = 13)

Lately I've been trying to keep class definitions used for fixtures to be internally defined within the fixture itself, and not pass around the class definition.

However, here, given that there are multiple op fixtures, including ones defined exclusively for this file (e.g. arbitrary_worker_op), I see why this is necessary

azure-iot-device/tests/common/pipeline/test_pipeline_flow.py
```diff
@@ -49,188 +50,206 @@ def stage(mocker, arbitrary_exception, arbitrary_base_exception):
 
 @pytest.mark.describe("PipelineFlow - ._send_worker_op_down()")
 class TestSendWorkerOpDown(object):
+    @pytest.fixture
+    def arbitrary_worker_op(self, mocker):
+        op = FakeOperation(callback=mocker.MagicMock())
+        op.name = "arbitrary_worker_op"
+        return op
+
     @pytest.mark.it("Runs the new op and does not continue running the original op")
```
cartertinney: > new op [](start = 29, length = 7)

worker op

azure-iot-device/tests/common/pipeline/test_pipeline_flow.py
```diff
@@ -49,188 +50,206 @@ def stage(mocker, arbitrary_exception, arbitrary_base_exception):
...
-    def test_runs_new_op(self, mocker, stage, op, new_op):
-        stage._send_worker_op_down(worker_op=new_op, op=op)
+    def test_runs_worker_op_op(self, mocker, stage, arbitrary_op, arbitrary_worker_op):
```
cartertinney: >op_op [](start = 25, length = 5)

stutter

azure-iot-device/tests/common/pipeline/fixtures.py
```diff
@@ -14,25 +14,13 @@
 )
 
 
-# TODO: remove this fixture
```
cartertinney: Hooray

PR #351, opened 2019-10-15
enhancement(azure-iot-device): Custom Payload for DPS

olivakar:
DPS accepting custom payload.

If payload is None, it must send just the registration_id.
Otherwise it should send a combination of registration_id and payload.

### DIFF
--- a/azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
+++ b/azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
@@ -34,9 +34,7 @@ class AbstractProvisioningDeviceClient(object):
         self._provisioning_pipeline = provisioning_pipeline
 
     @classmethod
-    def create_from_symmetric_key(
-        cls, provisioning_host, registration_id, id_scope, symmetric_key, protocol_choice=None
-    ):
+    def create_from_symmetric_key(cls, provisioning_host, registration_id, id_scope, symmetric_key):
         """
         Create a client which can be used to run the registration of a device with provisioning service
         using Symmetric Key authentication.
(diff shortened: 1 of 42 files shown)


azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
```diff
@@ -103,6 +104,15 @@ def cancel(self):
         """
         pass
 
+    def set_provisioning_payload(self, custom_payload):
```
cartertinney: >set_provisioning_payload [](start = 8, length = 24)

Use a @property decorator instead.

Or better yet, just let them set the attribute directly

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning_mqtt.py
```diff
@@ -143,3 +151,18 @@ def _handle_pipeline_event(self, event):
...
+    The class representing the payload that needs to be sent to the service.
+    """
+
+    def __init__(self, registration_id, custom_payload=None):
+        # This is not a convention to name variables in python but the
+        # DPS service spec needs the name to be exact for it to work
+        self.registrationId = registration_id
+        self.payload = custom_payload
+
+    def get_json(self):
```
cartertinney: >get_json [](start = 8, length = 8)

get_json_string

azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
olivakar: its now a property

---
In reply to: [335731680](https://github.com/Azure/azure-iot-sdk-python/pull/351#discussion_r335731680) [](ancestors = 335731680)

PR #361, opened 2019-10-18
feat(azure-iot-device): add payload in response for provisioning

olivakar:


azure-iot-device/azure/iot/device/provisioning/models/registration_result.py
```diff
@@ -111,5 +114,10 @@ def last_update_date_time(self):
     def etag(self):
         return self._etag
 
+    @property
+    def response_payload(self):
+        payload = "" if self._response_payload is None else self._response_payload
+        return payload
```
cartertinney: Any particular reason we use empty string over a None value? Is it just for the ease of text formatting?

PR #368, opened 2019-10-22
feat(azure-iot-device): Custom User Agent / product_info option 

YoDaMa:
This implements the custom user agent for MQTT for azure iot hub as defined in the Azure IoT Hub User Agent Spec. The User Agent is usually a meta-string defined in HTTP packets for denoting the browser and other environment related information when sending an HTTP packet.


azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -55,11 +55,17 @@ def _execute_op(self, op):
...
-            username = "{hostname}/{client_id}/?{query_params}".format(
-                hostname=op.hostname,
-                client_id=client_id,
-                query_params=urllib.parse.urlencode(query_param_seq),
-            )
+            # According to the Azure IoT Hub User Agent Spec for MQTT:
+            # Append entire string to username field in CONNECT packet
+            # The spec also defines defines a custom user agent that the user can set.
+            if self.pipeline_root.pipeline_configuration.product_info:
+                username = self.pipeline_root.pipeline_configuration.product_info
```
cartertinney: >username = self.pipeline_root.pipeline_configuration.product_info [](start = 16, length = 65)

This seems to replace the username, not append the `product_info` string to it.

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,29 @@
...
+logger = logging.getLogger(__name__)
+
+
+class IoTHubPipelineConfig(BasePipelineConfig):
+    """A base class for storing all configurations/options shared across the Azure IoT Python Device Client Library.
+    More specific configurations such as those that only apply to the IoT Hub Client will be found in the respective
+    config files.
+    """
+
+    super
```
cartertinney: >super [](start = 4, length = 5)

?

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self, product_info=None, **kwargs):
+        """Initializer for IoTHubPipelineConfig which passes all unrecognized keyword-args down to BasePipelineConfig
+        to be evaluated. This stacked options setting is to allow for unique configuration options to exist between the
+        IoTHub Client and the Provisioning Client, while maintaining a base configuration class with shared config options.
+
+        :param string product_info: A custom identification string for the type of device connecting to Azure IoT Hub.
+        """
+        super(IoTHubPipelineConfig, self).__init__(**kwargs)
+        self.product_info = product_info
```
cartertinney: Why is this called `product_info` rather than `user_agent` or something

azure-iot-device/azure/iot/device/iothub/config.py
```diff
@@ -0,0 +1,29 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
```
cartertinney: Still think this should be in the pipeline package

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
YoDaMa: resolved in a rework.

---
In reply to: [337671382](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r337671382) [](ancestors = 337671382)

azure-iot-device/azure/iot/device/iothub/config.py
YoDaMa: User Agent defines the actual user agent that we are generating for the SDK. `custom_user_agent` could be a good name, but the powers that be have decided that it should be called Product Info. It is that way in C, `product_info`, and in node, `productInfo`.

---
In reply to: [337672739](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r337672739) [](ancestors = 337672739)

azure-iot-device/azure/iot/device/iothub/config.py
YoDaMa: removed.

---
In reply to: [337672538](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r337672538) [](ancestors = 337672538)

azure-iot-device/azure/iot/device/iothub/config.py
YoDaMa: moved to pipeline

---
In reply to: [337672921](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r337672921) [](ancestors = 337672921)

azure-iot-device/azure/iot/device/iothub/pipeline/config.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self, product_info="", **kwargs):
+        """Initializer for IoTHubPipelineConfig which passes all unrecognized keyword-args down to BasePipelineConfig
+        to be evaluated. This stacked options setting is to allow for unique configuration options to exist between the
+        IoTHub Client and the Provisioning Client, while maintaining a base configuration class with shared config options.
+
+        :param str product_info: A custom identification string for the type of device connecting to Azure IoT Hub.
+        """
+        super(IoTHubPipelineConfig, self).__init__(**kwargs)
+        self.product_info = product_info
```
BertKleewein: >product_info [](start = 13, length = 12)

Are you sure we don't need this for the provisioning client?

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -53,7 +53,9 @@ def create_from_connection_string(cls, connection_string, ca_cert=None, **kwargs
         :param str connection_string: The connection string for the IoTHub you wish to connect to.
         :param str ca_cert: The trusted certificate chain. Only necessary when using a
             connection string with a GatewayHostName parameter.
-        :param bool websockets: Default is False. Set to true if using MQTT over websockets.
+
+        :param bool websockets: Configuration Option. Default is False. Set to true if using MQTT over websockets.
+        :param str product_info: Configuration Option. Default is empty string. Custom product info to be appended to user agent string.
```
BertKleewein: >Custom product info to be appended to user agent string. [](start = 80, length = 56)

This sounds like a sentence fragment.  Suggest something like "This string contains arbitrary product info which is appended to user agent string." or something like that.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -51,14 +51,19 @@ def _execute_op(self, op):
             else:
                 client_id = op.device_id
 
+            # For MQTT, the entire user agent string should be appended to the username field in the connect packet
+            # For example, the username may look like this without custom parameters:
+            # yosephsandboxhub.azure-devices.net/alpha/?api-version=2018-06-30&DeviceClientType=py-azure-iot-device%2F2.0.0-preview.12
+            # The customer user agent string would simply be appended to the end of this username, in URL Encoded format.
```
BertKleewein: >in URL Encoded format. [](start = 99, length = 22)

Maybe add the note about URL-encoding to the user doc for this parameter.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -51,14 +51,19 @@ def _execute_op(self, op):
...
             query_param_seq = [
                 ("api-version", pkg_constant.IOTHUB_API_VERSION),
                 ("DeviceClientType", pkg_constant.USER_AGENT),
             ]
-            username = "{hostname}/{client_id}/?{query_params}".format(
+            username = "{hostname}/{client_id}/?{query_params}{optional_product_info}".format(
```
BertKleewein: >username = "{hostname}/{client_id}/?{query_params}{optional_product_info}".format( [](start = 12, length = 82)

What, no tests?  You need at least 1 test for a string that doesn't need URI encoding and a different test for a string that does need URI encoding.  If we don't have a test that validates the username without optional_product_info, we should have that, but I assume that is already there.  And, just for giggles, since this is user input, I might add tests for a few different types which aren't strings.  (integers at least, but I'm not sure about dictionary or list).  

And, now that I look at it, I don't see any URI-encoding in here.  I assume it's our responsibility to encode it?  Or does the user pass a string which is already encoded?  Either way, the responsibly party should be documented in create_from_connection_string, etc.

YoDaMa: This is a bug. Fixed this and will push soon. 

---
In reply to: [338265615](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338265615) [](ancestors = 338265615)

YoDaMa: As for tests: yes no tests yet for the custom param yet. It's passing all our current tests without the product info stuff, but I'll be adding tests for validation soon. 

---
In reply to: [338272541](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338272541) [](ancestors = 338272541,338265615)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
YoDaMa: changed the sentence.

---
In reply to: [338260330](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338260330) [](ancestors = 338260330)

azure-iot-device/azure/iot/device/iothub/pipeline/config.py
YoDaMa: yes.

---
In reply to: [338259006](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338259006) [](ancestors = 338259006)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
YoDaMa: I've added documentation in the config.py stage, and in abstract_clients.py

---
In reply to: [338261318](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338261318) [](ancestors = 338261318)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
YoDaMa: So I've added tests for test_pipeline_stages_iothub_mqtt now. Those tests manually set the pipeline_root configuration product_info to multiple values and validate that they are appended correctly after being url encoded. Those values include: empty string, normal string, array, dict, and number. All non-string values are cast to their string representation before being URL Encoded because that is how the code I've written handles non-string user input values.

---
In reply to: [338272889](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r338272889) [](ancestors = 338272889,338272541,338265615)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -226,6 +227,42 @@ def set_connection_args_for_module(set_connection_args):
 
 
 class IoTHubMQTTConverterStageTestBase(StageTestBase):
+
+    # This is initially defined in helpers.py under the common pipeline tests,
+    # However that definition is for IoTHub specifically, because we need to
+    # make sure that root is configured using the IoTHubPipelineConfig.
+    # YMTODO: The comment below states that it's not guaranteed to be called
+    # before other fixtures run. Because of that, maybe I need to do something different...
```
BertKleewein: This doesn't matter. 
I think that "not guaranteed" comment is a little esoteric because it's much more specific than I had thought at the time.  Because this fixture is autouse, it gets applied before any fixtures that are not autouse.  However, we can't guarantee that it gets applied before any other autouse fixtures, but I don't think that's going to be a problem in this particular case.    I would remove this comment and the one below.  (and the source comment too).

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -353,6 +390,40 @@ def test_sets_module_username_if_yes_gateway(self, stage, set_connection_args_fo
...
         )
 
+    @pytest.mark.it(
+        "Appends product_info to connection_args.username to if self.pipeline_root.pipeline_configuration.product_info is not None"
+    )
+    @pytest.mark.parametrize(
+        "fake_product_info, expected_product_info",
+        [
+            ("", ""),
+            ("__fake_product_info__", "__fake_product_info__"),
```
BertKleewein: >("__fake_product_info__", "__fake_product_info__"), [](start = 12, length = 51)

What about just a string that needs encoding, like "fake,product,info", "fake%2Cproduct%2Cinfo"?    (I don't know why I'm so stuck on URI encoding).

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -69,6 +69,72 @@ def test_sets_on_method_request_received_handler_in_pipleline(
...
+        assert mock_pipeline_init.call_args[0][1].product_info == ""
+
+    @pytest.mark.it("Sets all valid configuration options to the user supplied values")
+    @pytest.mark.parametrize(
+        "websockets, product_info",
+        [
+            pytest.param((None, None), (None, None), id=" Setting to None"),
+            pytest.param(
+                (True, True),
+                ("__fake_product_info__", "__fake_product_info__"),
```
BertKleewein: >("__fake_product_info__", "__fake_product_info__"), [](start = 16, length = 51)

All this parametrization is overly complex IMHO.

BertKleewein: But I wouldn't change it.

---
In reply to: [340337876](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r340337876) [](ancestors = 340337876)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
YoDaMa: I can add another one. 

---
In reply to: [340334859](https://github.com/Azure/azure-iot-sdk-python/pull/368#discussion_r340334859) [](ancestors = 340334859)

PR #369, opened 2019-10-23
refactor(azure-iot-device): Pipeline Refactor (Part 1)

cartertinney:
The first chunk of my changes to simplify the pipeline tests.

* `PipelineFlow` class was folded into `PipelineStage`, and the methods are now no longer convention-private
* Tests for the formerly `PipelineFlow` methods are now applied to every stage
* Added base classes for `.run_op(`) and .handle_pipeline_event() test classes
    * Note that these are only currently used for the shared tests defined in `pipeline_stage_test.py`, and will be applied to stage specific tests in a follow up PR


azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -289,3 +215,280 @@ def test_enters_pipeline(self, mocker, stage, method_name, fake_non_pipeline_thr
                 assert pipeline_executor.submit.call_count == 1
 
     setattr(module, "Test{}PipelineThreading".format(cls.__name__), LocalTestObject)
+
+
+def _add_pipeline_flow_tests(cls, module):
+    class PipelineFlowTestBase(object):
```
BertKleewein: >PipelineFlowTestBase [](start = 10, length = 20)

I assume this base class is  here because this is a WIP.  otherwise it doesn't make much sense.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -70,10 +76,10 @@ def add_unknown_ops_tests(cls, module, all_ops, handled_ops):
     unknown_ops = all_except(all_items=all_ops, items_to_exclude=handled_ops)
 
     @pytest.mark.describe("{} - .run_op() -- unknown and unhandled operations".format(cls.__name__))
-    class LocalTestObject(StageTestBase):
-        @pytest.fixture
-        def op(self, op_cls, mocker):
-            op = make_mock_op_or_event(op_cls)
+    class LocalTestObject(StageRunOpTestBase):
```
BertKleewein: >StageRunOpTestBase [](start = 26, length = 18)

You may know this, but this is a hidden gotcha with this style of testing that you might not have seen before.  StageRunOpBaseTests tests will run only with ops in the unkonwn_op array because that's how the op fixture is defined in this class.  I think the coverage is OK for now, better than it was before, but it's something to keep in mind.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: I use it to provide the stage fixture. If you have a better idea I'm interested to hear it

---
In reply to: [338066181](https://github.com/Azure/azure-iot-sdk-python/pull/369#discussion_r338066181) [](ancestors = 338066181)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: Yup. That's why in the next step of the revisions StageRunOpTestBAse, this base class needs to be extended to be used across all other .run_op() tests in all testing modules.

---
In reply to: [338069969](https://github.com/Azure/azure-iot-sdk-python/pull/369#discussion_r338069969) [](ancestors = 338069969)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
BertKleewein: Duh, never mind.  I missed the other classes below.  

---
In reply to: [338198894](https://github.com/Azure/azure-iot-sdk-python/pull/369#discussion_r338198894) [](ancestors = 338198894,338066181)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
```diff
@@ -16,9 +16,9 @@
 @pytest.mark.describe("PipelineOperation")
 class TestPipelineOperation(object):
     @pytest.mark.it("Can't be instantiated")
-    def test_instantiate(self):
+    def test_instantiate(self, mocker):
         with pytest.raises(TypeError):
-            pipeline_ops_base.PipelineOperation()
+            pipeline_ops_base.PipelineOperation(mocker.MagicMock())
```
BertKleewein: >mocker [](start = 48, length = 6)

oh damn.  I can't believe I missed this.

azure-iot-device/tests/iothub/pipeline/conftest.py
```diff
@@ -8,4 +8,6 @@
     fake_pipeline_thread,
     fake_non_pipeline_thread,
     unhandled_error_handler,
+    arbitrary_op,
+    arbitrary_event,
```
BertKleewein: do these actually need to be here?  (and in tests/provisioning/pipeline)  I assume this is bleed through from another change that didn't make it into this PR, but I wouldn't rule out some weird fixture thing causing this to suddenly be necessary.

cartertinney: I think this is bleedthrough yeah - right now I believe these are only needed in common tests, but they will be needed here shortly as well (I reverted some in-progress changes in this package in order to get this PR completed).

---
In reply to: [338219175](https://github.com/Azure/azure-iot-sdk-python/pull/369#discussion_r338219175) [](ancestors = 338219175)

PR #379, opened 2019-10-30
fix(azure-iot-device): reconnect on dropped connections

BertKleewein:


azure-iot-device/azure/iot/device/common/handle_exceptions.py
```diff
@@ -24,7 +25,8 @@ def handle_background_exception(e):
 
     # @FUTURE: We should add a mechanism which allows applications to receive these
     # exceptions so they can respond accordingly
-    logger.error(msg="Exception caught in background thread.  Unable to handle.", exc_info=e)
+    logger.error(msg="Exception caught in background thread.  Unable to handle.")
+    logger.error(traceback.format_exception_only(type(e), e))
```
BertKleewein: >logger.error(traceback.format_exception_only(type(e), e)) [](start = 4, length = 57)

Very subtle but important difference here.  Horton checks for memory leaks after every test, but the old code was causing something that looked like a memory leak (but wasn't actually a leak per-se).

When you use exc_info=e or exc_info=True, it causes the Exception itself, including the traceback, to be saved by pytest until the test is done.  The problem with tracebacks is that they hold onto the actual objects in the stack, so everything in the stack "leaks" until the test is done.

This "false leak report" by itself could be fixed by better leak detection, but the other side effect (that we don't see yet) is that object lifetime is far longer than it needs to be (but only when running under pytest).  So, if we had a long-run-time test that ran under pytest, any objects associated with any exceptions (even when handled correctly) would not be collected until the end of the run.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_exceptions.py
```diff
@@ -35,6 +29,12 @@ class PipelineTimeoutError(PipelineException):
 
 
 class PipelineError(PipelineException):
-    """Error in Pipeline"""
+    """Error in pipeline"""
+
+    pass
+
+
+class PipelineConfigurationError(PipelineException):
```
BertKleewein: >PipelineConfigurationError [](start = 6, length = 26)

A PipelineConfigurationError is not transient.  a PipelineError _could_ be transient.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -1,4 +1,3 @@
-# --------------------------------------------------------------------------------------------
```
BertKleewein: >-------------------------------------------------------------------------------------------- [](start = 2, length = 92)

Note to self: add back in next iteration

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -378,6 +379,18 @@ def test_client_raises_unexpected_error(
             transport.connect(fake_password)
         assert e_info.value.__cause__ is arbitrary_exception
 
+    @pytest.mark.it(
+        "Raises a ConnectionFailedException if Paho connect raises a socket.error Exception"
```
olivakar: >ConnectionFailedException [](start = 18, length = 25)

should this be a c`onnectionfailederror` or a `connectionfailederror` exception ?

PR #381, opened 2019-11-01
fix: return proper errors for failed connections and prevent exception objects from leaking in pytest runs

BertKleewein:


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -294,6 +298,10 @@ def connect(self, password=None):
                 rc = self._mqtt_client.connect(
                     host=self._hostname, port=8883, keepalive=DEFAULT_KEEPALIVE
                 )
+        except socket.error as e:
```
cartertinney: >socket.error [](start = 15, length = 12)

socket.error is actually just an alias for OSError. I go back and forth as to which one is preferable to catch. I think sticking with socket.error is probably the safest in terms of maintenance.

PR #384, opened 2019-11-04
chore: Added Security MD file to the repo

jebrando:


SECURITY.MD
```diff
@@ -0,0 +1,41 @@
+<!-- BEGIN MICROSOFT SECURITY.MD V0.0.3 BLOCK -->
```
olivakar: Do we need to have additional begin and end blocks ?

It does not provide an additional format style.

jebrando: I don't know, I was scared to remove since we got it from MSRC

PR #386, opened 2019-11-07
chore(azure-iot-provisioning): Implement samples

zolvarga:


azure-iot-provisioning/samples/iot_provisioning_service_client_query_sample.py
```diff
@@ -0,0 +1,72 @@
+# -------------------------------------------------------------------------
```
olivakar: >------------------------------------------------------------------------- [](start = 2, length = 73)

wouldn't it be nice to have each of these queries before and after the replace and delete operations in the individual samples to make sure that things got replace or deleted ?

zolvarga: Yes, we will revisit the logic here.

---
In reply to: [343798654](https://github.com/Azure/azure-iot-sdk-python/pull/386#discussion_r343798654) [](ancestors = 343798654)

PR #389, opened 2019-11-08
fix (azure-iot-device): reconnect transport on dropped connection and connection failure

BertKleewein:


azure-iot-device/azure/iot/device/iothub/pipeline/iothub_pipeline.py
```diff
@@ -56,10 +56,11 @@ def __init__(self, auth_provider, pipeline_configuration):
             .append_stage(pipeline_stages_iothub.TwinRequestResponseStage())
             .append_stage(pipeline_stages_base.CoordinateRequestAndResponseStage())
             .append_stage(pipeline_stages_iothub_mqtt.IoTHubMQTTTranslationStage())
-            .append_stage(pipeline_stages_base.RetryStage())
-            .append_stage(pipeline_stages_base.OpTimeoutStage())
+            .append_stage(pipeline_stages_base.ReconnectStage())
             .append_stage(pipeline_stages_base.AutoConnectStage())
             .append_stage(pipeline_stages_base.ConnectionLockStage())
```
BertKleewein: To answer Yoseph's question from my last PR, this order isn't fully flushed out yet because we don't have a higher level retry in place yet.  The reason I moved this is not to move AutoConnect up, but rather to move Timeout and Retry down.  I want Timeout and Retry to operate at the lowest levels.  If an MQTT publish fails, I only want to retry the MQTT publish.  If retry was above AutoConnect, then a failed publish could lead to an automatic reconnection, and I don't want that.  Eventually, I want that reconnection to happen according to the retry policy, and I want that to happen higher up in the pipeline.  I'm still forming this thought, but I know that it doesn't happen all the way at the bottom of the pipeline.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -17,7 +17,7 @@
 from . import pipeline_ops_base, pipeline_ops_mqtt
 from . import pipeline_thread
 from . import pipeline_exceptions
-from azure.iot.device.common import handle_exceptions
+from azure.iot.device.common import handle_exceptions, transport_exceptions
```
cartertinney: I'm not concerned about this right now, but I think at some point we should revisit some of how and where these exceptions are defined - I'm a bit worried about certain things creeping beyond the scope of where they should be used.

PR #391, opened 2019-11-13
refactor(azure-iot-device): Polling Machine To Pipeline

olivakar:
There are couple of TODOs 
Add a time out stage specifically for DPS. The basic time out stage is defined within common. 
Add a custom handle exception block to the client.
Delete all cancel code if no longer needed.
I have to check all tests.

As far as the request response operations are concerned I added 2 attributes to the normal request and response operations instead of redefining these operations for DPS specifically.


azure-iot-device/azure/iot/device/common/pipeline/pipeline_events_base.py
```diff
@@ -50,17 +50,17 @@ class ResponseEvent(PipelineEvent):
...
+    :type retry_after: int
     """
 
-    def __init__(self, request_id, status_code, response_body):
+    def __init__(self, request_id, status_code, response_body, retry_after=None):
         super(ResponseEvent, self).__init__()
         self.request_id = request_id
         self.status_code = status_code
         self.response_body = response_body
+        # TODO Question : If not here we have to use DPSResponse
```
BertKleewein: ># TODO Question : If not here we have to use DPSResponse [](start = 8, length = 56)

This is OK here.  The concept of "retry_after" is something that exists outside of DPS, so we may need to use it in other situations

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -664,6 +666,8 @@ def _handle_pipeline_event(self, event):
                 del (self.pending_responses[event.request_id])
                 op.status_code = event.status_code
                 op.response_body = event.response_body
+                # DPS Specific Code
```
BertKleewein: ># DPS Specific Code [](start = 16, length = 19)

No need for this comment.  Not _really_ DPS specific :)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -241,10 +267,15 @@ def __init__(self, request_type, method, resource_location, request_body, reques
...
         """
         super(RequestOperation, self).__init__(callback=callback)
         self.method = method
         self.resource_location = resource_location
         self.request_type = request_type
         self.request_body = request_body
         self.request_id = request_id
+        # TODO Question : If not here we have to define new op DPSRequestOperation
+        self.registration_id = registration_id
+        self.operation_id = operation_id
```
BertKleewein: I can see why you need these here, and I can see why they're useful, but they don't belong in this object as you have them today, but I don't think you need to add a DPSRequestOperation object either. I would suggest:

1. registration_id does not affect the topic.  It only affects the payload.  I would rather you create the DeviceRegistrationPayload object in pipeline_starges_provisioing.py instead of pipeline_stages_provisioning_mqtt.py.  That way, you don't have to pass registration_id down in the op.

2. operation_id does affect the topic, and it's like a query parameter on a URL, so I would add query_parameters to RequestOperation.  In pipeline_stages_provisioning, you could set op.query_parameters={"operationId", operation_id} before you pass it down, and in pipeline_stages_provisioning_mqtt.py, you could get the operation_id out of this dict when you build the topic
This way, RequestOperation still carries the info you need, but it's not stored in DPS-specific attributes.

azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
```diff
@@ -17,11 +17,16 @@
...
     log_on_register_complete,
 )
-from azure.iot.device.provisioning.internal.polling_machine import PollingMachine
+from azure.iot.device.provisioning.pipeline import constant as dps_constant
 
 logger = logging.getLogger(__name__)
 
 
+async def handle_result(callback):
+    # TODO Define exceptions and handle
```
BertKleewein: ># TODO Define exceptions and handle [](start = 4, length = 35)

This is probably going to look exactly like the code that Carter added to the iothub device clients.  Maybe that code can be moved into a common function used by all clients.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -4,6 +4,10 @@
 # license information.
 # --------------------------------------------------------------------------
 from azure.iot.device.common.pipeline.pipeline_ops_base import PipelineOperation
+from azure.iot.device.common.pipeline.pipeline_ops_base import (
+    RequestAndResponseOperation,
+    RequestOperation,
+)
```
BertKleewein: These aren't used in this file.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+    #     Return None if there is an error or else return the retry-after value obtained from the
+    #     service response. If the service has not given a retry-afer then return a default value.
+    #     The only reason a operation will not be retried is if it has moved to a different operation or else it has completed successfully.
+    #     In this case it means that a SendRegistrationRequestOperation
+    #     """
+    #     key_values_dict = provisioning_op.key_values
+    #     retry_after = (
+    #         None if "retry-after" not in key_values_dict else str(key_values_dict["retry-after"][0])
+    #     )
+    #     return retry_after
```
BertKleewein: please remove

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _process_error(self, provisioning_op, prov_op_name, error):
```
BertKleewein: >_process_error [](start = 8, length = 14)

the word "process" doesn't tell me what this function does. Suggest naming it "_get_error_from_provisioning_response"

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+                if "createdDateTimeUtc" not in decoded_state
+                else str(decoded_state["createdDateTimeUtc"]),
+                last_update_date_time=None
+                if "lastUpdatedDateTimeUtc" not in decoded_state
+                else str(decoded_state["lastUpdatedDateTimeUtc"]),
+                etag=None if "etag" not in decoded_state else str(decoded_state["etag"]),
+            )
+
+        registration_result = RegistrationResult(
+            # request_id=request_id,
```
BertKleewein: ># request_id=request_id, [](start = 12, length = 24)

please remove

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+        decoded_state = (
+            None
+            if "registrationState" not in decoded_response
+            else decoded_response["registrationState"]
+        )
+        registration_state = None
+        if decoded_state is not None:
+            # Everything needs to be converted to string explicitly for python 2
+            # as everything is by default a unicode character
+            registration_state = RegistrationState(
```
BertKleewein: >registration_state [](start = 12, length = 18)

(I hate how our pretty-printer works sometimes.  It does not make this code readable.)
suggestion: I see this pattern a bunch.  How about something like this: device_id=get_optional_element(decoded_state, "deviceId") where get_optional_element is your helper that returns None if the element is missing from the dict?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+        registration_result = RegistrationResult(
+            # request_id=request_id,
+            operation_id=operation_id,
+            status=status,
+            registration_state=registration_state,
+        )
+        return registration_result
+
+    @abc.abstractmethod
+    def _execute_op(self, op):
```
BertKleewein: You shouldn't need this.  It should inherit the abstract method from the base class.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+                        if registration_status == "assigned" or registration_status == "failed":
+                            # process complete response here
+                            complete_registration_result = self._form_complete_result(
+                                request_id=provisioning_op.request_id,
+                                operation_id=operation_id,
+                                decoded_response=decoded_response,
+                                status=registration_status,
+                            )
+                            op.registration_result = complete_registration_result
+                            self.complete_op(op, error=error)
```
BertKleewein: >self.complete_op(op, error=error [](start = 28, length = 32)

At this point, if registration_status == 'failed', it will complete the op with error=None.  I would assume that we would want to complete this op with an error since there was a failure.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+                        elif registration_status == "assigning":
+                            self.send_op_down(
+                                op=pipeline_ops_provisioning.SendQueryRequestOperation(
+                                    request_payload=" ",
+                                    operation_id=operation_id,
+                                    callback=op.callback,
```
BertKleewein: >callback=op.callback, [](start = 36, length = 21)

This concerns me.  Moving a callback from one op into another op will cause problems, though I'm not 100% sure what will happen.  I do know that the original SendRegistrationRequestOperation will probably become "orphaned" at this point.  When the SendQueryRequestOperation completes, it will call the (original) SendRegistrationRequestOperation callback with the (new) SendQueryRequestOperation object as the first parameter.  This, in effect, tells the caller that some new operation (which they know nothing about) just completed, and that is nonsensical.  

This is a scenario where the SendQueryReqestOperation is almost a worker_op for SendRegistrationRequestOperation, except, when the SendQueryReqestOperation completes, it needs to copy the registration result from the (new) SendQueryReqestOperation object into the (original) SendRegistrationRequestOperation object.  i think you need to write a callback for the (new) SendQueryReqestOperation which copies the result into the (original0 SendQueryRequestOperation and then completes that op.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+                        stage_name=self.name,
+                        op_name=op.name,
+                        status_code=query_op.status_code,
+                        # TODO Populate with operation Id
+                        oper_id=query_op.operation_id,
+                    )
+                )
+
+                # This could be an error that has been reported to this stage Or this
+                # could be an error because the service responded with status code 300
```
BertKleewein: >status code 300 [](start = 71, length = 15)

not just 300.  This could be an error with any status code >=300

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+
+                        @pipeline_thread.invoke_on_pipeline_thread_nowait
+                        def do_polling():
+                            this = self_weakref()
+                            logger.info(
+                                "{stage_name}({op_name}): retrying".format(
+                                    stage_name=this.name, op_name=op.name
+                                )
+                            )
+                            op.polling_timer.cancel()
```
BertKleewein: >cancel [](start = 45, length = 6)

Also set to None please

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -48,3 +63,331 @@ def _execute_op(self, op):
...
+                            )
+                            op.registration_result = complete_registration_result
+                        else:
+                            error = exceptions.ServiceError(
+                                "Query Status Operation encountered an invalid registration status {status} with a status code of {status_code}".format(
+                                    status=registration_status, status_code=success_status_code
+                                )
+                            )
+
+                        self.complete_op(op, error=error)
```
BertKleewein: >complete_op [](start = 29, length = 11)

same comment as above.  If registration_status== "failed", error should be set on this completion

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: renamed to only `_get_error_from_response` as it is evident from the stage name that it is meant for provisioning.

---
In reply to: [345903764](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345903764) [](ancestors = 345903764)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
olivakar: good idea. i also need query params in the requestandresponseoperation as well.

---
In reply to: [345896656](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345896656) [](ancestors = 345896656)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: changed to a helper function defined in `mqtt_topic.py`. I feel that should be renamed to `mqtt_helper.py` since it is extracting stuff from topic as well as response...
The thing from which the optional element could be retrieved could be anything either string or dictionary

---
In reply to: [345906762](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345906762) [](ancestors = 345906762)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: probably it was giving me a warning saying I must have all abstract methods...but yes its not needed...

---
In reply to: [345907439](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345907439) [](ancestors = 345907439)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: I have set an additional error as well now. The polling machine implementation actually did not have this error. Hence I skipped it. I also do not know (and i have asked the DPS team) what is the nature of response from the service when the status is `failed`

---
In reply to: [345918256](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345918256) [](ancestors = 345918256)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: same comment as below

---
In reply to: [345909570](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345909570) [](ancestors = 345909570)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: that was the reason I did worker op initially and later decided against it. Also I felt that the worker ops are converted op from the original op whereas the query was a separate op altogether.

---
In reply to: [345915183](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r345915183) [](ancestors = 345915183)

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -35,7 +53,6 @@ def __init__(self, provisioning_pipeline):
         :type provisioning_pipeline: :class:`azure.iot.device.provisioning.pipeline.ProvisioningPipeline`
         """
         super(ProvisioningDeviceClient, self).__init__(provisioning_pipeline)
```
BertKleewein: nit: you could remove this whole implementation since it only calls the base.

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -12,11 +12,29 @@
...
-from .internal.polling_machine import PollingMachine
+from azure.iot.device.provisioning.pipeline import constant as dps_constant
+from .pipeline import exceptions as pipeline_exceptions
+from azure.iot.device import exceptions
+
 
 logger = logging.getLogger(__name__)
 
 
+def handle_result(callback):
```
BertKleewein: >handle_result [](start = 4, length = 13)

suggest renaming.  "handle" doesn't tell me what this function does.  maybe "wait_for_registration_completion" or something like that?

azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -77,7 +99,22 @@ def cancel(self):
...
+        """Enable to receive responses from Device Provisioning Service.
+
+        This is a synchronous call, meaning that this function will not return until the feature
+        has been enabled.
+
+        """
+        logger.info("Enabling reception of response from Device Provisioning Service...")
+
+        callback = EventedCallback()
+        self._provisioning_pipeline.enable_responses(callback=callback)
```
BertKleewein: >enable_responses [](start = 36, length = 16)

this call could raise any of the exceptions that you translate in handle_result above.  You should probably translate those exceptions if they're raised from here too.

BertKleewein:
>     def cancel(self):

Didn't we decide to remove the cancel() API?

---
Refers to: azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py:86 in df662a8. [](commit_id = df662a8cba5e02a212901603cced1296c285c041, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
```diff
@@ -17,11 +17,28 @@
...
 )
-from azure.iot.device.provisioning.internal.polling_machine import PollingMachine
+from azure.iot.device.iothub.pipeline import exceptions as pipeline_exceptions
+from azure.iot.device import exceptions
+from azure.iot.device.provisioning.pipeline import constant as dps_constant
 
 logger = logging.getLogger(__name__)
 
 
+async def handle_result(callback):
```
BertKleewein: All comments from the sync client apply to the async client as well.

azure-iot-device/azure/iot/device/provisioning/pipeline/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+# --------------------------------------------------------------------------
+"""This module defines an exception surface, exposed as part of the pipeline API"""
+
+# For now, present relevant transport errors as part of the Pipeline API surface
+# so that they do not have to be duplicated at this layer.
+from azure.iot.device.common.pipeline.pipeline_exceptions import *
+from azure.iot.device.common.transport_exceptions import (
+    ConnectionFailedError,
+    ConnectionDroppedError,
+    # TODO: UnauthorizedError (the one from transport) should probably not surface out of
```
BertKleewein: Nit: good explanation, but please add your initials, like "OKTODO" so we know who wrote it when we come across it in the future.

azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key.py
```diff
@@ -8,6 +8,12 @@
 import asyncio
 from azure.iot.device.aio import ProvisioningDeviceClient
 
+# TODO Remove logging
```
BertKleewein: please remove before pulling.

azure-iot-device/samples/legacy-samples/provision_symmetric_key.py
```diff
@@ -6,6 +6,11 @@
 import os
 from azure.iot.device import ProvisioningDeviceClient
 
+# TODO Remove logging
```
BertKleewein: please remove before pulling

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,327 @@ def _execute_op(self, op):
...
         else:
             self.send_op_down(op)
+
+
+class CommonProvisioningStage(PipelineStage):
+    @pipeline_thread.runs_on_pipeline_thread
+    def _get_error(self, provisioning_op, prov_op_name, error):
+        if error:
+            return error
+        elif 300 <= provisioning_op.status_code < 429:
```
BertKleewein: >provisioning_op.status_code < 429: [](start = 20, length = 34)

what about status codes 430?  It looks like those all get converted into success.  Is that what you want to do?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,327 @@ def _execute_op(self, op):
...
+                    retry_interval = (
+                        int(op.retry_after, 10)
+                        if op.retry_after is not None
+                        else constant.DEFAULT_POLLING_INTERVAL
+                    )
+                    decoded_response = self._decode_response(op)
+                    operation_id = self._get_operation_id(decoded_response)
+
+                    # retry after scenario
+                    if success_status_code >= 429:
```
BertKleewein: >if success_status_code >= 429: [](start = 20, length = 30)

This confuses me.  You're treating everything 429+ as success, but then treating the 429+ failures as a special kind of success that's not really a success.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: may be we need to speak. This was the exact behavior in polling state machine. Anything >= 429 was going to retry the process again in the PUT Register as well as GET Query

---
In reply to: [350960869](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r350960869) [](ancestors = 350960869)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: may be we need to speak. This was the exact behavior in polling state machine. Anything >= 429 was going to retry the process again in the PUT Register as well as GET Query. It is not a true error unless I am mistaken

---
In reply to: [350962570](https://github.com/Azure/azure-iot-sdk-python/pull/391#discussion_r350962570) [](ancestors = 350962570)

olivakar:
>     def cancel(self):

yes. I kept it as a TODO for a future PR , but have removed it now.

---
In reply to: [558762617](https://github.com/Azure/azure-iot-sdk-python/pull/391#issuecomment-558762617) [](ancestors = 558762617)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py:86 in df662a8. [](commit_id = df662a8cba5e02a212901603cced1296c285c041, deletion_comment = False)

olivakar:
Closing PR and taking the polling changes to a different master branch

PR #392, opened 2019-11-13
refactor(azure-iot-device): Pipeline Refactor Part 1

cartertinney:
* Methods related to completion of operations are now part of Operation classes
* Methods related to flow through the pipeline remain on the Stage classes
* Interception of completions no longer needed due to the new .add_callback() method on Operations
* Operations now support multiple callbacks
* Operation tests now universally applied to all operations following a hierarchical pattern
* Stage tests

**PLEASE NOTE:**
* As a result of the changes that have been made, there has been a slight reduction of coverage in PipelineStage tests. This will be resolved in the Part 2 PR
* In addition, all `provisioning_pipeline.py` tests have been disabled, pending an upcoming PR from Oliva. This too, will be resolved in the Part 2 PR
* Further changes to callback patterns may be made in Part 2 (likely removal of the `error` parameter, which is no longer necessary due to `error` being stored as an attribute on an Operation)


azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -44,10 +51,79 @@ def __init__(self, callback):
...
-        self.callback = callback
+        self.callbacks = []
         self.needs_connection = False
         self.completed = False
 
+        self.callbacks.append(callback)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def add_callback(self, callback):
+        self.callbacks.append(callback)
```
BertKleewein: >append [](start = 23, length = 6)

Something to think about:  If our collection of callbacks is a list (in the abstract sense), then the call order is not obvious (do we call the first item in the list first, or do we call the last item first?).  Since we're using append here, i think we call the last one first, but that's the opposite of what I would assume.  

However, if we think of our collection of callbacks as a stack, then the order is much more intuitive.  When we add a callback, we push it on the stack.  When the operation is complete, we pop the first callback and call it, then we pop the next.  This matches the order we want for callbacks (LIFO).

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -44,10 +51,79 @@ def __init__(self, callback):
...
+        else:
+            self.completed = True
+
+            while self.callbacks:
+                if not self.completed:
+                    logger.debug(
+                        "{}: Operation no longer complete! Suspending completion".format(self.name)
+                    )
+                    break
+                callback = self.callbacks.pop()
```
BertKleewein: >pop [](start = 42, length = 3)

ah ha!  there's the pop.  If only the list object had a method named push().

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -44,10 +51,79 @@ def __init__(self, callback):
...
+                try:
+                    callback(op=self, error=error)
```
BertKleewein: >error=error [](start = 38, length = 11)

I like this code, but there are 2 big implications here:

1. You suspend callback processing by "uncompleting" the op, which also clears the error.
2. Inside a callback, you can't change the error to a different error, and you also can't change success to error.

I think these are both perfectly acceptable, and maybe even desirable.  I just want to put it out there and keep it in mind.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -44,10 +51,79 @@ def __init__(self, callback):
...
+    @pipeline_thread.runs_on_pipeline_thread
+    def uncomplete(self):
+        logger.debug("{}: Uncompleting".format(self.name))
+        self.completed = False
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def spawn_worker_op(self, worker_op_type, **kwargs):
+        logger.debug("{}: creating worker op of type {}".format(self.name, worker_op_type.__name__))
+
+        # CT-TODO: does this need to use a weakref?
```
BertKleewein: ># CT-TODO: does this need to use a weakref? [](start = 8, length = 43)

personal opinion: it doesn't hurt to use weak references, but why make the code more confusing?  I think we only need weak references if we're passing a callback to a timer function or to an external component.  All other cases are (I think) superfluous.  Maybe we need a coding guideline for this in the wiki?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -44,10 +51,79 @@ def __init__(self, callback):
...
+        def on_worker_op_complete(op, error):
+            logger.debug("{}: Worker op ({}) has been completed".format(self.name, op.name))
+            self.complete(error=error)
+
+        if "callback" in kwargs:
+            provided_callback = kwargs["callback"]
+            kwargs["callback"] = on_worker_op_complete
+            worker_op = worker_op_type(**kwargs)
+            worker_op.add_callback(provided_callback)
+        else:
```
BertKleewein: I don't think I like this code.  You're making some big assumptions here about the nature of the pre-existing callback, and you're making for a confusing flow if there is a pre-existing callback.  If I understand the code, I think completing a worker with a pre-existing callback will first call all the callbacks on the parent op, then, after all that is done, it will call the pre-existing callback.  This doesn't seem useful.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -748,21 +630,19 @@ def on_timeout():
...
             self.send_op_down(op)
 
     @pipeline_thread.runs_on_pipeline_thread
-    def _on_intercepted_return(self, op, error):
+    def _clear_timer(self, op, error):
         # When an op comes back, delete the timer and pass it right up.
         if op.timeout_timer:
             op.timeout_timer.cancel()
             op.timeout_timer = None
         logger.debug("{}({}): Op completed".format(self.name, op.name))
```
BertKleewein: >logger.debug("{}({}): Op completed".format(self.name, op.name)) [](start = 8, length = 63)

this debug string seems like noise.  I would maybe change it to "cancelling timer" and move it into the if block above.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -793,11 +673,8 @@ def _execute_op(self, op):
         Send all ops down and intercept their return to "watch for retry"
         """
         if self._should_watch_for_retry(op):
-            self.send_op_down_and_intercept_return(
-                op=op, intercepted_return=self._on_intercepted_return
-            )
-        else:
-            self.send_op_down(op)
+            op.add_callback(self._do_retry_if_necessary)
```
BertKleewein: >_do_retry_if_necessary [](start = 33, length = 22)

This is great.  I like that the name of the callback describes what it does.  Huge improvement.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -852,14 +728,14 @@ def do_retry():
...
 
             # if we don't keep track of this op, it might get collected.
+            op.uncomplete()
             self.ops_waiting_to_retry.append(op)
             op.retry_timer = Timer(self.retry_intervals[type(op)], do_retry)
             op.retry_timer.start()
 
         else:
             if op.retry_timer:
                 op.retry_timer = None
```
BertKleewein: >retry_timer [](start = 19, length = 11)

possible unrelated bug: op.retry_timer.cancel() should be called.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -255,6 +255,8 @@ def _on_mqtt_disconnected(self, cause=None):
             op = self._pending_connection_op
             self._pending_connection_op = None
 
+            # Swallow any errors, because we intended to disconnect - even if something went wrong, we
```
BertKleewein: >errors [](start = 26, length = 6)

You have the (almost) exact same comment 3 lines down.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -35,32 +35,29 @@ def _execute_op(self, op):
...
-                    device_id=self.auth_provider.device_id,
-                    module_id=getattr(self.auth_provider, "module_id", None),
-                    hostname=self.auth_provider.hostname,
-                    gateway_hostname=getattr(self.auth_provider, "gateway_hostname", None),
-                    ca_cert=getattr(self.auth_provider, "ca_cert", None),
-                    sas_token=self.auth_provider.get_current_sas_token(),
-                    callback=op.callback,
-                ),
-                op=op,
+            worker_op = op.spawn_worker_op(
```
BertKleewein: >spawn_worker_op [](start = 27, length = 15)

Just thinking out loud, does the verb "spawn" imply that this function also sends it down?  It feels like it does, but I don't know why.  Why is this "spawn_worker_op" instead of "create_worker_op"?  What does "spawn" do that "create" doesn't do?

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -97,100 +95,91 @@ def _execute_op(self, op):
...
                         "{}({}) token update succeeded.  reconnecting".format(self.name, op.name)
                     )
 
-                    self.send_op_down(
-                        pipeline_ops_base.ReconnectOperation(callback=on_reconnect_complete)
+                    # Stop completion of Token Update op, and only continue upon completion of Reconnect op
+                    op.uncomplete()
+                    worker_op = op.spawn_worker_op(
+                        worker_op_type=pipeline_ops_base.ReconnectOperation,
+                        callback=on_reconnect_complete,
```
BertKleewein: >on_reconnect_complete [](start = 33, length = 21)

You could remove this callback (it only logs), especially if this is the only example we have with a callback inside the kwargs on a spawn_worker_op call.

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+            assert op.callbacks[0] is init_kwargs["callback"]
+
+    # If an extended operation instantiation test class is provided, use those tests as well.
+    # By using the extended_op_instantation_test_class as the first parent class, this ensures that
+    # tests from OperationBaseInstantiationTests (e.g. test_needs_connection) can be overwritten by
+    # tests provided in extended_op_instantiation_test_class.
+    if extended_op_instantiation_test_class:
+
+        class OperationInstantiationTests(
+            extended_op_instantiation_test_class, OperationBaseInstantiationTests
```
BertKleewein: >extended_op_instantiation_test_class [](start = 12, length = 36)

Language question: if you want a method implementation in extended_op_initialization_test_class to "replace" the implementation in OperationBaseInstantiationTests, wouldn't you reverse the order of these classes?  Doesn't the second base class get "applied" after the first?

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+            worker_op = op.spawn_worker_op(worker_op_type, **worker_op_kwargs)
+            assert op.complete.call_count == 0
+
+            worker_op.complete(error=error)
+
+            assert op.complete.call_count == 1
+            assert op.complete.call_args == mocker.call(error=error)
+
+        @pytest.mark.it(
+            "Returns a worker operation, which, when completed, triggers the 'callback' optionally provided in the **kwargs parameter, prior to completing the operation that spawned it"
```
BertKleewein: > prior to completing the operation that spawned it [](start = 134, length = 50)

This doesn't actually test the order that these callbacks are called.

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+
+            def cb2(op, error):
+                # Callback 3 has already been triggered, but Callback 1 has not
+                assert cb1_mock.call_count == 0
+                assert cb2_mock.call_count == 1
+                assert cb3_mock.call_count == 1
+                assert len(op.callbacks) == 1
+
+            def cb3(op, error):
+                # Callback 3 has been triggered, but no others have been.
```
BertKleewein: >allback 3 has been triggered, but no others have been. [](start = 19, length = 54)

Nit: Comments on these 3 callbacks are not symmetrical.  First 2 comments describe pre-conditions (state before the callback is called) and last commend describes post-condition (state after the callback is called).

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+            assert cb2_mock.call_count == 1
+            assert cb2_mock.call_args == mocker.call(op=op, error=error)
+            assert cb1_mock.call_count == 1
+            assert cb1_mock.call_args == mocker.call(op=op, error=error)
+
+            # Because exceptions raised in callbacks are caught and sent to the background exception handler,
+            # the assertions in the above callbacks won't be able to directly raise AssertionErrors that will
+            # allow for testing normally. Instead we should check the background_exception_handler to see if
+            # any of the assertions raised errors and sent them there.
+            assert handle_exceptions.handle_background_exception.call_count == 0
```
BertKleewein: >assert handle_exceptions.handle_background_exception.call_count == 0 [](start = 12, length = 68)

This assumption may be correct but it's an assumption nonetheless.  In other words, you assume the asserts in cb1, cb2, and cb3 all pass because there are no calls to handle_background_exception, but you don't verify this.  If I comment out `cb1_mock.side_effect = cb1`, this test would still pass without ever running the checks inside cb1.  Likewise if I break the exception handling around the callback invocation, it's possible that those asserts could fail, but the test would pass.

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+
+        @pytest.mark.it(
+            "Handles any Exceptions raised during execution of a callback by sending them to the background exception handler, and continuing on with completion"
```
BertKleewein: >continuing on with completion [](start = 131, length = 29)

I'm not convinced that continuing with completion is the correct behavior in this scenario.  Yes, we are reporting the exception (as best we can), but continuing with completion is also like ignoring the exception and that seems wrong.

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
```diff
@@ -0,0 +1,461 @@
...
+            assert len(op.callbacks) == 0
+
+            # The exception raised by the callback was passed to the background exception handler
+            assert handle_exceptions.handle_background_exception.call_count == 1
+            assert handle_exceptions.handle_background_exception.call_args == mocker.call(
+                arbitrary_exception
+            )
+
+        @pytest.mark.it(
+            "Allows any BaseExceptions raised during execution of a callback to propagate"
```
BertKleewein: >propagate [](start = 80, length = 9)

Do you want to add asserts that verify that cb3_mock was called but cb1_mock was not called?

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
```diff
@@ -7,64 +7,268 @@
...
+
+
+class EnableFeatureOperationTestConfig(object):
+    @pytest.fixture
+    def cls_type(self):
+        return pipeline_ops_base.EnableFeatureOperation
+
+    @pytest.fixture
+    def init_kwargs(self, mocker):
+        kwargs = {"feature_name": "some_feature", "callback": mocker.MagicMock()}
```
BertKleewein: >kwargs = {"feature_name": "some_feature" [](start = 8, length = 40)

Maybe I missed it, but is there a test that verifies that op.feature_name gets set to some_feature? (same for all other op attributes)  There used to be.  If you removed this, I want to make sure it was intentional.

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
```diff
@@ -7,64 +7,268 @@
...
+        return kwargs
+
+
+class EnableFeatureInstantiationTests(EnableFeatureOperationTestConfig):
+    @pytest.mark.it(
+        "Initializes 'feature_name' attribute with the provided 'feature_name' parameter"
+    )
+    def test_feature_name(self, cls_type, init_kwargs):
+        op = cls_type(**init_kwargs)
+        assert op.feature_name == init_kwargs["feature_name"]
```
BertKleewein: What's the benefit of writing these tests by hand rather than using kwargs to parametrize them?

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -80,8 +80,8 @@ class LocalTestObject(StageRunOpTestBase):
         @pytest.fixture(params=unknown_ops)
         def op(self, request, mocker):
             op = make_mock_op_or_event(request.param)
-            op.callback = mocker.MagicMock()
-            add_mock_method_waiter(op, "callback")
+            op.callbacks.append(mocker.MagicMock())  # TODO: make this simpler
+            # add_mock_method_waiter(op, "callback")
```
BertKleewein: ># add_mock_method_waiter(op, "callback") [](start = 12, length = 40)

If these tests pass without this, then you've removed all consumers of this and you can remove this line completely

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -80,8 +80,8 @@ class LocalTestObject(StageRunOpTestBase):
         @pytest.fixture(params=unknown_ops)
         def op(self, request, mocker):
             op = make_mock_op_or_event(request.param)
-            op.callback = mocker.MagicMock()
-            add_mock_method_waiter(op, "callback")
+            op.callbacks.append(mocker.MagicMock())  # TODO: make this simpler
```
BertKleewein: ># TODO: make this simpler [](start = 53, length = 25)

please take ownership of this TODOD (# CT-TODO) and/or make it easier to understand.  I'm not sure how this particular line of code could be any simpler.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: Agree, one of the internal changes I've been considering is to refer to this as the `callback_stack`

---
In reply to: [345985415](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345985415) [](ancestors = 345985415)

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
cartertinney: Good call.

---
In reply to: [346506390](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346506390) [](ancestors = 346506390)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: Okay good to know, thanks.

---
In reply to: [345994008](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345994008) [](ancestors = 345994008)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: I believe you are mistaken, unless I am mistaken.

What this does is, if the user provides a callback as kwarg for creating a worker op, that callback will be added to the callback_stack AFTER the callback that completes the original op. And because it's a stack, that means that this user provided callback will be called FIRST and BEFORE completing the parent op that spawned the worker. Completing the parent op is ALWAYS last.

---
In reply to: [345996131](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345996131) [](ancestors = 345996131)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -351,7 +233,8 @@ def __init__(self, pipeline_configuration):
         self.pipeline_configuration = pipeline_configuration
 
     def run_op(self, op):
-        op.callback = pipeline_thread.invoke_on_callback_thread_nowait(op.callback)
+        # CT-TODO: make this more elegant
+        op.callbacks[0] = pipeline_thread.invoke_on_callback_thread_nowait(op.callbacks[0])
```
olivakar: >op.callbacks[0] [](start = 75, length = 15)

is this the top of the stack ?

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
olivakar: i think spawn refers to creating something that will start on a different process. Also verb wise spawn refers to multiple instances and create refers to only 1. 

---
In reply to: [346001572](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346001572) [](ancestors = 346001572)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
BertKleewein: OK, I can see the utility in this.  As long as there's a UT to validate the ordering, I'm OK with this.  

---
In reply to: [346547761](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346547761) [](ancestors = 346547761,345996131)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -97,100 +95,91 @@ def _execute_op(self, op):
                 )
             )
 
-            # make a callback that can call the user's callback after the reconnect is complete
             def on_reconnect_complete(reconnect_op, error):
```
olivakar: >on_reconnect_complete [](start = 16, length = 21)

should this have a complete ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: Yes, I've double checked, there are tests for this, and as a result requirements as well.

---
In reply to: [347011108](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347011108) [](ancestors = 347011108,346547761,345996131)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
cartertinney: I don't think spawn implies that it sends it down at all. The reason I chose "spawn" over "create" is that I think it makes the relationship between the original op and the worker op clearer. The original op spawns a worker op, i.e. the worker op comes from the original op. With "create", that isn't as clear.

---
In reply to: [346998748](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346998748) [](ancestors = 346998748,346001572)

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
cartertinney: It actually does, due to the assertion in the callback. However, there were some bugs in this test anyway, which I have fixed up, as well as adding some better clarity.

---
In reply to: [346495978](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346495978) [](ancestors = 346495978)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: It is, however, `.pop()` removes the item from the stack, which we do not want.

I agree however that this is not ideal at all, and want to come up with something better.

---
In reply to: [346992405](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346992405) [](ancestors = 346992405)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
cartertinney: As discussed offline, it allows for simpler, more readable tests. I don't mind using a function to add a series of test classes dynamically into a test module, but only when those tests are still statically laid out in a common place. When the tests themselves are dynamically created, I feel it becomes harder to follow (especially for someone not familiar with the codebase)

---
In reply to: [346512691](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346512691) [](ancestors = 346512691)

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
cartertinney: Perhaps the comment is unclear - doing an assertion raises an `AssertionError`. The `AssertionError` normally is able to bubble up because no Exception handling accounts for it, until it is intercepted by the testing framework, which uses that to fail the test. However, due to us catching ALL exceptions that come from a callback, the AssertionError is also sent to the background handler - like any other exception would be.

I agree that if you comment out the mock side effect lines, this test will be broken, but that's because that's a crucial line of the test. You could say that about MANY lines of tests. So long as the mock side effects ARE set up, this test works correctly.

To your second point, if the Exception handling around the callback invocation were broken within the `.complete()` method, then this test would still fail, as suddenly the `AssertionError`s raised by the callback would be able to propagate to the test framework and immediately cause a failure.

---
In reply to: [346502193](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346502193) [](ancestors = 346502193)

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
cartertinney: I'm open to changing the implementation, I'm not particularly tied to this logic.

---
In reply to: [346503501](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346503501) [](ancestors = 346503501)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Good catch

---
In reply to: [345998377](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345998377) [](ancestors = 345998377)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
cartertinney: Yeah, I had considered it but felt the log might be valuable. If you think it can be removed I'll go ahead and do it though.

---
In reply to: [346003283](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r346003283) [](ancestors = 346003283)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
olivakar: hmm...this seems to be the only way to do peek....

---
In reply to: [347044769](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347044769) [](ancestors = 347044769,346992405)

cartertinney: Actually, now that I think about it, we could pop and re-append... that said, I feel like that's maybe a little less clear about what's happening?

---
In reply to: [347050315](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347050315) [](ancestors = 347050315,347044769,346992405)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: Added some further documentation, I think this is clearer now.

---
In reply to: [345991747](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345991747) [](ancestors = 345991747)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
olivakar: +1. 

---
In reply to: [345999596](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345999596) [](ancestors = 345999596)

cartertinney: I'll attempt to fix this in PR #2

---
In reply to: [345999596](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345999596) [](ancestors = 345999596)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
cartertinney: No. A worker op automatically completes its parent op.

Also this callback no longer exists.

---
In reply to: [347018814](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347018814) [](ancestors = 347018814)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I've made the change (and also in the `do_retry` callback slightly above), but it seems like there are no tests for this entire code branch. The `do_retry` callback has tests that I've adjusted, but I can't seem to find them for this else block. Am I just missing something?

---
In reply to: [345999596](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r345999596) [](ancestors = 345999596)

azure-iot-device/tests/common/pipeline/pipeline_ops_test.py
BertKleewein: I _think_ an exception in a callback becomes the op error.  This is a deviation from the old behavior, but I think your callback stack makes this better.

Say we have an op that has a stack with 3 callbacks.  The op completes with some status.  For our example, lets say it succeeds.

Now, inside the first callback, let's say there is some exception.   This means that something failed in the completion path for this op.  Can we say that this op succeeded?  No.  At this point, the best thing to do is to catch this exception and pass it as the op error to the next callback.

In a different comment thread, I talked about ops that change from error1->error2 or success->error.  This is how that might be accomplished and it feels perfectly intuitive to me.  

---
In reply to: [347047620](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347047620) [](ancestors = 347047620,346503501)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: You're correct.  I have a test to make sure the callback gets called but we don't have a test to make sure the timer is cancelled:
@pytest.mark.it("Calls the original callback with success when the retried op succeeds")
In truth, I'm having trouble figuring out how this code could even be hit because it means that op completes while the op is waiting to be retried (which means it must be completed already).  Can you mark this with a BKTODO please?  I'm reworking this code path and I need to make sure I think this through.  


---
In reply to: [347059768](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347059768) [](ancestors = 347059768,345999596)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: my $.02: popping, adding the wrapper, and then pushing again seems completely clear to me.

---
In reply to: [347050862](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347050862) [](ancestors = 347050862,347050315,347044769,346992405)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
```diff
@@ -7,64 +7,268 @@
...
+    @pytest.mark.it("Initializes the 'method' attribute with the provided 'method' parameter")
+    def test_method(self, cls_type, init_kwargs):
+        op = cls_type(**init_kwargs)
+        assert op.method == init_kwargs["method"]
+
+    @pytest.mark.it(
+        "Initializes the 'resource_location' attribute with the provided 'resource_location' parameter"
+    )
+    def test_resource_location(self, cls_type, init_kwargs):
+        op = cls_type(**init_kwargs)
```
BertKleewein: >op = cls_type(**init_kwargs) [](start = 8, length = 28)

nit: this line is repeated many many times, op could be a fixture, either in this class, RequestOperationTestConfig, or in some base class.  If you do this, you can remove cls_type from the test parameters while you add op.   Feel free to ignore.  I tend to over-design.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -36,17 +43,178 @@ def __init__(self, callback):
...
+            e = pipeline_exceptions.OperationError(
+                "Attempting to complete an already-completed operation: {}".format(self.name)
+            )
+            # This could happen in a foreground or background thread, so err on the side of caution
+            # and send it to the background handler.
+            handle_exceptions.handle_background_exception(e)
+        else:
+            # Operation is now in the process of completing
+            self.completing = True
+            self.error = error
```
BertKleewein: >self.error = error [](start = 12, length = 18)

<grumble> not sure what I think about this.  Now error is a parameter that's passed to the callback and also an attribute of the op.  I think we have to pick one.  If it stays as an attribute, then maybe we should remove it from the callback parameter list.  (not as part of this CR)

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -1,525 +1,525 @@
...
-    ):
-        mock_provisioning_pipeline.connect()
-        mock_mqtt_transport.on_mqtt_connected_handler()
-        mock_provisioning_pipeline.wait_for_on_connected_to_be_called()
-        mock_provisioning_pipeline.disable_responses(None)
-
-        assert mock_mqtt_transport.unsubscribe.call_count == 1
-        assert mock_mqtt_transport.unsubscribe.call_args[1]["topic"] == fake_sub_unsub_topic
+# # -------------------------------------------------------------------------
+# # Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: Pet peeve/Nit: huge deltas like this make for painful source management (merges, rebases, git blame, etc)  If you want to comment out the tests, just add an underscore to the beginning of the class/test name so pytest doesn't pick it up.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -36,17 +43,178 @@ def __init__(self, callback):
...
             )
         self.name = self.__class__.__name__
-        self.callback = callback
+        self.callback_stack = []
         self.needs_connection = False
-        self.completed = False
+        self.completed = False  # Operation has been fully completed
+        self.completing = False  # Operation is in the process of completing
+        self.error = None  # Error associated with Operation completion
+
```
BertKleewein: Something to think about: These 3 attributes (completed, completing, and error) are all part of one thing.  I hesitate to use the word state, but these combine together to form the state of the op, and not all combinations are valid.  For states, I think we have: not_complete, completing_with_error, completing_without_error, completed_with_error, and completed_without_error.  With these 3 attributes, there's 8 possible combinations, but only 5 valid states.  This worries me, especially as code grows and the possibility of internally inconsistent ops gets bigger.

cartertinney: Yeah I'm a little concerned about this as well - for now I feel safe with the value checks I put on each operation, but it's not an ideal situation. As I was working on this class, some of the functionality started to resemble a state machine a bit too much for my liking.

---
In reply to: [347141190](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347141190) [](ancestors = 347141190)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
cartertinney: I agree, this is scheduled for Part 2

---
In reply to: [347139684](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347139684) [](ancestors = 347139684)

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
cartertinney: Yeah that would have been smarter

---
In reply to: [347139962](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347139962) [](ancestors = 347139962)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_base.py
cartertinney: The rationale here is that as the test is testing the instantiation, the instantiation should be in the test

---
In reply to: [347139500](https://github.com/Azure/azure-iot-sdk-python/pull/392#discussion_r347139500) [](ancestors = 347139500)

PR #397, opened 2019-11-21
IotHub Service Client full implementation and unit tests

zolvarga:


azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -453,3 +479,424 @@ def delete_module(self, device_id, module_id, etag=None):
...
+            if the HTTP response status is not in [200].
+
+        :returns: Configuration object containing the updated configuration.
+        """
+        if etag is None:
+            etag = "*"
+
+        return self.protocol.service.delete_configuration(configuration_id, etag)
+
+    def get_configurations(self, max_count=None):
```
olivakar: >get_configurations [](start = 8, length = 18)

How many will it return by default ?
can that be mentioned in docs ?

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -453,3 +479,424 @@ def delete_module(self, device_id, module_id, etag=None):
...
+        status,
+        progress,
+        input_blob_container_uri,
+        input_blob_name,
+        output_blob_container_uri,
+        output_blob_name,
+        exclude_keys_in_export,
+        failure_reason,
+    ):
+        """Creates a new import/export job on an IoT hub.
```
olivakar: >Creates a new import/export job on an IoT hub. [](start = 11, length = 46)

Is import export job very self evident ? 
Coz I am not sure what this means.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
zolvarga: I did not find documentation about that. I will update the doc if we got to know the default.

---
In reply to: [349339528](https://github.com/Azure/azure-iot-sdk-python/pull/397#discussion_r349339528) [](ancestors = 349339528)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
zolvarga: No, it is not. I will add a link to the documentation of it.

---
In reply to: [349340093](https://github.com/Azure/azure-iot-sdk-python/pull/397#discussion_r349340093) [](ancestors = 349340093)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -439,10 +451,24 @@ def get_module(self, device_id, module_id):
         """
         return self.protocol.service.get_module(device_id, module_id)
 
+    def get_modules(self, device_id):
```
cartertinney: `get_all_modules`?

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -453,3 +479,426 @@ def delete_module(self, device_id, module_id, etag=None):
...
+        :param Configuration configuration: The configuration to create.
+
+        :raises: `HttpOperationError<msrest.exceptions.HttpOperationError>`
+            if the HTTP response status is not in [200].
+
+        :returns: Configuration object containing the created configuration.
+        """
+        return self.protocol.service.create_or_update_configuration(configuration_id, configuration)
+
+    def update_configuration(self, configuration_id, configuration, etag):
```
cartertinney: Isn't this optional?

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -453,3 +479,426 @@ def delete_module(self, device_id, module_id, etag=None):
...
+        end_time,
+        job_type,
+        status,
+        progress,
+        input_blob_container_uri,
+        input_blob_name,
+        output_blob_container_uri,
+        output_blob_name,
+        exclude_keys_in_export,
+        failure_reason,
```
cartertinney: Are all of these parameters really required?

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
zolvarga: I am using the convention of our APIs here.

---
In reply to: [349798734](https://github.com/Azure/azure-iot-sdk-python/pull/397#discussion_r349798734) [](ancestors = 349798734)

capfish:
Hi it looks like this is very close to being approved is there anything I can do to help speed it up?  Perhaps add samples?  We're having a terrible time trying to build the C dependencies for azure-iothub-service-client 1.4.6 on OSX 10.15 and it would really simplify things if we could just use this release.  Alternatively, is there a way with azure-iot-hub 2.0.0rc2 to query for device names?  Thanks!

zolvarga:
We are finishing this PR with samples. Yes, there is a "query_iot_hub" API on RegistryManager 

---
In reply to: [561771987](https://github.com/Azure/azure-iot-sdk-python/pull/397#issuecomment-561771987) [](ancestors = 561771987)

azure-iot-hub/azure/iot/hub/iothub_configuration_manager.py
```diff
@@ -0,0 +1,138 @@
...
+        self.auth = ConnectionStringAuthentication(connection_string)
+        self.protocol = protocol_client(self.auth, "https://" + self.auth["HostName"]
+        )
+
+    def get_configuration(self, configuration_id):
+        """Retrieves the IoTHub configuration for a particular device.
+
+        :param str configuration_id: The id of the configuration.
+
+        :raises: `HttpOperationError<msrest.exceptions.HttpOperationError>`
```
cartertinney: >`HttpOperationError<msrest.exceptions.HttpOperationError>` [](start = 17, length = 58)

Something to consider is if we want to expose these lower level exceptions as part of our API.

It wouldn't be incorrect to allow this exception from `msrest` to propagate through our API, plenty of Python libraries do such things.

However, we are then making a contract with the user, that this won't change, which is somewhat out of our control. Not sure how strict we need to be about potential breaking changes on service clients though.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -238,31 +231,11 @@ def delete_device(self, device_id, etag=None):
 
         self.protocol.service.delete_device(device_id, etag)
 
-    def get_service_statistics(self):
```
cartertinney: >def get_service_statistics(self): [](start = 4, length = 33)

We don't need these?

zolvarga: They have been moved to the bottom of the file

---
In reply to: [354966538](https://github.com/Azure/azure-iot-sdk-python/pull/397#discussion_r354966538) [](ancestors = 354966538)

azure-iot-hub/azure/iot/hub/iothub_configuration_manager.py
zolvarga: As we discussed we should rely on the error of MSRest. It is very unlikely they will change this and in service  client we should follow the generated code.

---
In reply to: [354965843](https://github.com/Azure/azure-iot-sdk-python/pull/397#discussion_r354965843) [](ancestors = 354965843)

PR #403, opened 2019-11-27
feat(azure-iot-device): Upload to Blob / Method Invoke

YoDaMa:


azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+    2) Initially I atomized each http_client call. We would create the client in one call, we would connect in another,
+    disconnect in yet another. Seemed fine... ish. But then Carter was like hey we should make all the http_client specific
+    stuff down here and only expose a relevant surface. What's a relevant surface for HTTP? Probably just request. So that's
+    what I'm exposing. Everything, the insantiation of the HTTPSConnection, the sending of the request, the getting of the response,
+    all happens in this one call, and that's better. Probably.
+    """
+
+    @pipeline_thread.invoke_on_http_thread_nowait
+    # TODO: This star syntax is incompatible with Python 2, change it so that the callback is in front of the optional params.
+    def request(self, method, hostname, path, callback, body=None, headers={}, query_params=None):
```
cartertinney: Isn't the hostname provided at instantiation time?

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+            connection = http_client.HTTPSConnection(
+                self._hostname, context=self._create_ssl_context()
+            )
+            logger.debug("connecting to host tcp socket")
+            connection.connect()
+            logger.debug("connection succeeded")
+            # formatted_headers = self._format_headers(headers)
+            url = "https://{hostname}/{path}?{query_params}".format(
+                hostname=hostname, path=path, query_params=query_params
+            )
```
cartertinney: Should this be formatted outside of the Transport?

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+            connection.request(method, url, body=body.encode("utf-8"), headers=headers)
```
cartertinney: >encode("utf-8") [](start = 54, length = 15)

I'd like encoding to happen in a consistent place regardless of which pipeline/transport we're using. Not to say the way it's currently being done for the primary hub/MQTT pipeline is necessarily more correct than this, but it shouldn't be inconsistent. In MQTT we do this well before it gets down to the transport.

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+    def request(self, method, hostname, path, callback, body=None, headers={}, query_params=None):
+        # Sends a complete request to the server
+        logger.info("sending https request")
+        try:
+            logger.debug("creating to https connection")
+            connection = http_client.HTTPSConnection(
+                self._hostname, context=self._create_ssl_context()
+            )
+            logger.debug("connecting to host tcp socket")
+            connection.connect()
```
cartertinney: Why not do this when the transport is instantiated?

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+
+            logger.debug("response received.")
+            logger.debug("closing connection to https host")
+            connection.close()
+            logger.debug("connection closed")
+            logger.info("https request sent")
+            callback(status_code, response_string)
+        except Exception as e:
+            # TODO: This exception needs to be returned in the callback, I cannot raise here because it's in a different thread than the pipeline
+            # and would therefore do nothing. Instead pass an exception to the callback as an error if this is excepted.
```
cartertinney: It is not my understanding that this is true, but I could be wrong.

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+            response = connection.getresponse()
+            status_code = response.status
+            response_string = response.read()
+
+            logger.debug("response received.")
+            logger.debug("closing connection to https host")
+            connection.close()
+            logger.debug("connection closed")
+            logger.info("https request sent")
+            callback(status_code, response_string)
```
cartertinney: Shouldn't all this be in a callback?

azure-iot-device/azure/iot/device/common/http_transport.py
cartertinney: I don't think this is true?

azure-iot-device/azure/iot/device/common/http_transport.py
YoDaMa: yes this is an important discussion topic. The thing with HTTP is that every time it makes a request, it opens a connection. It's not like MQTT in this way, and it's also why I have no connection logic anywhere you'll notice. Because every time I make a request, I open a connection, aka connect to a socket on the remote server, then during the request the TLS stack actually opens a connection to the server for the POST, waits till a response is received, then closes the connection. 

The reason I don't have connect on instantiation is because each connection is specific to a request. It has to be this way if we want to allow multiple requests simultaneously.

---
In reply to: [351475326](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351475326) [](ancestors = 351475326)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -36,14 +36,15 @@ class AbstractIoTHubClient(object):
...
-    def __init__(self, iothub_pipeline):
+    def __init__(self, iothub_pipeline, upload_pipeline):
         """Initializer for a generic client.
 
         :param iothub_pipeline: The pipeline used to connect to the IoTHub endpoint.
         :type iothub_pipeline: :class:`azure.iot.device.iothub.pipeline.IoTHubPipeline`
         """
         self._iothub_pipeline = iothub_pipeline
-        self._edge_pipeline = None
+        self._upload_pipeline = upload_pipeline
```
cartertinney: >self._upload_pipeline = upload_pipeline [](start = 8, length = 39)

Not sure Modules support Upload right now. Maybe this has changed since I last heard? It was a fairly arbitrary restriction at the time.

azure-iot-device/azure/iot/device/common/http_transport.py
YoDaMa: I can understand that, but I think this is a very simple way to do it? It follows HTTP conventions of hostname, path, and query parameters. Abstracting it further might be confusing. It's possible, but then for instance, the hostname cannot be set in instantiation in the HTTP Transport. You might be rubbing into just how HTTP and MQTT are ultimately different protocols with totally different paradigms 

---
In reply to: [351474403](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351474403) [](ancestors = 351474403)

azure-iot-device/azure/iot/device/common/http_transport.py
YoDaMa: ok, good suggestion, I'll move the encoding further up.

---
In reply to: [351475147](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351475147) [](ancestors = 351475147)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -158,16 +174,16 @@ def receive_message(self):
 
 @six.add_metaclass(abc.ABCMeta)
 class AbstractIoTHubModuleClient(AbstractIoTHubClient):
-    def __init__(self, iothub_pipeline, edge_pipeline=None):
+    def __init__(self, iothub_pipeline, upload_pipeline=None, edge_pipeline=None):
```
cartertinney: >upload_pipeline=None [](start = 40, length = 20)

Do modules support upload? I didn't think they did for arbitrary reasons.

azure-iot-device/azure/iot/device/iothub/pipeline/method_invoke_pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MethodInvokePipeline(object):
+    """Pipeline to communicate with Edge.
+    Uses HTTP.
+    """
+
+    def __init__(self, auth_provider, pipeline_configuration):
+        self._pipeline = (
```
cartertinney: I thought we wanted retry or something in this pipeline?

azure-iot-device/azure/iot/device/iothub/pipeline/upload_pipeline.py
```diff
@@ -0,0 +1,91 @@
...
+        # Event Handlers - Will be set by Client after instantiation of this object
+        self.on_connected = None
+        self.on_disconnected = None
+        self.on_c2d_message_received = None
+        self.on_input_message_received = None
+        self.on_method_request_received = None
+        self.on_twin_patch_received = None
+
+        self._pipeline = (
+            pipeline_stages_base.PipelineRootStage(pipeline_configuration=pipeline_configuration)
```
cartertinney: As with the method invoke pipeline, aren't there some other stages we want here?

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+    all happens in this one call, and that's better. Probably.
+    """
+
+    @pipeline_thread.invoke_on_http_thread_nowait
+    # TODO: This star syntax is incompatible with Python 2, change it so that the callback is in front of the optional params.
+    def request(self, method, path, callback, body=None, headers={}, query_params=None):
+        # Sends a complete request to the server
+        logger.info("sending https request")
+        try:
+            logger.debug("creating to https connection")
```
olivakar: >to [](start = 35, length = 2)

`creating a https connection`

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_http.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    A PipelineOperation object
+    """
+
+    def __init__(self, method, path, headers, body, query_params, callback):
+        """
+        Initializer for HTTPPublishOperation objects.
+
+        :param str path: The name of the path to publish to
+        :param str payload: The payload to publish
```
olivakar: >payload [](start = 19, length = 7)

not present

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,193 @@
...
+#     sc = response["statusCode"]
+#     if sc == 400:
+#         # `translateError` shall return an `ArgumentError` if the HTTP response status code is `400`.
+#         error = "ArgumentError({})".format(message)
+#     elif sc == 401:
+#         # `translateError` shall return an `UnauthorizedError` if the HTTP response status code is `401`.
+#         error = "UnauthorizedError({})".format(message)
+#     elif sc == 403:
+#         # `translateError` shall return an `TooManyDevicesError` if the HTTP response status code is `403`.
+#         error = "ooManyDevicesError({})".format(message)
```
BertKleewein: >ooManyDevicesError [](start = 19, length = 18)

typo

azure-iot-device/azure/iot/device/common/http_transport.py
BertKleewein: If we're looking for parity, the MQTT transport is not a good standard to base ourselves on.  Paho has 3 different ways to return errors: an API can raise, an API can return a filing RC code, and _some_ callbacks can return failures.  
No matter how we do threading in the HTTP transport, we will _always_ need a way to return some errors via callback.  Forcing some errors to be raised in the invocation thread feels like parity for the sake of parity without any real benefit.

---
In reply to: [351480077](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351480077) [](ancestors = 351480077)

azure-iot-device/azure/iot/device/common/http_transport.py
BertKleewein: What is this _main_ thread of which you speak?  I think you mean that it appears we block the _calling_ thread waiting for an HTTP response.  

Those friggin thread decorators have too much hidden magic.  I'm wondering if we want to make thread switching much more explicit.  Just thinking out loud.

---
In reply to: [351477195](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351477195) [](ancestors = 351477195)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_method_invoke_http.py
```diff
@@ -0,0 +1,131 @@
...
+                    "http operation returned status {}".format(http_op.status_code)
+                )
+
+        if isinstance(op, pipeline_ops_iothub.SetIoTHubConnectionArgsOperation):
+            self.device_id = op.device_id
+            self.module_id = op.module_id
+            if op.module_id:
+                client_id = "{}/{}".format(op.device_id, op.module_id)
+            else:
+                client_id = op.device_id
```
BertKleewein: client_id is an MQTT concept, and I don't think there is a specific HTTP analog.  It's possible that we don't need an SetHTTPConnectionArgsOperation object and we can just let the HTTPTransportStage use the SetIoTHubConnectionArgsOperation object directly.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_http.py
```diff
@@ -0,0 +1,80 @@
...
+            @pipeline_thread.invoke_on_pipeline_thread_nowait
+            def on_request_completed(status_code, response):
+                logger.debug("{}({}): Request completed. Completing op.".format(self.name, op.name))
+
+                logger.debug("HTTP Response Status: {}".format(status_code))
+                logger.debug(response)
+
+                op.status_code = status_code
+                op.response_body = response
+                op.complete()  # TODO: Do we need to put an error in here?
```
BertKleewein: >op.complete()  # TODO: Do we need to put an error in here? [](start = 16, length = 58)

That's a good question.    I think the mapping from status_code->error should maybe happen at UploadHTTPTranslationStage since that mapping is very iothub specific.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
```diff
@@ -166,6 +166,15 @@ def invoke_on_callback_thread_nowait(func):
     return _invoke_on_executor_thread(func=func, thread_name="callback", block=False)
 
 
+def invoke_on_http_thread_nowait(func):
+    """
+    Run the decorated function on the callback thread, but don't wait for it to complete
+    """
+    # TODO: Refactor this since this is not in the pipeline thread anymore, so we need to pull this into common.
```
BertKleewein: ># TODO: Refactor this since this is not in the pipeline thread anymore, so we need to pull this into common. [](start = 4, length = 108)

We should see if the code can better with manual threading.  Same model, just without the decorator magic.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
BertKleewein: >upload_pipeline [](start = 32, length = 15)

Do we need to have upload_pipeline and _invoke_method_pipeline as 2 distinct things?  Could there be a single pipeline that implements both?  I think that would save a ton of code and I don't see why they _need_ to be separate.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -67,8 +68,13 @@ def create_from_connection_string(cls, connection_string, ca_cert=None, **kwargs
         authentication_provider = auth.SymmetricKeyAuthenticationProvider.parse(connection_string)
         authentication_provider.ca_cert = ca_cert  # TODO: make this part of the instantiation
         pipeline_configuration = IoTHubPipelineConfig(**kwargs)
+        upload_pipeline = None
+        if pipeline_configuration.blob_upload:
```
BertKleewein: >if pipeline_configuration.blob_upload: [](start = 8, length = 38)

Do we really need to give the user the ability to decide that they don't want the blob upload client or not?  I know we do this in other languages, and I think we do it to keep packages small.  I don't think we have the same need in Python.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -323,6 +323,47 @@ def receive_twin_desired_properties_patch(self, block=True, timeout=None):
         logger.info("twin patch received")
         return patch
 
+    def get_blob_shared_access_signature(self, blob_name):
```
BertKleewein: >get_blob_shared_access_signature [](start = 8, length = 32)

Is this function only returning a sas?  Or is it returning more?  If this returns an access signature, what else does the caller need to actually upload.  It looks like there's a bunch of logic in the sample apps that could maybe be put into the client.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -323,6 +323,47 @@ def receive_twin_desired_properties_patch(self, block=True, timeout=None):
...
+            callback = EventedCallback(return_arg_name="storage_info")
+            self._upload_pipeline.get_blob_shared_access_signature(
+                blob_name=blob_name, callback=callback
+            )
+            storage_info = handle_result(callback)
+            logger.info("Successfully retrieved storage_info")
+            return storage_info
+
+    def notify_blob_upload_status(
+        self, correlation_id, upload_response, status_code, status_description
```
BertKleewein: what is a correlation id and where does the caller get it?

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -323,6 +323,47 @@ def receive_twin_desired_properties_patch(self, block=True, timeout=None):
...
+                "No Upload Pipeline Initialized. get_blob_shared_access_signature cannot be called without an Upload Pipeline set."
+            )
+        else:
+            callback = EventedCallback(return_arg_name="storage_info")
+            self._upload_pipeline.get_blob_shared_access_signature(
+                blob_name=blob_name, callback=callback
+            )
+            storage_info = handle_result(callback)
+            logger.info("Successfully retrieved storage_info")
+            return storage_info
```
BertKleewein: Vague name suggests vague definition.  What is a "storage info" the function name suggests that it's a signature, but I'm not really sure.

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -52,8 +52,8 @@ def __init__(self, **kwargs):
 
         :param iothub_pipeline: The IoTHubPipeline used for the client
         :type iothub_pipeline: :class:`azure.iot.device.iothub.pipeline.IoTHubPipeline`
-        :param edge_pipeline: The EdgePipeline used for the client
-        :type edge_pipeline: :class:`azure.iot.device.iothub.pipeline.EdgePipeline`
+        :param edge_pipeline: The MethodInvokePipeline used for the client
```
BertKleewein: >edge_pipeline [](start = 15, length = 13)

My $.02: I really dislike the name "edge_pipeline"

azure-iot-device/azure/iot/device/iothub/pipeline/method_invoke_pipeline.py
BertKleewein: Maybe, but I think we need to add those later.

---
In reply to: [351491603](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351491603) [](ancestors = 351491603)

azure-iot-device/samples/advanced-hub-scenarios/get_storage_sas.py
```diff
@@ -0,0 +1,153 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: >Microsoft [](start = 16, length = 9)

Are you planning to leave this sample here?  It looks like this code is using HTTP directly without going through our client.

azure-iot-device/azure/iot/device/common/http_transport.py
YoDaMa: This is invoked on the thread "azure_iot_http", which is defined in the pipeline_thread file. I think Bert is alluding to how it's confusing to see this?

---
In reply to: [352768796](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352768796) [](ancestors = 352768796,351477195)

azure-iot-device/azure/iot/device/common/http_transport.py
YoDaMa: The threading model used in MQTT shouldn't exactly be used to compare this since what I'm implementing is the equivalent of Paho and the MQTT Transport, not just the MQTT Transport. But without debating that, raising an error here would be confusing and less than ideal. Passing it up as a callback parameter is necessary if this is running on it's own thread asynchronously.

---
In reply to: [352765989](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352765989) [](ancestors = 352765989,351480077)

azure-iot-device/azure/iot/device/iothub/pipeline/method_invoke_pipeline.py
YoDaMa: agreed

---
In reply to: [352791072](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352791072) [](ancestors = 352791072,351491603)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_method_invoke_http.py
YoDaMa: I've removed client_id but I'm keeping SetIoTHubConnectionArgsOperation separate for now. It has enough distinctions that I'm more comfortable with that.

---
In reply to: [352772765](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352772765) [](ancestors = 352772765)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_http.py
YoDaMa: I've done a mapping of status_code to error further up.

---
In reply to: [352774318](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352774318) [](ancestors = 352774318)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_thread.py
YoDaMa: I can work on that after this PR is finished. It's a good idea but maybe save for later.

---
In reply to: [352774880](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352774880) [](ancestors = 352774880)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
YoDaMa: this returns a sas and a correlation_id in an object. I've updated the api name to get_storage_info, since I think that's a better option.

---
In reply to: [352787099](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352787099) [](ancestors = 352787099)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
YoDaMa: This returns a correlation id, sas, hostname, and container name, which are all used for communicating with storage 

---
In reply to: [352788427](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352788427) [](ancestors = 352788427)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
YoDaMa: correlation_id is used by IoT Hub to correlate the upload status of each blob you are communicating with. I can add documentation specifying this. 

---
In reply to: [352787749](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352787749) [](ancestors = 352787749)

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
YoDaMa: this is fair, I'm changing everything to http_pipeline.

---
In reply to: [352789079](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352789079) [](ancestors = 352789079)

azure-iot-device/samples/advanced-hub-scenarios/get_storage_sas.py
YoDaMa: no this will be removed, but for now I'm going to leave it for testing purposes.

---
In reply to: [352793611](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r352793611) [](ancestors = 352793611)

azure-iot-device/azure/iot/device/iothub/pipeline/upload_pipeline.py
YoDaMa: no

---
In reply to: [351491895](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r351491895) [](ancestors = 351491895)

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,110 @@
...
+            ssl_context.load_cert_chain(
+                self._x509_cert.certificate_file,
+                self._x509_cert.key_file,
+                self._x509_cert.pass_phrase,
+            )
+
+        return ssl_context
+
+    @pipeline_thread.invoke_on_http_thread_nowait
+    # TODO: This star syntax is incompatible with Python 2, change it so that the callback is in front of the optional params.
```
cartertinney: ># TODO: This star syntax is incompatible with Python 2, change it so that the callback is in front of the optional params. [](start = 4, length = 122)

Get rid of this TODO

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -0,0 +1,110 @@
...
+        # Sends a complete request to the server
+        logger.info("sending https request.")
+        try:
+            logger.debug("creating an https connection")
+            connection = http_client.HTTPSConnection(self._hostname, context=self._ssl_context)
+            logger.debug("connecting to host tcp socket")
+            connection.connect()
+            logger.debug("connection succeeded")
+            # formatted_headers = self._format_headers(headers)
+            url = "https://{hostname}/{path}{query_params}".format(
```
cartertinney: Remove this line

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_http.py
```diff
@@ -0,0 +1,63 @@
...
+            has completed or failed.
+        """
+        super(SetHTTPConnectionArgsOperation, self).__init__(callback=callback)
+        self.hostname = hostname
+        self.ca_cert = ca_cert
+        self.client_cert = client_cert
+        self.sas_token = sas_token
+
+
+class HTTPRequestAndResponseOperation(PipelineOperation):
```
cartertinney: As we've discussed offline, there should be a way to restructure this pipeline in order to not need this stage. The generic RequestAndResponseOperation from pipeline_stages_base should be sufficient, and if it's not, we need to get to the bottom of why.

azure-iot-device/azure/iot/device/iothub/pipeline/config.py
```diff
@@ -25,3 +25,8 @@ def __init__(self, product_info="", **kwargs):
         """
         super(IoTHubPipelineConfig, self).__init__(**kwargs)
         self.product_info = product_info
+
+        # Now, the parameters below are not exposed to the user via kwargs. They need to be set by manipulating the IoTHubPipelineConfig object.
+        # They are not in the BasePipelineConfig because these do not apply to the provisioning client.
+        self.blob_upload = False
+        self.method_invoke = False
```
cartertinney: I wonder if these might be better in some sort of structure. Again, not super important now, but something to consider.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -366,26 +368,56 @@ def receive_message(self, block=True, timeout=None):
         logger.info("Message received")
         return message
 
+    def get_storage_info(self, blob_name):
```
cartertinney: Is this a clear enough name on the API? `get_storage_info_for_blob` or something to show the connection between this and the other blob method below?

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -366,26 +368,56 @@ def receive_message(self, block=True, timeout=None):
...
+        """Call up to the IoT Hub over HTTP to get information on
+        the storage blob for uploads.
```
cartertinney: Needs properly updated docstring

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -366,26 +368,56 @@ def receive_message(self, block=True, timeout=None):
...
+        return storage_info
+
+    def notify_blob_upload_status(
+        self, correlation_id, is_success, status_code, status_description
+    ):
+        """Provide the IoT Hub with information on the status of an upload to blob attempt. This is used by IoT Hub to notify listening clients
+
+        :param string correlation_id: Provided by IoT Hub on get_storage_info request.
+        :param string is_success:
+
```
cartertinney: Needs fully detailed docstring

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -449,3 +481,16 @@ def receive_message_on_input(self, input_name, block=True, timeout=None):
             message = None
         logger.info("Input message received on: " + input_name)
         return message
+
+    def invoke_method(self, method_params, device_id, module_id=None):
+        """
+        method_params should contain a method_name, payload, conenct_timeout_in_seconds, response_timeout_in_seconds
+        method_result should contain a status, and a payload
```
cartertinney: Again, needs a proper docstring

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -295,14 +295,56 @@ def _on_disconnected(self):
...
 
+    async def get_storage_info(self, blob_name):
+        """Call up to the IoT Hub Endpoint over HTTP to get information on
+        the storage blob for uploads.
+        """
+        # if not self._http_pipeline:
+        #     raise exceptions.ClientError(
+        #         "No Upload Pipeline Initialized. get_storage_info cannot be called without an Upload Pipeline set."
+        #     )
+        # else:
```
cartertinney: Remove commented code if unncessary

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -295,14 +295,56 @@ def _on_disconnected(self):
...
+    async def notify_blob_upload_status(
+        self, correlation_id, is_success, status_code, status_description
+    ):
+        """
+        """
+        # if not self._http_pipeline:
+        #     raise exceptions.ClientError(
+        #         "No Upload Pipeline Initialized. notify_blob_upload_status cannot be called without an Upload Pipeline set."
+        #     )
+        # else:
```
cartertinney: Remove commented code if unnecessary

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -295,14 +295,56 @@ def _on_disconnected(self):
...
+        await get_storage_info_async(blob_name=blob_name, callback=callback)
+        storage_info = await handle_result(callback)
+        logger.info("Successfully retrieved storage_info")
+        return storage_info
+
+    async def notify_blob_upload_status(
+        self, correlation_id, is_success, status_code, status_description
+    ):
+        """
+        """
```
cartertinney: Needs proper docstring

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -295,14 +295,56 @@ def _on_disconnected(self):
         logger.info("twin patch received")
         return patch
 
+    async def get_storage_info(self, blob_name):
+        """Call up to the IoT Hub Endpoint over HTTP to get information on
+        the storage blob for uploads.
```
cartertinney: Needs proper docstring

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -408,3 +448,22 @@ def __init__(self, iothub_pipeline, edge_pipeline=None):
         message = await inbox.get()
         logger.info("Input message received on: " + input_name)
         return message
+
+    async def invoke_method(self, method_params, device_id, module_id=None):
+        # TODO: Should the pipeline level be split into two? According to everyone,
+        # it should be only one in the pipeline level, so I should change this.
```
cartertinney: Remove TODO

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -408,3 +448,22 @@ def __init__(self, iothub_pipeline, edge_pipeline=None):
...
+        """
+        method_params should contain a method_name, payload, conenct_timeout_in_seconds, response_timeout_in_seconds
+        method_result should contain a status, and a payload
+        """
+        # if not self._http_pipeline:
+        #     # TODO: Is this the right type of Error? CC: Carter
+        #     raise exceptions.ClientError(message="Method Invoke only avaiable on Edge Modules")
```
cartertinney: Remove commented code

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -408,3 +448,22 @@ def __init__(self, iothub_pipeline, edge_pipeline=None):
...
         logger.info("Input message received on: " + input_name)
         return message
+
+    async def invoke_method(self, method_params, device_id, module_id=None):
+        # TODO: Should the pipeline level be split into two? According to everyone,
+        # it should be only one in the pipeline level, so I should change this.
+        """
+        method_params should contain a method_name, payload, conenct_timeout_in_seconds, response_timeout_in_seconds
+        method_result should contain a status, and a payload
+        """
```
cartertinney: Needs correct docstring

azure-iot-device/azure/iot/device/iothub/pipeline/http_pipeline.py
```diff
@@ -0,0 +1,126 @@
...
+                auth_provider=auth_provider, callback=callback
+            )
+
+        self._pipeline.run_op(op)
+        callback.wait_for_completion()
+
+    def invoke_method(self, device_id, method_params, callback, module_id=None):
+        """
+        Send a method response to the service.
+        """
```
cartertinney: Here and everywhere in this file: These methods should have docstrings that spell out the errors that can be returned (even though it's via the callback parameter) - it's important for our own documentation to know what can possibly escape this layer.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub_http.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, target_device_id, target_module_id, method_params, callback):
+        """
+        Initializer for MethodInvokeOperation objects.
+
+        :param str target_module_id: The device id of the target device/module
+        :param str target_module_id: The module id of the target module
+        :param method_params: The parameters used to invoke the method, as defined by the IoT Hub specification.
+        :param callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept a PipelineOperation object which indicates the specific operation has which
+         has completed or failed.
```
cartertinney: Use a full indent for docstrings on multiple lines

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub_http.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    def __init__(self, blob_name, callback):
+        """
+        Initializer for GetStorageInfo objects.
+
+        :param str blob_name: The name of the blob that will be created in Azure Storage
+        :param callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept a PipelineOperation object which indicates the specific operation has which
+         has completed or failed.
```
cartertinney: Full indent

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub_http.py
```diff
@@ -0,0 +1,80 @@
...
+    A PipleineOperation object which contains arguments used to get the storage information from IoT Hub.
+    """
+
+    def __init__(self, correlation_id, is_success, status_code, status_description, callback):
+        """
+        Initializer for GetStorageInfo objects.
+
+        :param callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept a PipelineOperation object which indicates the specific operation has which
+         has completed or failed.
```
cartertinney: Full indent

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub_http.py
```diff
@@ -0,0 +1,80 @@
...
+        :type callback: Function/callable
+
+
+        :ivar storage_info: Upon completion, this contains the storage information which was retrieved from the service.
+        :type storage_info: Storage Info
```
cartertinney: > Storage Info [](start = 27, length = 13)

What is that? JSON? Dictionary? What?

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_ops_iothub_http.py
```diff
@@ -0,0 +1,80 @@
...
+        Initializer for GetStorageInfo objects.
+
+        :param str blob_name: The name of the blob that will be created in Azure Storage
+        :param callback: The function that gets called when this operation is complete or has failed.
+         The callback function must accept a PipelineOperation object which indicates the specific operation has which
+         has completed or failed.
+        :type callback: Function/callable
+
+        :ivar storage_info: Upon completion, this contains the storage information which was retrieved from the service.
+        :type storage_info: Storage Info
```
cartertinney: >:type storage_info: Storage Info [](start = 8, length = 32)

See below comment

azure-iot-device/samples/advanced-hub-scenarios/get_storage_sas.py
cartertinney: Make sure this is removed

---
In reply to: [353494570](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r353494570) [](ancestors = 353494570,352793611)

azure-iot-device/samples/advanced-hub-scenarios/get_storage_x509.py
```diff
@@ -0,0 +1,164 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney: This file also appears that it does not use our library and should be removed (or updated to use it)

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        )
+
+        assert http_transport_constructor.call_count == 1
+
+        actual_constructor_args = http_transport_constructor.call_args[1]
+        expected_constructor_args = mocker.call(
+            hostname=fake_hostname, ca_cert=fake_ca_cert, x509_cert=fake_x509_cert
+        )[2]
+        assert actual_constructor_args == expected_constructor_args
+
```
cartertinney: Seems this test is unnecessary? Not really sure what this is saying.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        mock_client = mock_client_constructor.return_value
+        response_value = mock_client.getresponse.return_value
+        response_value.status = 1234
+        response_value.reason = "__fake_reason__"
+        response_value.read.return_value = "__fake_response_read_value__"
+        return mock_client_constructor
+
+    @pytest.fixture
+    def transport(self, mock_http_client_constructor):
+        return HTTPTransport(hostname=fake_hostname)
```
cartertinney: This isn't strictly necessary here like it is on Stage `.run_op()` and `.handle_event()` tests. You could just define these fixtures on the test class below.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        self, mocker, transport, mock_http_client_constructor
+    ):
+        mock_http_client_request = mock_http_client_constructor.return_value.request
+        fake_method = "__fake_method__"
+        fake_path = "__fake_path__"
+        fake_query_params = "__fake_query_params__"
+        expected_url = "https://{}/{}?{}".format(fake_hostname, fake_path, fake_query_params)
+        # output = {"response": None, "error": None}
+
+        # def request_callback(response=None, error=None):
```
cartertinney: Remove commented code

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        expected_url = "https://{}/{}".format(fake_hostname, fake_path)
+        done = transport.request(fake_method, fake_path, mocker.MagicMock())
+        done.result()
+
+        assert mock_http_client_constructor.call_count == 1
+        assert mock_http_client_request.call_count == 1
+        assert mock_http_client_request.call_args[0][0] == fake_method
+        assert mock_http_client_request.call_args[0][1] == expected_url
+        assert mock_http_client_request.call_args[1]["body"] == ""
+        assert bool(mock_http_client_request.call_args[1]["headers"]) is False
```
cartertinney: Here and everywhere - try to avoid checking individual arguments whenever possible

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        assert mock_http_client_constructor.call_count == 2
+
+    @pytest.mark.it("Uses the HTTP Transport SSL Context.")
+    def test_uses_ssl_context(self, mocker, transport, mock_http_client_constructor):
+        ssl_context_mocker = transport._ssl_context
+        done = transport.request(fake_method, fake_path, mocker.MagicMock())
+        done.result()
+
+        assert mock_http_client_constructor.call_count == 1
+        assert mock_http_client_constructor.call_args[1]["context"] == ssl_context_mocker
```
cartertinney: I think this assertion is clearer without the intermediate variable.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        done.result()
+
+        assert mock_http_client_constructor.call_count == 1
+        assert mock_http_client_request.call_count == 1
+        assert mock_http_client_request.call_args[0][0] == fake_method
+        assert mock_http_client_request.call_args[0][1] == expected_url
+        assert mock_http_client_request.call_args[1]["body"] == ""
+        assert bool(mock_http_client_request.call_args[1]["headers"]) is False
+
+    @pytest.mark.it("Formats the request URL correctly given a path and query parameters.")
```
cartertinney: > correctly [](start = 44, length = 10)

Words like "correctly" are generally unnecessary in test names and tend to make them sound more confusing.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        done.result()
+        assert mock_http_client_constructor.call_count == 1
+        assert mock_http_client_request.call_count == 1
+        assert mock_http_client_request.call_args[0][0] == fake_method
+        assert mock_http_client_request.call_args[0][1] == expected_url
+        assert mock_http_client_request.call_args[1]["body"] == ""
+        assert bool(mock_http_client_request.call_args[1]["headers"]) is False
+
+    @pytest.mark.it(
+        "Calls HTTP Client Connection request method with correct method, url, body, and headers."
```
cartertinney: >Connection [](start = 27, length = 10)

"Connection" makes this more confusing than it needs to be.

Also, I try to avoid referring to specific methods, and instead write a requirement based on what it does, to avoid coupling requirements with implementation (thus avoiding some requirement rot) - Something like "Sends an HTTP Request via the HTTP Client" is probably better. Listing the args, and the method name is unnecessary to understanding what this requirement is.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+            pytest.param(
+                "__fake_method__", "__fake_path__", "__fake_query_params__", "__fake_body__", None
+            ),
+            pytest.param(
+                "__fake_method__",
+                "__fake_path__",
+                "__fake_query_params__",
+                "__fake_body__",
+                "__fake_headers__",
+            ),
```
cartertinney: These could probably use some ids in order to format better in the output

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        done = transport.request(fake_method, fake_path, cb)
+        done.result()
+
+        assert mock_http_client_constructor.call_count == 1
+        assert cb.call_count == 1
+        assert cb.call_args[1]["response"]["status_code"] == 1234
+        assert cb.call_args[1]["response"]["reason"] == "__fake_reason__"
+        assert cb.call_args[1]["response"]["resp"] == "__fake_response_read_value__"
+
+    @pytest.mark.it("Raises a ProtocolClientError if request raises an unexpected Exception")
```
cartertinney: >ProtocolClientError [](start = 30, length = 19)

This is not actually tested. You only check the cause.

azure-iot-device/tests/common/test_http_transport.py
```diff
@@ -0,0 +1,279 @@
...
+        actual_body = mock_http_client_request.call_args[1]["body"]
+        actual_headers = mock_http_client_request.call_args[1]["headers"]
+        if body:
+            assert actual_body == body
+        else:
+            assert not bool(actual_body)
+        if headers:
+            assert actual_headers == headers
+        else:
+            assert not bool(actual_headers)
```
cartertinney: This could probably be structured better to be less confusing, but it's not the end of the world.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_http.py
```diff
@@ -0,0 +1,323 @@
...
+        assert not hasattr(stage.pipeline_root, "transport")
+
+        stage.run_op(op)
+
+        assert mock_transport.call_count == 1
+        assert mock_transport.call_args == mocker.call(
+            hostname=op.hostname, ca_cert=op.ca_cert, x509_cert=op.client_cert
+        )
+        assert stage.transport is mock_transport.return_value
+        assert stage.pipeline_root.transport is mock_transport.return_value
```
cartertinney: Horton no longer needs the transport set on the root.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_http.py
```diff
@@ -0,0 +1,98 @@
...
+    def _run_op(self, op):
+        if isinstance(op, pipeline_ops_http.SetHTTPConnectionArgsOperation):
+            # pipeline_ops_http.SetHTTPConenctionArgsOperation is used to create the HTTPTransport object and set all of it's properties.
+            logger.debug("{}({}): got connection args".format(self.name, op.name))
+            self.sas_token = op.sas_token
+            self.transport = HTTPTransport(
+                hostname=op.hostname, ca_cert=op.ca_cert, x509_cert=op.client_cert
+            )
+
+            self.pipeline_root.transport = self.transport
```
cartertinney: >self.pipeline_root.transport = self.transport [](start = 12, length = 45)

No longer required

azure-iot-device/tests/common/pipeline/test_pipeline_stages_http.py
```diff
@@ -0,0 +1,323 @@
...
+        return pipeline_ops_http.HTTPRequestAndResponseOperation(
+            method=fake_method,
+            path=fake_path,
+            headers=fake_headers,
+            body=fake_body,
+            query_params=fake_query_params,
+            callback=mocker.MagicMock(),
+        )
+
+    @pytest.mark.it("Provides the correct parameters to the HTTPTransport")
```
cartertinney: >"Provides the correct parameters to the HTTPTransport" [](start = 20, length = 54)

Sends an HTTP request via the HTTPTransport

azure-iot-device/tests/common/pipeline/test_pipeline_stages_http.py
```diff
@@ -0,0 +1,323 @@
...
+            hostname="fake_hostname",
+            ca_cert="fake_ca_cert",
+            client_cert="fake_client_cert",
+            sas_token="fake_sas_token",
+            callback=mocker.MagicMock(),
+        )
+        stage.run_op(op)
+        assert stage.transport is mock_transport.return_value
+
+        return stage
```
cartertinney: I think this should probably be parametrized for both SAS and X509

azure-iot-device/tests/common/pipeline/test_pipeline_stages_http.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # This is a way for us to mock the transport invoking the callback
+        stage.transport.request.side_effect = mock_request_callback
+        stage.run_op(op)
+        assert op.reason == "__fake_reason__"
+        assert op.response_body == "__fake_response__".encode("utf-8")
+        assert op.status_code == "__fake_status_code__"
+
+    @pytest.mark.it(
+        "Completes the operation with an error if the request invokes the provided callback with an error"
```
cartertinney: >"Completes the operation with an error if the request invokes the provided callback with an error" [](start = 8, length = 98)

Specify that it's the same error

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -1503,6 +1526,144 @@ class TestIoTHubDeviceClientReceiveTwinDesiredPropertiesPatch(
...
 
 
+@pytest.mark.describe("IoTHubDeviceClient (Synchronous) -.get_storage_info()")
+class TestIoTHubDeviceClientGetStorageInfo(WaitsForEventCompletion, IoTHubDeviceClientTestsConfig):
+    @pytest.mark.it("Begins a 'get_storage_info' HTTPPipeline operation")
+    def test_calls_pipeline_get_storage_info(self, client, http_pipeline):
+        fake_blob_name = "__fake_blob_name__"
+        client.get_storage_info(fake_blob_name)
+        assert http_pipeline.get_storage_info.call_count == 1
+        assert http_pipeline.get_storage_info.call_args[0][0] is fake_blob_name
```
cartertinney: Please try to avoid using specific args, instead mock the whole call

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -1503,6 +1526,144 @@ class TestIoTHubDeviceClientReceiveTwinDesiredPropertiesPatch(
...
+        "Raises a client error if the `get_storage_info` pipeline operation calls back with a pipeline error"
+    )
+    @pytest.mark.parametrize(
+        "pipeline_error,client_error",
+        [
+            pytest.param(
+                pipeline_exceptions.ProtocolClientError,
+                client_exceptions.ClientError,
+                id="ProtocolClientError->ClientError",
+            ),
```
cartertinney: Is this really the only expected error?

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -1503,6 +1526,144 @@ class TestIoTHubDeviceClientReceiveTwinDesiredPropertiesPatch(
...
+        "Raises a client error if the `notify_blob_upload_status` pipeline operation calls back with a pipeline error"
+    )
+    @pytest.mark.parametrize(
+        "pipeline_error,client_error",
+        [
+            pytest.param(
+                pipeline_exceptions.ProtocolClientError,
+                client_exceptions.ClientError,
+                id="ProtocolClientError->ClientError",
+            ),
```
cartertinney: Again, is this the only expected error?

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -2417,6 +2570,76 @@ class TestIoTHubModuleClientReceiveTwinDesiredPropertiesPatch(
     pass
 
 
+@pytest.mark.describe("IoTHubModuleClient (Synchronous) -.invoke_method()")
```
cartertinney: Here and everywhere in this file - please add a space between the dash and the unit under test

azure-iot-device/tests/iothub/test_sync_clients.py
YoDaMa: as far as i can see yes, only this or an unhandled error

---
In reply to: [358449708](https://github.com/Azure/azure-iot-sdk-python/pull/403#discussion_r358449708) [](ancestors = 358449708)

PR #405, opened 2019-12-02
refactor(azure-iot-device): Polling Machine To Pipeline

olivakar:
There are couple of TODOs

- Add a time out stage specifically for DPS. The basic time out stage is defined within common.
- Tests for provisioning pipeline

Duplicate for master PR
https://github.com/Azure/azure-iot-sdk-python/pull/391


azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
```diff
@@ -104,13 +108,6 @@ def register(self):
         """
         pass
 
-    @abc.abstractmethod
-    def cancel(self):
```
BertKleewein: > def cancel(self): [](start = 3, length = 18)

Removing cancel is something that should be done before GA since it's a public interface change.

PR #407, opened 2019-12-05
feat(azure-iot-device): create from registration result after provisioning

olivakar:


azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key_and_send_telemetry.py
```diff
@@ -0,0 +1,69 @@
...
+            id_scope=id_scope,
+            symmetric_key=symmetric_key,
+        )
+
+        return await provisioning_device_client.register()
+
+    results = await asyncio.gather(register_device())
+    registration_result = results[0]
+    print("The complete registration result is")
+    print(registration_result.registration_state)
```
BertKleewein: This is a strange style.  Why not just skip the register_device coroutine and the gather call.  Neither of these seem to do anything?

e.g.
```
async def main():
    provisioning_device_client = ProvisioningDeviceClient.create_from_symmetric_key(
        provisioning_host=provisioning_host,
        registration_id=registration_id,
        id_scope=id_scope,
        symmetric_key=symmetric_key,
    )
    registration_result = await provisioning_device_client.register()

    print("The complete registration result is")
    print(registration_result.registration_state)
```

azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key_and_send_telemetry.py
```diff
@@ -0,0 +1,69 @@
...
+
+    print("Will send telemetry from the provisioned device")
+    device_client = IoTHubDeviceClient.create_from_registration_result_and_symmetric_key(
```
BertKleewein: >create_from_registration_result_and_symmetric_key [](start = 39, length = 49)

This sample doesn't check for a failed registration.  Or does provisioning_device_client.register() raise an exception if provisioning fails?

olivakar: good idea to check the status. In the current state machine it should give an error but I think it is better to check

---
In reply to: [354615024](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r354615024) [](ancestors = 354615024)

azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key_and_send_telemetry.py
```diff
@@ -0,0 +1,69 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
BertKleewein: We should also have a sample for symmetric key with group.  And we should figure out how to fold the derive_device_key logic into our api.  I don't like having to force the user to do the hmac and signing for this.  This is something that we should be doing for them.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -151,6 +151,37 @@ def create_from_x509_certificate(cls, x509, hostname, device_id, **kwargs):
         iothub_pipeline = pipeline.IoTHubPipeline(authentication_provider, pipeline_configuration)
         return cls(iothub_pipeline)
 
+    @classmethod
+    def create_from_registration_result_and_symmetric_key(
```
BertKleewein: >create_from_registration_result_and_symmetric_key [](start = 8, length = 49)

I don't have a better answer, but this API (and the x509 one) both feel clunky.  This is much easier than making the user build a connection string, but it feels slapped together.

cartertinney: I agree, I was under the impression that a registration result was sufficient. 

---
In reply to: [354670450](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r354670450) [](ancestors = 354670450)

azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key_and_send_telemetry.py
cartertinney: Yeah, I agree. Gather only makes sense when there are multiple operations.

---
In reply to: [354614793](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r354614793) [](ancestors = 354614793)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -1153,6 +1153,32 @@ def connection_string(self, device_connection_string):
...
+    def fake_device_id(self):
+        return "elder_wand"
+
+    @pytest.fixture
+    def symmetric_key(self):
+        return "Zm9vYmFy"
+
+    @pytest.fixture
+    def registration_result(self, fake_assigned_hub, fake_device_id):
+        class FakeRegistrationState(object):
```
BertKleewein: >FakeRegistrationState [](start = 14, length = 21)

You have RegistrationState and RegistrationResult models in the client.  Why not use those?

azure-iot-device/samples/advanced-hub-scenarios/provision_x509_and_send_telemetry.py
```diff
@@ -0,0 +1,74 @@
...
+
+        return await provisioning_device_client.register()
+
+    results = await asyncio.gather(register_device())
+    registration_result = results[0]
+    print("The complete registration result is")
+    print(registration_result.registration_state)
+
+    print("Will send telemetry from the provisioned device")
+    device_client = IoTHubDeviceClient.create_from_registration_result_and_x509(
```
BertKleewein: >create_from_registration_result_and_x509 [](start = 39, length = 40)

Suggest `create_from_registration_result_and_x509_certificate` to match `create_from_x509_certificate`

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
olivakar: yes it feels like that. The other thing that can be done is to somehow give back the key as part of registration result but that would also mean that x509 becomes an attribute of the result as well. But it does not really make sense as the user already has those

---
In reply to: [354962358](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r354962358) [](ancestors = 354962358,354670450)

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1007,6 +1007,36 @@ def connection_string(self, device_connection_string):
     def sas_token_string(self, device_sas_token_string):
         return device_sas_token_string
 
+    @pytest.fixture
+    def fake_assigned_hub(self):
+        return "DumbledoreArmy.azure-devices.net"
+
+    @pytest.fixture
+    def fake_device_id(self):
+        return "elder_wand"
```
cartertinney: I don't think this fixture is needed

cartertinney: I don't believe `fake_assigned_hub` or `fake_device_id` are necessary fixtures, as the values are accessible from the `registration_result`  fixture

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1007,6 +1007,36 @@ def connection_string(self, device_connection_string):
...
+
+    @pytest.fixture
+    def symmetric_key(self):
+        return "Zm9vYmFy"
```
cartertinney: If this has to be a fixture, this might be better suited to `client_fixtures.py`

azure-iot-device/samples/advanced-hub-scenarios/provision_symmetric_key_and_send_telemetry.py
olivakar: there is a sample already for group but just that the group sample does not include sending telemetry. are you referring to a group sample that will also send telemetry ?

on a separate note , in the group sample the derive device key is part of the sample. as per my understanding the derivation of device key is something that customers should do but i can verify once with jelani.

---
In reply to: [354668549](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r354668549) [](ancestors = 354668549)

azure-iot-device/tests/iothub/aio/test_async_clients.py
olivakar: i had this in the client fixtures but I brought it here as I thought its more specific to Device and DPS. Like the connection string fixture which actually returns the `device_connection_string` ultimately. 

---
In reply to: [355083352](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r355083352) [](ancestors = 355083352)

azure-iot-device/tests/iothub/test_sync_clients.py
olivakar: Technically I only need the device id and assigned hub from those. But yes i can change it.

---
In reply to: [355038779](https://github.com/Azure/azure-iot-sdk-python/pull/407#discussion_r355038779) [](ancestors = 355038779)

BertKleewein:
>     registration_result = register_device(device_id=device_id)

`TypeError: register_device() takes exactly 2 arguments (1 given)`


---
Refers to: azure-iot-device/samples/legacy-samples/provision_symmetric_key_group.py:58 in b0209f2. [](commit_id = b0209f283835b402e0847a6d35fb8d1d0a65f006, deletion_comment = False)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -1153,6 +1155,26 @@ def connection_string(self, device_connection_string):
     def sas_token_string(self, device_sas_token_string):
         return device_sas_token_string
 
+    # @pytest.fixture
+    # def symmetric_key(self):
+    #     return "Zm9vYmFy"
+    #
+    # @pytest.fixture
+    # def x509(self):
+    #     return None
```
BertKleewein: please remove

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1049,6 +1065,370 @@ class TestIoTHubDeviceClientCreateFromSharedAccessSignature(
     pass
 
 
+@pytest.mark.describe(
+    "IoTHubDeviceClient (Synchronous) - .create_from_registration_result() -- Error"
+)
+class TestIoTHubDeviceClientCreateFromRegistrationResultError(IoTHubDeviceClientTestsConfig):
```
BertKleewein: >TestIoTHubDeviceClientCreateFromRegistrationResultError [](start = 6, length = 55)

Having a single implementation that is shared between sync and async makes me wonder if we should have a single set of tests that are shared between sync and async.  It looks like `TestIoTHubDeviceClientCreateFromSharedAccessSignature`  and `TestIoTHubDeviceClientCreateFromConnectionString` do this, maybe `TestIoTHubDeviceClientCreateFromRegistrationResult*` should too.

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1040,13 +913,179 @@ class TestIoTHubDeviceClientCreateFromConnectionString(
...
+                **kwargs
+            )
+
+
+@pytest.mark.describe(
+    "IoTHubDeviceClient (Asynchronous) - .create_from_symmetric_key() -- Configuration"
+)
+class TestIoTHubDeviceClientFromSymmetricKey(IoTHubDeviceClientTestsConfig):
+    @pytest.mark.it(
+        "Uses the symmetrci key and CA certificate combination to create a SymmetricKeyAuthenticationProvider"
```
BertKleewein: >symmetrci [](start = 18, length = 9)

sp

PR #409, opened 2019-12-06
refactor(azure-iot-device): Pipeline Refactor For Tests Pt 2

cartertinney:
* Connect/Disconnect handlers have been removed from the PipelineRootStage in favor of using the existing `send_event_up` architecture
* Infrastructure has been added for more flexible testing of Pipeline Stages
* All stages in the `common` pipeline package have had their tests updated to use the new infrastructure
* Stages from `iothub` and `provisioning` have not yet been moved over to the new infrastructure
* Old infrastructure still (mostly...) exists, and has been updated with comments to be formally deprecated. It will be removed once the remaining test modules no longer require it
* Threading tests have been removed from some test modules pending review of how they are handled


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -99,15 +99,15 @@ def run_op(self, op):
...
+            # within ._run_op()
+            logger.error(msg="Unexpected error in {}._run_op() call".format(self), exc_info=e)
             op.complete(error=e)
 
-    @abc.abstractmethod
-    def _execute_op(self, op):
+    @pipeline_thread.runs_on_pipeline_thread
+    def _run_op(self, op):
         """
         Abstract method to run the actual operation.  This function is implemented in derived classes
```
BertKleewein: >Abstract [](start = 8, length = 8)

Nit: No longer abstract

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -276,32 +250,33 @@ def _handle_pipeline_event(self, event):
...
-            pipeline_thread.invoke_on_callback_thread_nowait(self.on_pipeline_event_handler)(event)
-        else:
-            logger.warning("incoming pipeline event with no handler.  dropping.")
-
-    @pipeline_thread.runs_on_pipeline_thread
-    def on_connected(self):
-        logger.debug(
-            "{}: on_connected.  on_connected_handler={}".format(
-                self.name, self.on_connected_handler
+        if isinstance(event, pipeline_events_base.ConnectedEvent):
```
BertKleewein: >if isinstance(event, pipeline_events_base.ConnectedEvent): [](start = 8, length = 58)

Thanks!  I think this will be much better.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -650,7 +624,7 @@ def on_timeout():
             logger.debug("{}({}): Sending down".format(self.name, op.name))
             self.send_op_down(op)
         else:
-            self.send_op_down(op)
+            super(OpTimeoutStage, self)._run_op(op)
```
BertKleewein: I'm still chewing on this change and i'm not sure if making PipelineStage._run_op concrete was a good call or not.
This changes the fundamental nature of _run_op from "do everything necessary to handle and pass on" to "do a few special things and let the base class handle everything else".  This specific code illustrates one of my concerns.  Why does this line use super._run_op(), but the code 3 lines up uses send_op_down?  The behavior is the same (right now), but we've introduced a new convention for sending ops to the next stage.  Even though there is no functional distinction between the 2 right now, there is a conceptual distinction, and a functional distinction will eventually develop.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -29,6 +30,17 @@ class MQTTTransportStage(PipelineStage):
     is not in the MQTT group of operations, but can only be run at the protocol level.
     """
 
+    def __init__(self):
+        super(MQTTTransportStage, self).__init__()
+
+        # The sas_token will be set when Connetion Args are received
+        self.sas_token = None
```
BertKleewein: >sas_token [](start = 13, length = 9)

observation with no action required: we are getting better at marking methods as convention private, but we almost never mark attributes this way (or at least I don't). Are convention private attributes on internal objects a practice we should encourage?  My gut says yes.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -14,29 +14,202 @@
...
     StageRunOpTestBase,
     StageHandlePipelineEventTestBase,
 )
 from azure.iot.device.common.pipeline.pipeline_stages_base import PipelineStage, PipelineRootStage
-from tests.common.pipeline.pipeline_event_test import add_instantiation_test
 from azure.iot.device.common.pipeline import pipeline_thread, pipeline_exceptions
 from azure.iot.device.common import handle_exceptions
-from .fixtures import FakeOperation
-from .helpers import assert_callback_succeeded
+from .fixtures import ArbitraryOperation
```
BertKleewein: not used.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -14,29 +14,202 @@
...
+
+    #######################
+    # INSTANTIATION TESTS #
+    #######################
+
+    @pytest.mark.describe("{} -- Instantiation".format(stage_class_under_test.__name__))
+    class StageBaseInstantiationTests(StageTestConfig):
+        @pytest.mark.it("Initializes 'name' attribute as the classname")
+        def test_name(self, cls_type, init_kwargs):
+            stage = cls_type(**init_kwargs)
```
BertKleewein: >stage [](start = 12, length = 5)

nit: why not use the stage fixture?  Here and everywhere cls_type is called with **init_kwargs and nothing else.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -14,29 +14,202 @@
...
 from azure.iot.device.common.pipeline import pipeline_thread, pipeline_exceptions
 from azure.iot.device.common import handle_exceptions
-from .fixtures import FakeOperation
-from .helpers import assert_callback_succeeded
+from .fixtures import ArbitraryOperation
 
 logging.basicConfig(level=logging.DEBUG)
 
 
 def add_base_pipeline_stage_tests(
```
BertKleewein: >add_base_pipeline_stage_tests [](start = 4, length = 29)

With this change, we should be able to remove the monkeypatching from the base pipeline stage tests. Should we?  (I'd be willing to try it as an experiment)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -14,29 +14,202 @@
...
+            stage.send_op_down(arbitrary_op)
+
+            assert arbitrary_op.completed
+            assert type(arbitrary_op.error) is pipeline_exceptions.PipelineError
+
+        @pytest.mark.it("Passes the op to the next stage's .run_op() method")
+        def test_passes_op_to_next_stage(self, mocker, stage, arbitrary_op):
+            stage.send_op_down(arbitrary_op)
+            assert stage.next.run_op.call_count == 1
+            assert stage.next.run_op.call_args == mocker.call(arbitrary_op)
```
BertKleewein: >assert stage.next.run_op.call_args == mocker.call(arbitrary_op) [](start = 12, length = 63)

observation: we should use this style (assert call_args == mocker_call(blah)) as often as possible.  Checking all args is much better than checking individual args.  (unless we have to like in test_no_previous_stage) below.  mocker.ANY can be used in cases where we don't want to assert a specific parameter value.

BertKleewein:
>     def stage(self, mocker, cls_type, init_kwargs):

I think you have a double-definition for this fixture.  Once here and once in StageTestConfig inside add_base_pipeline_stage_test and it's not clear which one is being used.
(I might be wrong.  The monkeypatch uses some strange gymnastics.)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:61 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

BertKleewein:
>         stage = pipeline_stages_base.PipelineRootStage(**init_kwargs)

nit: could use stage fixture here too (and below)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:71 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

BertKleewein:
>         time.sleep(0.1)  # CT-TODO: get rid of this

eeek!  Why do you need this?  My first though was pipeline threading, but fake_pipeline_thread should take care of this.

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:196 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

BertKleewein:
>         connect_op = stage.send_op_down.call_args[0][0]

I know that you verified that this is a ConnectOperation in the test above, but do you want to assert the type of the op anyway?  (insert debate about verifying one thing per test here).

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:317 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>         time.sleep(0.1)  # CT-TODO: get rid of this

Good question. I have no idea. It is quite concerning to me as well.

---
In reply to: [562682450](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562682450) [](ancestors = 562682450)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:196 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>         stage = pipeline_stages_base.PipelineRootStage(**init_kwargs)

I disagree. Instantiating the stage is the unit under test. I believe the unit under test should be executed within the test, not in a fixture.

---
In reply to: [562680329](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562680329) [](ancestors = 562680329)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:71 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: I fully agree, this is something I always try to do. It's more accurate, and more readable.

---
In reply to: [354875237](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r354875237) [](ancestors = 354875237)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: As discussed offline, I still feel that the monekypatching is superior here for safety - there are between 4-6 different test units (i.e. classes) added depending on the class, and you'd have to remember to import all 4-6 parent classes, and then subclass them (usually with no additional tests), which could lead to easily forgotten tests. Furthermore, the monkeypatch adds dynamic logic that only sometimes adds certain test units.

---
In reply to: [354873351](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r354873351) [](ancestors = 354873351)

cartertinney:
>         connect_op = stage.send_op_down.call_args[0][0]

It wouldn't hurt to do. I generally try to do stuff like this, must have slipped by here.

---
In reply to: [562685701](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562685701) [](ancestors = 562685701)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:317 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>     def stage(self, mocker, cls_type, init_kwargs):

The one in `pipeline_stage_test` is used for those tests, this one is used for the others.

It's something that bugs me slightly, but I think it's better than the alternatives - those tests in `pipeline_stage_test` should be as black-boxed as possible IMO, and as such, you shouldn't have to explicitly configure the stage fixture the exact way it needs it, on it's behalf (i.e. the correct mocks set, etc.). So it makes it's own (which overrides this one). This one here on the other hand, is clearly part of this testing module, and is used by any tests within this the scope of this testing module.

---
In reply to: [562680024](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562680024) [](ancestors = 562680024)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:61 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: As mentioned elsewhere, I feel that the unit under test (instantiation) should occur within the test itself, rather than a fixture.

---
In reply to: [354869536](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r354869536) [](ancestors = 354869536)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I'm open to changing this back. This made more sense before we stopped sharing the test for this conditional branch.

---
In reply to: [354859717](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r354859717) [](ancestors = 354859717)

BertKleewein:
>     def op(self, mocker):

random observation: I used to think that "op" was a good name for cases like this because it allowed for easy copy/paste of tests, but I'm beginning to believe that disconnect_op might be a better name because it makes the test logic a little easy to understand (you don't need to look up what kind of operation op is to understand what's happening).

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:484 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

BertKleewein:
>         return pipeline_ops_base.ReconnectOperation(callback=mocker.MagicMock())

FWIW, the current functionality is broken.  A ReconnectOperation (now ReauthorizeConnectionOperation) should be a no-op if the transport isn't connected.  I would leave the test as-is for now.

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:532 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

BertKleewein:
>         "Adds the operation to the queue, even if the pipeline is in a connection state where the operation is ready for completion"

This text is confusing because "the operation is ready for completion" is vague.  How about "Adds the operation to the queue, even if the operation could complete immediately because the requested state has already been reached." ?


---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:639 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>     def op(self, mocker):

Unfortunately, this cannot be changed.

The parent Run Op tests have to be generic, thus need to have a generic fixture called "op" provided.

---
In reply to: [562700585](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562700585) [](ancestors = 562700585)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:484 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>         "Adds the operation to the queue, even if the pipeline is in a connection state where the operation is ready for completion"

"Adds the operation to the queue, even if the operation's desired pipeline connection state already has been reached"

---
In reply to: [562706439](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562706439) [](ancestors = 562706439)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:639 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -849,37 +824,27 @@ def _clear_reconnect_timer(self):
...
+                            self.name
+                        )
                     )
-                )
+            self.send_event_up(event)
 
-        if self.previous:
-            self.previous.on_disconnected()
+        else:
+            super(ReconnectStage, self)._handle_pipeline_event(event)
```
olivakar: >super(ReconnectStage, self)._handle_pipeline_event(event) [](start = 12, length = 57)

To maintain similarity with the op_down for stages do we also need to do event_up for these ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -173,7 +179,8 @@ def on_unsubscribed():
             self.transport.unsubscribe(topic=op.topic, callback=on_unsubscribed)
 
         else:
-            self.send_op_down(op)
+            # This code block should not be reached in correct program flow.
```
YoDaMa: If this is something that should not correctly be reached, we should add a logger debug here in case there's ever a bug where this branch is reached.

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
```diff
@@ -14,29 +14,202 @@
...
+
+    # These tests are only run if the Stage in question has NOT overridden the PipelineStage base
+    # implementations of ._run_op() and/or ._handle_pipeline_event()
+
+    if stage_class_under_test._run_op is PipelineStage._run_op:
+
+        @pytest.mark.describe("{} - .run_op()".format(stage_class_under_test.__name__))
+        class StageRunOpUnhandledOp(StageTestConfig, StageRunOpTestBase):
+            @pytest.fixture
+            def op(self, arbitrary_op):
```
YoDaMa: >self [](start = 19, length = 4)

this fixture seems unnecessary, if it's just returning an arbitrary_op

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: An error will be raised

---
In reply to: [355024794](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r355024794) [](ancestors = 355024794)

azure-iot-device/tests/common/pipeline/pipeline_stage_test.py
cartertinney: It alters the name

---
In reply to: [355036486](https://github.com/Azure/azure-iot-sdk-python/pull/409#discussion_r355036486) [](ancestors = 355036486)

olivakar:
>         time.sleep(0.1)  # CT-TODO: get rid of this

it seems to pass without this. 

---
In reply to: [562685723](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562685723) [](ancestors = 562685723,562682450)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:196 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>         time.sleep(0.1)  # CT-TODO: get rid of this

Only sometimes.

---
In reply to: [562761374](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562761374) [](ancestors = 562761374,562685723,562682450)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:196 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

olivakar:
>     # @pytest.mark.it("Adds and starts a reconnect timer (with an interval of the default reconnect delay) to the stage, if the operation is completed unsuccessfully due to a transient connect error")

will this be here for later ?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:2280 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>     # @pytest.mark.it("Adds and starts a reconnect timer (with an interval of the default reconnect delay) to the stage, if the operation is completed unsuccessfully due to a transient connect error")

See comment at top of test class.

---
In reply to: [562784405](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562784405) [](ancestors = 562784405)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:2280 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

cartertinney:
>     # @pytest.mark.it("Adds and starts a reconnect timer (with an interval of the default reconnect delay) to the stage, if the operation is completed unsuccessfully due to a transient connect error")

But I removed this anyway.

---
In reply to: [562789011](https://github.com/Azure/azure-iot-sdk-python/pull/409#issuecomment-562789011) [](ancestors = 562789011,562784405)

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:2280 in 08e8570. [](commit_id = 08e85706857e5e6e6bf0e9d5fc2032f35a163752, deletion_comment = False)

PR #414, opened 2019-12-12
refactor(azure-iot-device): Pipeline Refactor For Tests Pt 3

cartertinney:
* All stages except for MQTTTranslationStages (both iothub and provisioning) have been ported to new testing model


azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -33,6 +33,8 @@ class IoTHubMQTTTranslationStage(PipelineStage):
     def __init__(self):
         super(IoTHubMQTTTranslationStage, self).__init__()
         self.feature_to_topic = {}
+        self.device_id = None
+        self.module_id = None
```
olivakar: why were these attributes included ? its not used anywhere. i see all device ids and module ids coming from ops. are u planning to use later ?

cartertinney: They are used elsewhere in the class. Before they were just being dynamically added, so the class wouldn't even have the attribute until it was added in ._run_op(). This isn't ideal, because for best maintenance, all potential attributes are set in the constructor, rather than on the fly.

PR #424, opened 2019-12-18
fix(azure-iot-hub): Add models into convenience layer API

cartertinney:
* Added a `models.py` file in the convenience layer, which imports all files from the `protocol.models` package. This now allows for models to be accessed by users without having to go through the protocols package which is not supposed to be exposed through our API.


azure-iot-hub/azure/iot/hub/models.py
```diff
@@ -0,0 +1 @@
+from .protocol.models import *
```
BertKleewein: Add copyright header?

PR #425, opened 2019-12-18
fix(azure-iot-device): Change logging level for unexpected disconnects.

BertKleewein:


azure-iot-device/azure/iot/device/common/handle_exceptions.py
```diff
@@ -48,6 +48,9 @@ def swallow_unraised_exception(e, log_msg=None, log_lvl="warning"):
         elif log_lvl == "error":
             logger.error(log_msg)
             logger.error(traceback.format_exc())
+        elif log_lvl == "info":
```
cartertinney: > elif log_lvl == "info": [](start = 7, length = 24)

How about we use the constants defined in `logging` instead, e.g. `logging.INFO`

PR #426, opened 2019-12-18
refactor(azure-iot-device): Rename ca_cert -> server_verification_cert

cartertinney:
Continuation of work started in #418, propagating changes throughout the stack.


BertKleewein:
>             using a connection string with a GatewayHostName parameter.

Out of scope, but this is not strictly true.  This is one common use, but there is no strict relationship between the GatewayHostName and the need for a server verification cert.  You can have a GatewayHostName connection that doesn't require a server_verification_cert, and you can have a protocol gateway which requires a server_verification_cert but no GatewayHostName parameter.  (this is probably non-trivial to set up, but it is possible.)

 Since this is going into end-user doc, I might say.  "Necessary when connecting to an endpoint which has a non-standard root of trust, such as an IoTEdge instance or other protocol gateway."  I don't really like "non-standard root of trust", but I'm not sure what would be better.

---
Refers to: azure-iot-device/azure/iot/device/iothub/abstract_clients.py:57 in 2d7bd56. [](commit_id = 2d7bd56078884d01bad874c6625c60583844ea02, deletion_comment = False)

PR #427, opened 2019-12-18
feat (azure-iot-device): add connected property to client object

BertKleewein:


cartertinney:
Not sure that this if this is the ideal test syntax or not, but we can figure that out later, I've approved this.

PR #432, opened 2020-01-06
refactor(azure-iot-device): Polling Machine Converted To Pipeline Wit…

olivakar:
Polling machine converted to pipeline stages. This part was already reviewed in the following PR but was never merged into master.

Timeout stages specific to provisioning have been added.

Provisioning pipeline and pipeline stages provisioning  tests were changed.


azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -12,31 +12,35 @@
...
-from .internal.polling_machine import PollingMachine
+from azure.iot.device.provisioning.pipeline import constant as dps_constant
+from .pipeline import exceptions as pipeline_exceptions
+from azure.iot.device import exceptions
+
 
 logger = logging.getLogger(__name__)
 
 
+def wait_for_completion(callback):
```
cartertinney: lets call this `handle_result` for parity with the iothub client, and to prevent confusion with the `.wait_for_completion()` method

azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
```diff
@@ -17,31 +17,34 @@
 from azure.iot.device.provisioning.abstract_provisioning_device_client import (
     log_on_register_complete,
 )
-from azure.iot.device.provisioning.internal.polling_machine import PollingMachine
+from azure.iot.device.iothub.pipeline import exceptions as pipeline_exceptions
```
cartertinney: There should be a `azure.iot.device.provisioning.pipeline.exceptions` file similar to the one in iothub, rather than using the same file. Nothing in provisioning should need anything from iothub

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
```diff
@@ -93,3 +93,22 @@ def extract_status_code_from_topic(topic):
...
+    Gets an optional element from json string , or dictionary.
+    :param content: The content from which the element needs to be retrieved.
+    :param element_name: The name of the element
+    :param index: Optional index in case the return is a collection of elements.
+    """
+    element = None if element_name not in content else content[element_name]
+    if element is None:
+        return None
+    else:
+        if type(element) is object or type(element) is dict:
```
cartertinney: >type(element) [](start = 11, length = 13)

isinstance is generally preferred to type checks

azure-iot-device/azure/iot/device/common/pipeline/pipeline_events_base.py
```diff
@@ -49,21 +49,20 @@ class ResponseEvent(PipelineEvent):
...
     :ivar request_id: The request ID which will eventually be used to match a RequestOperation
-        operation to this event.
-    :type request: str
+      operation to this event.
+    :type request_id: str
     :ivar response_body: The body of the response.
-    :type request_body: str
-
-    :ivar status_code:
-    :type status: int
```
BertKleewein: I think status_code got accidentally removed from the documentation block

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
BertKleewein: If you make this change, order of checks will matter.  isinstance(element, object) will check the inheritance hierarchy, so isinstance(o, object) will evaluate to True if o is a list, but type(o) is object will evaluate to False.

---
In reply to: [364023272](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364023272) [](ancestors = 364023272)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -124,7 +124,7 @@ class SendQueryRequestOperation(PipelineOperation):
     This operation is in the group of DPS operations because it is very specific to the DPS client.
     """
 
-    def __init__(self, request_id, operation_id, request_payload, callback):
+    def __init__(self, operation_id, request_payload, registration_result=None, callback=None):
```
BertKleewein: >callback=None [](start = 80, length = 13)

Please don't make callbacks optional  This is important.  Here and everywhere.

azure-iot-device/azure/iot/device/provisioning/pipeline/provisioning_pipeline.py
```diff
@@ -13,9 +13,9 @@
     pipeline_stages_provisioning,
     pipeline_stages_provisioning_mqtt,
 )
-from azure.iot.device.provisioning.pipeline import pipeline_events_provisioning
 from azure.iot.device.provisioning.pipeline import pipeline_ops_provisioning
 from azure.iot.device.provisioning.security import SymmetricKeySecurityClient, X509SecurityClient
+from azure.iot.device.provisioning.pipeline import constant as dps_constants
```
BertKleewein: >dps_constants [](start = 63, length = 13)

consistency nit: why use "dps" here when everywhere else uses "provisioning"?

BertKleewein:
>     def disable_responses(self, callback=None):

disable_responses appears to be unused.  I don't think we need this since there's no advantage to unsubscribing.  Can we remove it?

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/provisioning_pipeline.py:127 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
 
         else:
             super(UseSecurityClientStage, self)._run_op(op)
+
+
+class CommonProvisioningStage(PipelineStage):
```
BertKleewein: docstring for this class please

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+    @pipeline_thread.runs_on_pipeline_thread
+    def _get_error(self, provisioning_op, prov_op_name, error):
```
BertKleewein: >_get_error [](start = 8, length = 10)

This function concerns me because it's doing more than just getting the error.  It's also deciding to swallow all status codes 429 and above.  I understand why you might consider those to not be errors, but this is the wrong place to make that decision.

BertKleewein: >prov_op_name [](start = 42, length = 12)

Instead of passing prov_op_name as a parameter, why not use provisioning_op.request_type?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+    @pipeline_thread.runs_on_pipeline_thread
+    def _form_complete_result(self, operation_id, decoded_response, status):
+        """
+        Create the registration result from the complete decoded json response for details regarding the registration process.
+        """
+        decoded_state = (get_optional_element(decoded_response, "registrationState"),)
+        registration_state = None
+        if decoded_state is not None:
+            # Everything needs to be converted to string explicitly for python 2
+            # as everything is by default a unicode character
```
BertKleewein: I'm not sure what this comment means in this context.  I don't see any conversion taking place and I don't know what "everything is by default a unicode character" means.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+        return get_optional_element(decoded_response, "operationId")
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _form_complete_result(self, operation_id, decoded_response, status):
+        """
+        Create the registration result from the complete decoded json response for details regarding the registration process.
+        """
+        decoded_state = (get_optional_element(decoded_response, "registrationState"),)
+        registration_state = None
+        if decoded_state is not None:
```
BertKleewein: >decoded_state [](start = 11, length = 13)

The way you wrote this, decoded_state is guaranteed to never be None.  It might be (None,) but that is different.  Do you still have a reason to put this in a tuple and access it as decoded_state[0] or would it make more sense to remove the extra parens from your assignment above and access it as decoded_state?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+                        elif registration_status == "assigning":
+
+                            def copy_result_to_original_op(op, error):
+                                logger.debug(
+                                    "Copying registration result from Query Status Op to Registration Op"
+                                )
+                                initial_register_op.registration_result = op.registration_result
+                                initial_register_op.error = error
+
+                            query_worker_op = initial_register_op.spawn_worker_op(
```
BertKleewein: >spawn_worker_op [](start = 66, length = 15)

It looks like this code is _immediately_ doing a query status if registration_status == "assigning".  This is wrong.  It should sleep (either default_polling_interval or retry_after) before querying.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+                                error = exceptions.ServiceError(
+                                    "Registration operation returned failed registration status  with a status code of {status_code}".format(
+                                        status_code=success_status_code
+                                    )
+                                )
+                            # complete the original op in both cases
+                            initial_register_op.complete(error=error)
+                        elif registration_status == "assigning":
+
+                            def copy_result_to_original_op(op, error):
```
BertKleewein: What happens if the client has to query twice before it goes from assigning to assigned?  i don't see how this code handles the case where the SendQueryRequestOperation returns 429 or "assigning"

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+
+    For each operation that needs a timeout check, this stage will add a timer to
+    the operation.  If the timer elapses, this stage will fail the operation with
+    a PipelineTimeoutError.
+    """
+
+    def __init__(self):
+        super(ProvisioningTimeoutStage, self).__init__()
+        self.timeout_intervals = {
+            pipeline_ops_base.RequestOperation: constant.DEFAULT_TIMEOUT_INTERVAL
```
BertKleewein: >RequestOperation [](start = 30, length = 16)

I suspect that this stage isn't doing what you think it's doing.  The way I read this code, it is causing a timeout error if the request fails to publish.  However, if the request publishes successfully, but no response is received, then this stage does nothing.

Timing out on higher level operations like this is tricky.  I know that the node code had the concept of a registration operation timing out, but that code is weak and I don't think it has any real benefits. (I'm bad-mouthing my own code here since I originally wrote that.)

cartertinney:
>         """

Here and everywhere in this file - can you list the errors that can be returned by these methods? Although they aren't technically "raised" (they're instead returned through the callback), having it documented here is very convenient for being able to understand what's going on without knowing the implementation.

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/provisioning_pipeline.py:85 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
cartertinney: +1, it makes dealing with a lot of the difficult connective tissue of the pipeline much easier when there is always a callback.

---
In reply to: [364361479](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364361479) [](ancestors = 364361479)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+    def _decode_response(self, provisioning_op):
+        return json.loads(provisioning_op.response_body.decode("utf-8"))
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _get_registration_status(self, decoded_response):
+        return get_optional_element(decoded_response, "status")
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _get_operation_id(self, decoded_response):
+        return get_optional_element(decoded_response, "operationId")
```
cartertinney: Do these really need to be instance methods? The `self` argument is never used.

azure-iot-device/tests/provisioning/test_sync_provisioning_device_client.py
```diff
@@ -11,7 +11,9 @@
     RegistrationResult,
     RegistrationState,
 )
-from azure.iot.device.provisioning.pipeline import pipeline_ops_provisioning
+from azure.iot.device.iothub.pipeline import exceptions as pipeline_exceptions
```
cartertinney: >iothub [](start = 22, length = 6)

Make sure to use the provisioning pipeline exceptions

azure-iot-device/tests/provisioning/models/test_registration_result.py
```diff
@@ -100,17 +99,19 @@ def test_some_properties_of_state_are_not_settable(self, input_setter_code):
     )
     def test_registration_state_to_string_without_payload(self):
         registration_state = create_registration_state()
+        json_payload = json.dumps(None, default=lambda o: o.__dict__, sort_keys=True)
```
cartertinney: >json_payload = json.dumps(None, default=lambda o: o.__dict__, sort_keys=True) [](start = 8, length = 77)

Add comment explaining what's going on in this test?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+    This is the first stage that decides converts a registration request
+    into a normal request and response operation.
+    Upon the receipt of the response this stage decides whether
+    to send another registration request or send a query request.
+    Depending on the status and result of the response
+    this stage may also complete the registration process.
+    """
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _run_op(self, operation):
```
cartertinney: >operation [](start = 22, length = 9)

This PROBABLY isn't a problem, but just to be safe, make sure that this is called "op" instead like it is on all other stages to prevent any bugs.

cartertinney:
> class SendRegistrationRequestOperation(PipelineOperation):

Not sure this is an accurate name - this isn't just for sending the request, it also waits for the response, i.e. it is in fact the entire Registration process, rather than just the request.

Perhaps just simply `RegistrationOperation`?

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py:94 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
cartertinney: I don't think we have any high level timeout in iothub, even for request and response operations like `get_twin`. Do we need this here? And if so, do we also need it in iothub? If it's also needed in iothub, is there some kind of more generic shared solution we could implement?

---
In reply to: [364419178](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364419178) [](ancestors = 364419178)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+                        exceptions.ServiceError(
+                            "Operation timed out before provisioning service could respond for {type_of} operation".format(
+                                type_of=op.request_type
+                            )
+                        )
+                    )
+                )
+
+            logger.debug("{}({}): Creating timer".format(self.name, op.name))
+            op.timeout_timer = Timer(constant.DEFAULT_TIMEOUT_INTERVAL, on_timeout)
```
cartertinney: >timeout_timer [](start = 15, length = 13)

Also this is the same timer variable name used in the `OpTimeoutStage`. While I don't think any single operation can trigger both stages, I still think it's probably best if we can give this a separate name.

We also really ought to instantiate all these attributes initially (as None) on all operations via the parent Operation object, because dynamically applying them is asking for trouble. However, that's a change that spans both pipelines and all operations, so best to save it for a different PR.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+            op.timeout_timer.start()
+
+            # Send the op down, but intercept the return of the op so we can
```
cartertinney: > but intercept the return [](start = 31, length = 25)

Outdated comment

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
cartertinney: This also only seems to be used in one place, so why is it shared amongst many classes?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+                # could be an error because the service responded with status code >=300
+                error = self._get_error(op, REGISTER, error=error)
+
+                if not error:
+                    success_status_code = op.status_code
+                    retry_interval = (
+                        int(op.retry_after, 10)
+                        if op.retry_after is not None
+                        else constant.DEFAULT_POLLING_INTERVAL
+                    )
```
cartertinney: This is only relevant in the >= 429 block

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+                    decoded_response = self._decode_response(op)
```
cartertinney: >decoded_response = self._decode_response(op) [](start = 20, length = 44)

This is only relevant in the success block

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
```diff
@@ -99,21 +99,21 @@ class SendRegistrationRequestOperation(PipelineOperation):
...
         """
         super(SendRegistrationRequestOperation, self).__init__(callback=callback)
-        self.request_id = request_id
         self.request_payload = request_payload
         self.registration_id = registration_id
+        self.registration_result = registration_result
+        self.retry_after_timer = None
 
 
 class SendQueryRequestOperation(PipelineOperation):
```
cartertinney: >SendQueryRequestOperation [](start = 6, length = 25)

Likewise, this isn't just sending a request. It also waits for the response. Which is to say, it's the entire query process. Perhaps `QueryRegistrationStatusOperation`?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,388 @@ def _run_op(self, op):
...
+    Upon the receipt of the response this stage decides whether
+    to send another query request or complete the procedure.
+    """
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _run_op(self, op):
+        if isinstance(op, pipeline_ops_provisioning.SendQueryRequestOperation):
+            query_status_op = op
+
+            def on_query_response(op, error):
```
cartertinney: >def on_query_response(op, error): [](start = 12, length = 33)

It's kinda odd that this functionality essentially is duplicated on two different stages. I see why, but this doesn't seem an ideal solution because it effectively means any changes have to be made and tested in two different places.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_events_base.py
olivakar: its at the top

---
In reply to: [364355441](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364355441) [](ancestors = 364355441)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: is there a better place for this ? may be we need to talk regarding this. According to me , >= 429 are only things where we need to take a different decision.

---
In reply to: [364368562](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364368562) [](ancestors = 364368562)

olivakar: the `get_error` method us used for both stages

---
In reply to: [364538162](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364538162) [](ancestors = 364538162)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: left over from previous code.

---
In reply to: [364377207](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364377207) [](ancestors = 364377207)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: are you referring to the code on line 166 or something else. That code is present in `on_query_response` function.

---
In reply to: [364413575](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364413575) [](ancestors = 364413575)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: i should call retry after here as that would call register again but i should put sleep.

---
In reply to: [364412371](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364412371) [](ancestors = 364412371)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: resolved for first one.

---
In reply to: [364530593](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364530593) [](ancestors = 364530593)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: i initially attempted to have a common function in the super common stage to handle response. But the responses of the 2 stages are only similar in some respect and not for all the scenarios. Even if defined a common function I will need to override. 

---
In reply to: [365417577](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365417577) [](ancestors = 365417577)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
olivakar: I kept the name sort of similar to the registration operation. Both starting with send and both having the word request.  if you have any suggestion for the registration as well , we can consider renaming both.

---
In reply to: [365414678](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365414678) [](ancestors = 365414678)

olivakar:
> class SendRegistrationRequestOperation(PipelineOperation):

I thought of this name. Did not use it as the registration operation is the whole thing whereas in this operation we just send the initial request. 


---
In reply to: [572345256](https://github.com/Azure/azure-iot-sdk-python/pull/432#issuecomment-572345256) [](ancestors = 572345256)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py:94 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

olivakar:
> class SendRegistrationRequestOperation(PipelineOperation):

bot sure if inital registration operation and polling registration status operation sounds better or not.

---
In reply to: [573512573](https://github.com/Azure/azure-iot-sdk-python/pull/432#issuecomment-573512573) [](ancestors = 573512573,572345256)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py:94 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: converted to static.

---
In reply to: [364524177](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364524177) [](ancestors = 364524177)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
BertKleewein: I disagree.  If provisioning status == "assigning", then the client needs to retry regardless of the status_code.  

---
In reply to: [365408165](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365408165) [](ancestors = 365408165)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
BertKleewein: I think we do need it in iothub, but it's tricky because it starts to quickly descend into the topic of cancelling operations, and that's hard to get right, especially since the underlying protocol doesn't really help us here.  

---
In reply to: [364529899](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r364529899) [](ancestors = 364529899,364419178)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: i am not understanding what you mean. 

---
In reply to: [365408409](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365408409) [](ancestors = 365408409)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
cartertinney: I agree that the other also needs to be renamed - I thought I had left a comment about that as well. If you're sending a request and waiting for a response, it is not merely sending a request, and thus needs to be renamed to the high level operation itself.

---
In reply to: [365650046](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365650046) [](ancestors = 365650046,365414678)

cartertinney:
> class SendRegistrationRequestOperation(PipelineOperation):

My understanding is that this operation doesn't complete successfully until the polling is also complete. Thus, it is the whole operation, no?

---
In reply to: [573512733](https://github.com/Azure/azure-iot-sdk-python/pull/432#issuecomment-573512733) [](ancestors = 573512733,573512573,572345256)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py:94 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
cartertinney: Never mind, while decoded response itself is only elsewhere used in the success block, operation id is derived from it, and it is used in a couple different places, thus must be defined here.

---
In reply to: [366010391](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r366010391) [](ancestors = 366010391,365408409)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,403 @@ def _run_op(self, op):
...
+
+                if not error:
+                    success_status_code = op.status_code
+                    retry_interval = (
+                        int(op.retry_after, 10)
+                        if op.retry_after is not None
+                        else constant.DEFAULT_POLLING_INTERVAL
+                    )
+                    decoded_response = CommonProvisioningStage._decode_response(op)
+                    operation_id = CommonProvisioningStage._get_operation_id(decoded_response)
```
cartertinney: You can still access static methods with self

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
cartertinney: It's not used in the =="assigning" code block though. Are you saying it should be?

---
In reply to: [365977433](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365977433) [](ancestors = 365977433,365408165)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: what is the accepted pattern ? using `self._get_operation_id` or `CommonProvisioningStage._get_operation_id`

---
In reply to: [366024926](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r366024926) [](ancestors = 366024926)

olivakar:
> class SendRegistrationRequestOperation(PipelineOperation):

yes. are we all okay with renaming to `RegistrationOperation` and `PollingStatusOperation` ?

---
In reply to: [573863322](https://github.com/Azure/azure-iot-sdk-python/pull/432#issuecomment-573863322) [](ancestors = 573863322,573512733,573512573,572345256)

---
Refers to: azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py:94 in b898855. [](commit_id = b8988555657f4674cc31bae3dfc331e56d51cb2f, deletion_comment = False)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_ops_provisioning.py
olivakar: commented on the other one. 

---
In reply to: [366018370](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r366018370) [](ancestors = 366018370,365650046,365414678)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,403 @@ def _run_op(self, op):
...
+                                initial_register_op.error = error
+
+                            query_worker_op = initial_register_op.spawn_worker_op(
+                                worker_op_type=pipeline_ops_provisioning.SendQueryRequestOperation,
+                                request_payload=" ",
+                                operation_id=operation_id,
+                                callback=copy_result_to_original_op,
+                            )
+
+                            time.sleep(constant.DEFAULT_POLLING_INTERVAL)
```
BertKleewein: >me.sleep(constant.DEFAULT_POLLING_INTERVAL) [](start = 30, length = 43)

EEK!  No! You can't sleep inside the pipeline thread.  This needs to be triggered by some sort of timer.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
BertKleewein: I don't know what really happens or what should happen, but It's possible that retry_after could be set for == "assigning".  

---
In reply to: [366026623](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r366026623) [](ancestors = 366026623,365977433,365408165)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
BertKleewein: It's just a weird flow.  This function, which, according to the name, returns an error from the status_code.  But, as a sneaky hidden side-effect at the very bottom, turns some errors into success.  Then, in PollingStatusStage._run_op, it uses the status_code to handle these special "error that was turned into success" cases.

So, _get_error can return one of two things:
1. "definitely an error" if it return an Error object.
2. "maybe success" if it returns None.  
Callers can never look at the return value from this function to know if the operation succeeded.  They also sometimes need to look at the HTTP status code and sometimes they need to look at the provisioning status.    This is confusing.  This function doesn't make things clearer. The caller still needs add code to handle "maybe success" by looking at the status code and provisioning status.  

It would cleaner if this function could return an answer that callers could rely on.  One possibility is to return one of three things:
1. Operation is still in progress.   We either got 429+ or we got status=="assigning".  Caller knows that it needs to poll again.
1. Definitely an error.  We got 300-428 or we got status=="failed"  Caller knows that it needs to fail the operation.  
2. Definitely success.  We got status=="assigned".  Caller knows that the operation is a success.  


---
In reply to: [365639409](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365639409) [](ancestors = 365639409,364368562)

olivakar: pure error happens when error was returned from the lower levels or i am getting a code in between 300 to 428. Anything after 429 would need to be retried.
hence i designed this accordingly.

if we need to take a decision based on the return value of this function then this function can not return a status code sometimes and a string status other times. I have to assign another status called "retrying" whenever i receive >= 429. For the registration stage these 2 things ("assigning" and "retrying") means 2 separate things.

and for 300-428 i have to assign something like "error" but we already have a valid status called "error". I don't know what to call this 300-428.

let me know if this sounds fine. 



---
In reply to: [366481124](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r366481124) [](ancestors = 366481124,365639409,364368562)

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
olivakar: i have debugged this. The timing out part is fine but the clearing of the timer happens when the request operation completes which is wrong.
The way i see it , clearing of timer should only happen once we receive response from service. 

the only response handlers/callbacks that I have access to and which are specific to DPS are the `on_registration_response` and `on_query_response` functions as other responses are handled by generic stages in `pipeline_stages_base.py`. 

So I have omitted the `ProvisioningTimeoutStage` completely and have included timeout timers in the `RegistrationStage` and `PollingStatusStage` now.

if we want to included time out for all responses from service may be the `CoordinateRequestAndResponseStage` is a better stage to put the timer.

---
In reply to: [365998937](https://github.com/Azure/azure-iot-sdk-python/pull/432#discussion_r365998937) [](ancestors = 365998937,364529899,364419178)

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
```diff
@@ -93,3 +93,28 @@ def extract_status_code_from_topic(topic):
...
+    :param element_name: The name of the element
+    :param index: Optional index in case the return is a collection of elements.
+    """
+    element = None if element_name not in content else content[element_name]
+    if element is None:
+        return None
+    else:
+        # if type(element) is object or type(element) is dict:
+        #     return element
+        # elif type(element) is list:
```
BertKleewein: ># elif type(element) is list: [](start = 8, length = 29)

thanks for leaving the commented code in so we could see before/after, but please remove it before pulling to master.

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic.py
```diff
@@ -93,3 +93,28 @@ def extract_status_code_from_topic(topic):
...
+        #     return element[index]
+        # else:
+        #     return str(element)
+        if isinstance(element, list):
+            return element[index]
+        elif isinstance(element, dict) or isinstance(element, object):
```
BertKleewein: >elif isinstance(element, dict) or isinstance(element, object) [](start = 8, length = 61)

first clause is unnecessary because a dict is also an object.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -44,3 +58,446 @@ def _run_op(self, op):
 
         else:
             super(UseSecurityClientStage, self)._run_op(op)
+
+
+class CommonProvisioningStage(PipelineStage):
+    """
+    This is a super stage that the RegistrationStage and PollingStatusStage of
+    provisioning would both use. It  some common functions like decoding response
```
BertKleewein: >It  some [](start = 33, length = 8)

missing a word between "It" and "some"?

azure-iot-device/azure/iot/device/provisioning/pipeline/provisioning_pipeline.py
```diff
@@ -35,26 +37,21 @@ def __init__(self, security_client, pipeline_configuration):
         self._pipeline = (
             pipeline_stages_base.PipelineRootStage(pipeline_configuration=pipeline_configuration)
             .append_stage(pipeline_stages_provisioning.UseSecurityClientStage())
+            .append_stage(pipeline_stages_provisioning.RegistrationStage())
+            .append_stage(pipeline_stages_provisioning.PollingStatusStage())
+            .append_stage(pipeline_stages_base.CoordinateRequestAndResponseStage())
+            # .append_stage(pipeline_stages_provisioning.ProvisioningTimeoutStage())
```
BertKleewein: >pipeline_stages_provisioning [](start = 28, length = 28)

remove commented code please

PR #434, opened 2020-01-10
feature(azure-iot-device): Add MQTT support for cipher suites

cartertinney:
- Cipher suites can now be set via the "cipher" kwarg on instantiation, in both iothub and provisioning clients (**NOTE: ENABLED FOR MQTT ONLY. HTTP WILL BE IN A SEPARATE PR**)
- Cipher input will be sanitized so we can accept multiple forms of cipher suites
- BasePipelineConfig now is an abstract class
- Added ProvisioningPipelineConfig class
- Added tests for config classes
- Centralized tests for all kwarg options at the user API level
- Added missing provisioning client test infrastructure


azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -115,20 +115,31 @@ def test_stores_data(self, stage, op, mocker, mock_transport):
         stage.run_op(op)
         assert stage.sas_token == op.sas_token
 
-    # TODO: Should probably remove the requirement to set it on the root. This seems only needed by Horton
```
BertKleewein: >the [](start = 64, length = 3)

The mark.it also needs to be updated.

azure-iot-device/azure/iot/device/provisioning/pipeline/config.py
```diff
@@ -0,0 +1,17 @@
...
+# --------------------------------------------------------------------------
+
+import logging
+from azure.iot.device.common.pipeline.config import BasePipelineConfig
+
+logger = logging.getLogger(__name__)
+
+
+class ProvisioningPipelineConfig(BasePipelineConfig):
+    """A class for storing all configurations/options for IoTHub clients in the Azure IoT Python Device Client Library.
```
YoDaMa: nit: shouldn't this comment specify this is the COnfig class specifically for Provisioning?

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -230,6 +230,31 @@ def test_configures_tls_context_with_server_verification_certs(self, mocker, moc
             cadata=server_verification_cert
         )
 
+    # TODO: Finalize these test args once the format for ciphers is determined
+    @pytest.mark.it(
+        "Configures TLS/SSL context with provided cipher(s) if present during instantiation"
+    )
+    @pytest.mark.parametrize(
```
YoDaMa: the parametrization here is an unnecessary complication of this test. If we're talking about not wanting an explosion of tests. We're just testing here that the input is plugged to the input to set_ciphers. The parametrization comes in test_pipeline_stages_mqtt

cartertinney: The explosion I was referring to was at the client level, not here. However, I see the logic in removing this, as all validation happens by the time this is set in the config.

---
In reply to: [369286262](https://github.com/Azure/azure-iot-sdk-python/pull/434#discussion_r369286262) [](ancestors = 369286262)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -11,9 +11,10 @@
 import os
 import io
 import six
+import abc
```
YoDaMa: I don't see this being used anywhere.

YoDaMa:
I'm approving, with the nit: that I still have a comment about test parameterizations to be addressed.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -30,6 +27,28 @@
...
+    """Helper function to validate user provided kwargs.
+    Raises TypeError if an invalid option has been provided"""
+    valid_kwargs = ["product_info", "websockets", "cipher", "server_verification_cert"]
+
+    for kwarg in kwargs:
+        if kwarg not in valid_kwargs:
+            raise TypeError("Got an unexpected keyword argument '{}'".format(kwarg))
+
+
+def _get_pipeline_config_kwargs(**kwargs):
```
olivakar: >_get_pipeline_config_kwargs [](start = 4, length = 27)

I am not sure what is this doing

cartertinney: This is getting the subset of **kwargs that is relevant to a IoTHubPipelineConfig. Not all **kwargs are passed to IoTHubPipelineConfig - specifically, 'server_verification_cert'

---
In reply to: [374839219](https://github.com/Azure/azure-iot-sdk-python/pull/434#discussion_r374839219) [](ancestors = 374839219)

PR #443, opened 2020-01-22
chore (azure-iot-device): remove needless debug messages

BertKleewein:
I'm noticing that many customers set the debug level to DEBUG.  I removed a bunch of messages that just get in the way of understanding what's really happening.


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -380,19 +375,7 @@ def _run_op(self, op):
 
             @pipeline_thread.runs_on_pipeline_thread
             def on_operation_complete(op, error):
-                if error:
```
YoDaMa: is removing all error logging from this part of the code the best?

BertKleewein: you're right.  I'm going to put this back in.

---
In reply to: [369770005](https://github.com/Azure/azure-iot-sdk-python/pull/443#discussion_r369770005) [](ancestors = 369770005)

olivakar:
why did you consider these debug messages to be not useful ?

BertKleewein:
They're too noisy and only really useful in a small set of circumstances.  Before this change, every op and every event would log every stage it went through, with most of the stagers just passing the op to the next stage.  Any given op might have 20 lines of debug output that show it going through 10 stages that don't touch it, and 5 lines of "interesting" output that shows what happens in the stage that actually handles it.  
It's too bad that the default logging levels don't have something more verbose than DEBUG for "give me as much information as you possibly can, I don't care how noisy it is, I want to know everything."

---
In reply to: [578304713](https://github.com/Azure/azure-iot-sdk-python/pull/443#issuecomment-578304713) [](ancestors = 578304713)

PR #447, opened 2020-01-29
fix(azure-iot-device): Message size checks

olivakar:


azure-iot-device/azure/iot/device/constant.py
```diff
@@ -11,3 +11,4 @@
 IOTHUB_API_VERSION = "2018-06-30"
 PROVISIONING_API_VERSION = "2019-03-31"
 SECURITY_MESSAGE_INTERFACE_ID = "urn:azureiot:Security:SecurityAgent:1"
+SIZE_LIMIT = 262144
```
BertKleewein: >SIZE_LIMIT [](start = 0, length = 10)

size of what?  I suggest adding what size this limits.  (e.g. TELEMETRY_MESSAGE_SIZE_LIMIT)

BertKleewein:
>     def send_message(self, message):

Do we have a limit for the size of messages for send_message_to_output?

---
Refers to: azure-iot-device/azure/iot/device/iothub/sync_clients.py:129 in ce5d4ad. [](commit_id = ce5d4ad8007c1e043db2c5d8b8a3745e2638d537, deletion_comment = False)

BertKleewein:
>     def send_message(self, message):

I will test and tell you.  I'm poking in the right area and can do this quickly.

---
In reply to: [580416150](https://github.com/Azure/azure-iot-sdk-python/pull/447#issuecomment-580416150) [](ancestors = 580416150)

---
Refers to: azure-iot-device/azure/iot/device/iothub/sync_clients.py:129 in ce5d4ad. [](commit_id = ce5d4ad8007c1e043db2c5d8b8a3745e2638d537, deletion_comment = False)

BertKleewein:
>     def send_message(self, message):

The answer is: We need to add this code to send_message_to_output also because it fails if message size > 256 KB

---
In reply to: [580416322](https://github.com/Azure/azure-iot-sdk-python/pull/447#issuecomment-580416322) [](ancestors = 580416322,580416150)

---
Refers to: azure-iot-device/azure/iot/device/iothub/sync_clients.py:129 in ce5d4ad. [](commit_id = ce5d4ad8007c1e043db2c5d8b8a3745e2638d537, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -454,6 +460,10 @@ def send_message_to_output(self, message, output_name):
         """
         if not isinstance(message, Message):
             message = Message(message)
+
+        if message.get_size() > device_constant.TELEMETRY_MESSAGE_SIZE_LIMIT:
+            raise ValueError("Size of telemetry message can not exceed 256 KB.")
```
BertKleewein: >raise ValueError [](start = 12, length = 16)

Don't forget to add :raises: for this to the docstring above.  I would also remove the word "telemetry" from the exception string ("Size of message can not exceed 256 KB"

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -417,6 +422,9 @@ def __init__(self, iothub_pipeline, http_pipeline):
         if not isinstance(message, Message):
             message = Message(message)
 
+        if message.get_size() > device_constant.TELEMETRY_MESSAGE_SIZE_LIMIT:
```
BertKleewein: >message [](start = 11, length = 7)

same comment from sync_client.py on :raises: doc and exception message.

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -2266,6 +2303,50 @@ def fail_send_output_event(message, callback):
...
+        data_input = "serpensortia"
+        message = Message(data_input)
+        message.custom_properties["spell"] = data_input * 256000
+        with pytest.raises(ValueError) as e_info:
+            await client.send_message_to_output(message, output_name)
+        assert "256 KB" in e_info.value.args[0]
+        assert iothub_pipeline.send_output_event.call_count == 0
+
+    @pytest.mark.it("Does not raises error when message data size is equal to 256 KB")
+    async def test_raises_error_when_message_to_output_data_equal_to_256(
```
BertKleewein: >test_raises_error_when_message_to_output_data_equal_to_256 [](start = 14, length = 58)

I go back and forth on the utility of the equal_to_256 tests.  I don't think I would be sad to see them disappear.

olivakar: I want to keep it to check the boundary..but right now it is not serving any purpose. but i do want to address it after i do the content type and content encoding as things are bound to change then.

---
In reply to: [374956555](https://github.com/Azure/azure-iot-sdk-python/pull/447#discussion_r374956555) [](ancestors = 374956555)

PR #456, opened 2020-02-04
release: Bump package versions for release on 2020-02-04

olivakar:


azure-iot-device/.bumpverion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.0.0
+current_version = 2.0.1
```
jebrando: Not to ask a Newbie question here, but why is this version 1 behind the the other versions in this PR

PR #460, opened 2020-02-11
refactor(azure-iot-device): Refactor ReconnectStage to fix reconnect bugs and make first-connect behavior more explicit

BertKleewein:


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -647,7 +647,6 @@ def __init__(self):
         self.retry_intervals = {
             pipeline_ops_mqtt.MQTTSubscribeOperation: 20,
             pipeline_ops_mqtt.MQTTUnsubscribeOperation: 20,
-            pipeline_ops_base.ConnectOperation: 20,
```
BertKleewein: >pipeline_ops_base.ConnectOperation: 20, [](start = 12, length = 39)

This is a very significant change.  The RetryStage is no longer retrying ConnectOperation.  That's entirely the responsibility of the ReconnectStage.  Without this change, it's impossible to reason about the behavior of the pipeline because there would be two different stages that are both responsible for reconnecting depending on how the error happens.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,146 @@ def do_retry():
 ]
 
 
+class ReconnectState(object):
+    never_connected = "never_connected"
+    waiting_to_reconnect = "waiting_to_reconnect"
+    deferred = "deferred"
```
BertKleewein: >deferred = "deferred" [](start = 4, length = 21)

This is a strange name but I'm not sure what else to call it.  "deferred' means "defer the interpretation of the state to self.pipeline_root.connected." Maybe "whatever_the_pipeline_root says" is a better name.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,146 @@ def do_retry():
 ]
 
 
+class ReconnectState(object):
```
olivakar: >ReconnectState [](start = 6, length = 14)

it feels these are getting used liked enums. though we can not use enums for some reasons that i have quite forgotten , may be it is  a good idea to use upper caps for all these variables

olivakar: >ReconnectState [](start = 6, length = 14)

also doc

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,146 @@ def do_retry():
...
         # connect delay is hardcoded for now.  Later, this comes from a retry policy
         self.reconnect_delay = 10
+        self.waiting_ops = []
 
     @pipeline_thread.runs_on_pipeline_thread
     def _run_op(self, op):
         if isinstance(op, pipeline_ops_base.ConnectOperation):
-            self.virtually_connected = True
-            self.send_op_down(op)
+            if self.state in [ReconnectState.waiting_to_reconnect]:
```
olivakar: >in [](start = 26, length = 2)

will equals not work ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -1961,147 +2213,146 @@ def stage(self, mocker, cls_type, init_kwargs, connected, virtually_connected, h
...
         stage.handle_pipeline_event(event)
+        if old_timer:
+            assert old_timer.cancel.call_count == 1
+        assert stage.reconnect_timer != old_timer
 
-        # Original timer is cancelled
-        assert original_timer.cancel.call_count == 1
-        assert original_timer.cancel.call_args == mocker.call()
+    @pytest.mark.it(
+        "If not previosly connected, does not clears the previous reconnect timer if there was one"
```
olivakar: >previosly [](start = 16, length = 9)

spell

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,146 @@ def do_retry():
...
+
 class ReconnectStage(PipelineStage):
     def __init__(self):
         super(ReconnectStage, self).__init__()
         self.reconnect_timer = None
-        self.virtually_connected = False
+        self.state = ReconnectState.never_connected
         # connect delay is hardcoded for now.  Later, this comes from a retry policy
         self.reconnect_delay = 10
+        self.waiting_ops = []
```
olivakar: >waiting_ops [](start = 13, length = 11)

when i first read this i felt these are ops waiting for connection , but these are actually connect ops which are waiting..can this be renamed ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -814,32 +904,32 @@ def _clear_reconnect_timer(self):
...
-                    logger.info(
-                        "{}: disconnected, but not virtually connected.  Not triggering reconnect timer.".format(
-                            self.name
-                        )
-                    )
-            self.send_event_up(event)
-
-        else:
-            self.send_event_up(event)
+    def _complete_waiting_ops(self, error=None):
```
olivakar: >_complete_waiting_ops [](start = 8, length = 21)

can this be renamed too ? 
P.S. check the comment for renaming the list

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I feel as though these probably don't really need to be in a class at all, although it's not a big deal either way

---
In reply to: [377910185](https://github.com/Azure/azure-iot-sdk-python/pull/460#discussion_r377910185) [](ancestors = 377910185)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Add this information to the class docstring perhaps?

---
In reply to: [377748902](https://github.com/Azure/azure-iot-sdk-python/pull/460#discussion_r377748902) [](ancestors = 377748902)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+    CONNECTED_OR_DISCONNECTED: The transport is either connected or disconencted.  This stage doesn't really care which one, so
+    it doesn't keep track.
+    """
+
+    NEVER_CONNECTED = "NEVER_CONNECTED"
+    WAITING_TO_RECONNECT = "WAITING_TO_RECONNECT"
+    CONNECTED_OR_DISCONNECTED = "CONNECTED_OR_DISCONNECTED"
+
+
 class ReconnectStage(PipelineStage):
```
cartertinney: This class needs a docstring

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
 ]
```
cartertinney: Could we move this inside the ReconnectState class, since that's the only place it's used? I don't like having global definitions hanging around.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+                            "{}({}): State is {}.  Connect failed with transient error. Triggering reconnect timer".format(
+                                self.name, op.name, self.state
+                            )
+                        )
+                        self.state = ReconnectState.WAITING_TO_RECONNECT
+                        self._start_reconnect_timer()
+
+                    elif this.state == ReconnectState.WAITING_TO_RECONNECT:
+                        logger.info(
+                            "{}({}): non-tranient error.  Failing all waiting ops.n".format(
```
cartertinney: typo

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+            if self.state == ReconnectState.WAITING_TO_RECONNECT:
+                logger.info(
+                    "{}({}): State is {}.  Canceling waiting ops and sending disconnect down.".format(
+                        self.name, op.name, self.state
+                    )
+                )
+                self._clear_reconnect_timer()
+                self._complete_waiting_connect_ops(
+                    pipeline_exceptions.OperationCancelled("Explicit disconnect invoked")
+                )
```
cartertinney: It's somewhat odd that this is not a ConnectOperation, but is added to the list of.... ConnectOperations.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+                self.waiting_connect_ops.append(op)
+            else:
+                logger.info(
+                    "{}({}): State is {}.  Adding to wait list and sending new connect op down".format(
+                        self.name, op.name, self.state
+                    )
+                )
+                self.waiting_connect_ops.append(op)
+                self._send_new_connect_op_down()
```
cartertinney: I don't quite follow here.

So, some ConnectOperation comes down the pipe. We are NOT waiting to reconnect, so it is put in the list of operations that are waiting (?), and then a new ConnectOperation is sent down instead?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+    WAITING_TO_RECONNECT: This stage is in a waiting period before reconnecting.
+
+    CONNECTED_OR_DISCONNECTED: The transport is either connected or disconencted.  This stage doesn't really care which one, so
+    it doesn't keep track.
+    """
+
+    NEVER_CONNECTED = "NEVER_CONNECTED"
+    WAITING_TO_RECONNECT = "WAITING_TO_RECONNECT"
+    CONNECTED_OR_DISCONNECTED = "CONNECTED_OR_DISCONNECTED"
+
```
cartertinney: I don't quite get why these are the same state

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -736,32 +731,159 @@ def do_retry():
...
+                                this.name, op.name
+                            )
+                        )
+                        self.state = ReconnectState.CONNECTED_OR_DISCONNECTED
+                        self._clear_reconnect_timer()
+                        this._complete_waiting_connect_ops(error)
+
+                    else:
+                        logger.info(
+                            "{}({}): State is {}.  Connection failed. Not triggering reconnection".format(
```
cartertinney: Seems this log message isn't very useful re: state.

As I can tell, the only way this block triggers is if the state is in CONNECTED_OR_DISCONNECTED. So the state will ALWAYS be that in this case. Furthermore, connected and disconnected are pretty different. Is telling us that it is either connected or disconnected helpful?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Doesn't this exception have to be raised and caught first for proper traceback? Or is that handled elsewhere in the code? I can't quite remember.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -814,32 +917,32 @@ def _clear_reconnect_timer(self):
...
+           way to to complete all of the ops that are patiently waiting for the connection.
+
+        Right now, we only need to do this with ConnectOperation ops because these are the
+        only ops that need to wait because these are the only ops that cause a connection
+        to be established.  Other ops pass through this stage, and might fail in later
+        stages, but that's OK.  If they needed a connection, the AutoConnectStage before
+        this stage should be taking care of that.
+        """
+        logger.info("{}: completing waiting ops with error={}".format(self.name, error))
+        list_copy = self.waiting_connect_ops
```
cartertinney: >list_copy [](start = 8, length = 9)

Technically not a copy.

PR #461, opened 2020-02-11
feature(azure-iot-device):Support proxy via mqtt

olivakar:
Proxy is only working for MQTT via web-sockets with HTTP proxy servers. Hence the samples have been defined accordingly to show case these specific scenarios. We are only going to support proxy via the above option.

However the implementation is general and we are not going to disallow usage of SOCK proxy servers with normal MQTT as well as MQTT web-sockets , but this is left to the user as an exercise.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -7,5 +7,3 @@ pytest-timeout
 mock #remove this as soon as no references to it remain in the code
 flake8
 azure-iothub-provisioningserviceclient >= 1.2.0  # Only needed for end to end tests for DPS
-PySocks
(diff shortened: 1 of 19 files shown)


requirements_test.txt
```diff
@@ -7,3 +7,5 @@ pytest-timeout
 mock #remove this as soon as no references to it remain in the code
 flake8
 azure-iothub-provisioningserviceclient >= 1.2.0  # Only needed for end to end tests for DPS
+PySocks
+win-inet-pton; python_version == '2.7'  # Only needed for python 2
```
BertKleewein: If these are in setup.py, they don't need to be here.

PR #464, opened 2020-02-13
fix: User agent according to specification

olivakar:
Java example user agent is 

For Java I see ClientVersion string like this one:
```
com.microsoft.azure.sdk.iot.iot-device-client/1.20.0 (1.8.0_232; Windows 10; amd64),com.microsoft.azure.sdk.iot.iot-device-client,1.20.0,,1.8.0,Windows 10,,,,amd64,,Java,Device,Windows,,,
```
where the first part is the client OS info which gets parsed to:
•	SDKName = com.microsoft.azure.sdk.iot.iot-device-client
•	SDKVersion = 1.20.0
•	RuntimeName = <blank>
•	RuntimeVersion = 1.8.0
•	OSName = Windows 10
•	OSVersion = <blank>
•	OSSkuId = <blank>
•	HWPlatform = amd64
•	SQMId = <blank>
•	SDKLanguage = Java
•	SDKType = Device
•	OSSimplified = Windows


azure-iot-device/azure/iot/device/product_info.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+
+    @staticmethod
+    def get_iothub_user_agent():
+        """
+        Create the user agent for IotHub
+        """
+        return (
+            "{iothub_iden}/{version}".format(iothub_iden=IOTHUB_IDENTIFIER, version=VERSION)
+            + ProductInfo._get_common_user_agent()
```
BertKleewein: Instead of using format and then concatenating strings, how about 
`
        return (
            "{iothub_iden}/{version}{common}".format(
            iothub_iden=IOTHUB_IDENTIFIER, 
            version=VERSION, 
            common=ProductInfo._get_common_user_agent(),
        ))
`

BertKleewein: I'm mentioning this because the way you have it now, there's a heap allocation that immediately gets freed (the result of the format call).  It's not a big deal on cpython, but it can lead to heap fragmentation.  

---
In reply to: [384152287](https://github.com/Azure/azure-iot-sdk-python/pull/464#discussion_r384152287) [](ancestors = 384152287)

azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py
```diff
@@ -116,7 +116,7 @@ def get_trust_bundle(self):
         r = requests.get(
             self.workload_uri + "trust-bundle",
             params={"api-version": self.api_version},
-            headers={"User-Agent": urllib.parse.quote_plus(constant.USER_AGENT)},
+            headers={"User-Agent": urllib.parse.quote_plus(ProductInfo.get_iothub_user_agent())},
```
BertKleewein: >headers={"User-Agent": urllib.parse.quote_plus(ProductInfo.get_iothub_user_agent())}, [](start = 12, length = 85)

I don't think we need this.  This request only goes to the edge daemon (iotedged), and it never goes over any wires.  Does iotedged need any User-Agent at all?

azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py
```diff
@@ -162,7 +162,7 @@ def sign(self, data_str):
         r = requests.post(  # TODO: can we use json field instead of data?
             url=path,
             params={"api-version": self.api_version},
-            headers={"User-Agent": urllib.parse.quote_plus(constant.USER_AGENT)},
+            headers={"User-Agent": urllib.parse.quote_plus(ProductInfo.get_iothub_user_agent())},
```
BertKleewein: >headers [](start = 12, length = 7)

Same comment.  I don't think this needs any User-Agent.

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_http.py
```diff
@@ -3,34 +3,23 @@
...
 import pytest
 import json
 import sys
 import six.moves.urllib as urllib
-from azure.iot.device.common.pipeline import (
-    pipeline_events_base,
-    pipeline_ops_base,
-    pipeline_stages_base,
-    pipeline_ops_http,
-)
```
BertKleewein: Cool!  How did you know that these imports were unused?

azure-iot-device/azure/iot/device/iothub/auth/iotedge_authentication_provider.py
olivakar: I am not sure from the functional perspective of this. This code was already there. I am guessing there are unit tests also , so i have to be a little careful in removing this....but we can check his from Jelani

---
In reply to: [384155183](https://github.com/Azure/azure-iot-sdk-python/pull/464#discussion_r384155183) [](ancestors = 384155183)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_http.py
olivakar: I have an unused import warning level set in Pycharm i think..I set it long back...

---
In reply to: [384156889](https://github.com/Azure/azure-iot-sdk-python/pull/464#discussion_r384156889) [](ancestors = 384156889)

PR #465, opened 2020-02-14
feature(azure-iot-hub): Implement c2d using uamqp

jebrando:


azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -663,3 +663,16 @@ def invoke_device_module_method(self, device_id, module_id, direct_method_reques
...
+
+    def send_c2d_message(self, device_id, message):
+        """Send a C2D mesage to a IoTHub Device.
+
+        :param str device_id: The name (Id) of the device.
+        :param str message: The message that is to be delievered to the device.
+
+        :raises: Exception if the Send command is not able to send the message
+        """
+        svc_client = iothub_amqp_client(
```
zolvarga: >svc_client = iothub_amqp_client( [](start = 8, length = 32)

Are we creating the amqp_client every time when we send a message? I think the amqp_client should be a member of the class and should be initialized alogn with the protocol client.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+        msg_props = uamqp.message.MessageProperties()
+        msg_props.to = "/devices/{}/messages/devicebound".format(device_id)
+        msg_props.message_id = str(uuid4())
+        message = uamqp.Message(
+            msg_content, properties=msg_props, application_properties=app_properties
+        )
+        self.amqp_client.queue_message(message)
+        results = self.amqp_client.send_all_messages()
+        if uamqp.constants.MessageState.SendFailed in results:
+            raise Exception("amqp Send Failure")
```
zolvarga: >amqp Send Failure [](start = 29, length = 17)

Should be: "Telemetry Send Failure"

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+class IoTHubAmqpClient:
+    def __init__(self, hostname, shared_access_key_name, shared_access_key):
+        self.endpoint = _build_amqp_endpoint(hostname, shared_access_key_name, shared_access_key)
+        operation = "/messages/devicebound"
+        target = "amqps://" + self.endpoint + operation
+        self.amqp_client = uamqp.SendClient(target)
+
+    def send_message_to_device(self, device_id, message):
+        if not self.amqp_client:
+            raise RuntimeError("A Call to connect must be done before")
```
zolvarga: >A Call to connect must be done before [](start = 32, length = 37)

Should be: "Create AMQP client first" (or something)

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+
+import logging
+import os
+import sys
+import base64
+import time
+import hashlib
+import hmac
+from uuid import uuid4
+import urllib
```
olivakar: tests are failing in 2.7 due to absence of a method in this module. 
This should be imported as `import six.moves.urllib as urllib` and i think tests would pass

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+import uamqp
+from uamqp import utils, errors
+
+default_sas_expiry = 30
+
+
+def _generate_auth_token(uri, sas_name, sas_value):
+    """
+    Given a URI, a sas_name, and a sas_value, return a shared access signature.
+    """
```
olivakar: include `:param` as well

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+    expiry = str(int(time.time() + default_sas_expiry))
+    string_to_sign = (uri + "\n" + expiry).encode("utf-8")
+    signed_hmac_sha256 = hmac.HMAC(sas, string_to_sign, hashlib.sha256)
+    signature = urllib.parse.quote(base64.b64encode(signed_hmac_sha256.digest()))
+    return "SharedAccessSignature sr={}&sig={}&se={}&skn={}".format(
+        uri, signature, expiry, sas_name
+    )
+
+
+def _build_amqp_endpoint(hostname, shared_access_key_name, shared_access_key):
```
olivakar: >_build_amqp_endpoint [](start = 4, length = 20)

doc

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -663,3 +663,16 @@ def invoke_device_module_method(self, device_id, module_id, direct_method_reques
         return self.protocol.device_method.invoke_module_method(
             device_id, module_id, direct_method_request
         )
+
+    def send_c2d_message(self, device_id, message):
+        """Send a C2D mesage to a IoTHub Device.
+
+        :param str device_id: The name (Id) of the device.
+        :param str message: The message that is to be delievered to the device.
```
olivakar: >delievered [](start = 54, length = 10)

spell

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
olivakar: >svc_client [](start = 8, length = 10)

may be rename to amqp_svc_client

azure-iot-hub/tests/test_iothub_registry_manager.py
```diff
@@ -1099,3 +1172,15 @@ def test_invoke_device_module_method(
...
+@pytest.mark.describe("IoTHubRegistryManager - .send_c2d_message()")
+class TestSendC2dMessage(object):
+    @pytest.mark.it("Test send c2d message")
+    def test_send_c2d_message(
+        self, mocker, mock_uamqp_send_message_to_device, iothub_registry_manager
+    ):
+
+        iothub_registry_manager.send_c2d_message(fake_device_id, fake_message_to_send)
+
+        assert mock_uamqp_send_message_to_device.call_count == 1
```
olivakar: >assert [](start = 8, length = 6)

you can also assert the call_args here to check device id and mesage

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+from uuid import uuid4
+import urllib
+
+try:
+    from urllib import quote, quote_plus, urlencode  # Py2
+except Exception:
+    from urllib.parse import quote, quote_plus, urlencode
+
+import uamqp
+from uamqp import utils, errors
```
olivakar: unused may be

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
olivakar: >_build_amqp_endpoint [](start = 4, length = 20)

these could be static methods of the client class

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+        app_properties = {}
+        msg_props = uamqp.message.MessageProperties()
+        msg_props.to = "/devices/{}/messages/devicebound".format(device_id)
+        msg_props.message_id = str(uuid4())
+        message = uamqp.Message(
+            msg_content, properties=msg_props, application_properties=app_properties
+        )
+        self.amqp_client.queue_message(message)
+        results = self.amqp_client.send_all_messages()
+        if uamqp.constants.MessageState.SendFailed in results:
```
olivakar: > if uamqp.constants.MessageState.SendFailed in results: [](start = 7, length = 55)

do we raise error if 1 message only failed out of the lot ?

jebrando: As written there will only be one message in the queue, so yes.  This does bring about the question of should we batch messages?  For now I would say no, but maybe a future enhancement.

---
In reply to: [379646170](https://github.com/Azure/azure-iot-sdk-python/pull/465#discussion_r379646170) [](ancestors = 379646170)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -663,3 +663,16 @@ def invoke_device_module_method(self, device_id, module_id, direct_method_reques
         return self.protocol.device_method.invoke_module_method(
             device_id, module_id, direct_method_request
         )
+
+    def send_c2d_message(self, device_id, message):
```
olivakar: >send_c2d_message [](start = 8, length = 16)

This AMQP thing feels to be very separate from the registry manager. Is there a reason why this method should be here. Is this a place holder for a future method ?
Is it possible for client to be created from the connection string directly and use the send message or will that mean we have to keep on supporting that client forever ?

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
jebrando: Moved these to be methods in the class.

---
In reply to: [379600948](https://github.com/Azure/azure-iot-sdk-python/pull/465#discussion_r379600948) [](ancestors = 379600948)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
jebrando: Yes, it would be separate, but it will be removed as soon as the service implements this in the swagger.

---
In reply to: [379664342](https://github.com/Azure/azure-iot-sdk-python/pull/465#discussion_r379664342) [](ancestors = 379664342)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -663,3 +671,14 @@ def invoke_device_module_method(self, device_id, module_id, direct_method_reques
...
+        """Send a C2D mesage to a IoTHub Device.
+
+        :param str device_id: The name (Id) of the device.
+        :param str message: The message that is to be delievered to the device.
+
+        :raises: Exception if the Send command is not able to send the message
+        """
+
+        self.amqp_svc_client.send_message_to_device(device_id, message)
```
zolvarga: I think an explicit return is missing here

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -63,9 +63,17 @@ def __init__(self, connection_string):
...
         :rtype: :class:`azure.iot.hub.IoTHubRegistryManager`
         """
-
         self.auth = ConnectionStringAuthentication(connection_string)
         self.protocol = protocol_client(self.auth, "https://" + self.auth["HostName"])
+        self.amqp_svc_client = iothub_amqp_client(
+            self.auth["HostName"], self.auth["SharedAccessKeyName"], self.auth["SharedAccessKey"]
+        )
+
+    def __del__(self):
```
BertKleewein: >def __del__(self): [](start = 4, length = 18)

Random unrelated note: __del__ does not exist on micropython.  Luckily we don't use this too much.

azure-iot-hub/tests/test_iothub_amqp_client.py
```diff
@@ -0,0 +1,72 @@
...
+import copy
+import logging
+import uamqp
+from azure.iot.hub.iothub_amqp_client import IoTHubAmqpClient
+
+try:
+    from urllib import quote, quote_plus, urlencode  # Py2
+except Exception:
+    from urllib.parse import quote, quote_plus, urlencode
+
```
BertKleewein: nit: this is why we have the six library.  preferred style is to say "import six.moves.urllib as urllib" then access these as "urllib.parse.quote" in the code.  This way you don't have to try/except to get py2/py3 compatibility.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -0,0 +1,83 @@
...
+import time
+import hashlib
+import hmac
+from uuid import uuid4
+import six.moves.urllib as urllib
+
+try:
+    from urllib import quote, quote_plus, urlencode  # Py2
+except Exception:
+    from urllib.parse import quote, quote_plus, urlencode
```
BertKleewein: you don't need this if you're using six above.  Just reference these with urllib.parse.quote, etc in the code.  Or keep the 'from urllib.parse import quote, quote_plus, urlencode' part of this.  After using six, the py2 block is dead because six makes the py2 library match the py3 signature.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
jebrando: We are not returning anything here by design

---
In reply to: [380997235](https://github.com/Azure/azure-iot-sdk-python/pull/465#discussion_r380997235) [](ancestors = 380997235)

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -63,9 +63,17 @@ def __init__(self, connection_string):
...
+        """
+        Deinitializer for a Registry Manager Service client.
+        """
+        self.amqp_svc_client.disconnect_sync()
```
Garcia6l20: Should be None checked

lastcoolnameleft:
Thanks for implementing this (and the sample) as it was a blocker for one of my use cases.  Do you know when it will be an updated on pypi?  (the current version 2.0.2 doesn't have this)

PR #466, opened 2020-02-19
PNP sample changes after first review

BertKleewein:


azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -26,158 +28,154 @@
...
 
 # Note: this follows the method pattern from the pythonIoTHubDeviceClient object.
-async def environmental_command_listener(interface):
+async def environmental_model_listener(interface):
     while True:
-        command_request = await interface.receive_pnp_command()
+        incoming_message = await interface.receive_pnp_message()
 
-        if command_request.command_name == "blink":
+        if incoming_message.target == interface.commands.blink:
```
BertKleewein: >if incoming_message.target == interface.commands.blink: [](start = 8, length = 55)

Note: the incoming_message object could be a CommandRequest or it could be a PropertyUpdateRequest.  They both have a target attribute (and probably come from the same base class?)

azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -26,158 +28,154 @@
...
             else:
                 command_update = CommandUpdate.create_from_command_request(
-                    command_request, 200, "runDiagnostics update response"
+                    incoming_message, 200, "runDiagnostics update response"
                 )
                 try:
                     await interface.send_command_update(command_update)
                 except Exception as e:
                     print("Got an error on the update: {}".format(e))
+        elif isinstance(incoming_message, PropertyUpdateRequest):
```
BertKleewein: >lif isinstance(incoming_message, PropertyUpdateRequest): [](start = 9, length = 56)

Note: single listener for commands and property changes.  But, for reusability, each interface needs its own listener.

azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -26,158 +28,154 @@
...
 # Note: this follows the method pattern from the pythonIoTHubDeviceClient object.
-async def environmental_command_listener(interface):
+async def environmental_model_listener(interface):
     while True:
-        command_request = await interface.receive_pnp_command()
+        incoming_message = await interface.receive_pnp_message()
 
-        if command_request.command_name == "blink":
+        if incoming_message.target == interface.commands.blink:
             print("Got the blink command")
```
zolvarga: >print("Got the blink command") [](start = 12, length = 30)

Nitpick: Get the name from the command's name and just use one print after the first line of the while block

azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -193,33 +191,35 @@ def __init__(self, interface_instance_name):
...
 
     #  send telemetry every 5 seconds
     def send_telemetry():
         while True:
-            await environmental_sensor.sendTelemetry(
-                {"temp": 10 + random.random_int(0, 90), "humid": 1 + random.randint(0, 99)}
-            )
+            environmental_sensor.telemetry.temp = 10 + random.random_int(0, 90)
+            environmental_sensor.telemetry.humid = 1 + random.randint(0, 99)
+            await environmental_sensor.report_telemetry()  # only sends telemetry values that have changed
```
zolvarga: >await environmental_sensor.report_telemetry()  # only sends telemetry values that have changed [](start = 12, length = 94)

Don't forget the "force" flag we discussed to update everything regardless if it changed or not

azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -26,158 +28,154 @@
...
-class DeviceInformation(Interface):
+        self.telemetry.temp = Telemetry()
+        self.telemetry.humid = Telemetry()
+        self.properties.state = Property()
+        self.commands.blink = Command()
+        self.commands.turnOff = Command()
+        self.commands.turnOn = Command()
+        self.commands.runDiagnostics = Command()
+        self.properties.name = WriteableProperty()
+        self.properties.brightness = WriteableProperty()
```
jebrando: Would/Could this be generated code?

azure-iot-device/samples/async-pnp-samples/sample.py
```diff
@@ -193,33 +191,35 @@ def __init__(self, interface_instance_name):
...
-    await device_information.processor_architecture.report("4762")
-    await device_information.processor_manufacturer.report("Contoso Foundries")
-    await device_information.total_storage.report("64000")
-    await device_information.total_memory.report("640")
+    environmental_sensor.properties.state = True
+    await environmental_sensor.report_properties()  # only reports changed properties
+
+    device_information.properties.manufacturer = "Contoso Device Corporation"
+    device_information.properties.model = "Contoso 4762B-turbo"
+    device_information.properties.sw_version = "3.1"
```
jebrando: >sw_version [](start = 34, length = 10)

The assumption here is that these will come straight from the DTDL

PR #476, opened 2020-03-03
fix(azure-iot-device): several race condition bugs discovered via stress tests

BertKleewein:
All of these bugs were discovered with stress tests.


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -446,13 +448,10 @@ def disconnect(self):
         self._mqtt_client.loop_stop()
         if rc:
             # This could result in ConnectionDroppedError or ProtocolClientError
+            # No matter what, we always raise here to give upper layers a chance to respond
+            # to this error.
             err = _create_error_from_rc_code(rc)
-            # If we get a ConnectionDroppedError, swallow it, because we have successfully disconnected!
```
BertKleewein: ># If we get a ConnectionDroppedError, swallow it, because we have successfully disconnected! [](start = 12, length = 92)

This was a bad decision.  Yes, we are disconnected, but clients of this class are expecting the _on_disconnected function to be called.  If this code gets hit, then _on_disconnected is probably not going to be called, so we have to return the fact that we're disconnected somehow and this seemed like the most factually accurate.  (news flash: swallowing errors is once again proven to be bad).

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -283,13 +283,41 @@ class AutoConnectStage(PipelineStage):
...
-                    self.name, op.name
+        if op.needs_connection:
+            if self.pipeline_root.connected:
+
+                # If we think we're connected, we pass the op down, but we also check the result.
+                # If we fail the op and we're not connected when we're done, we run through this
+                # stage again to connect.  This is more common than you might think because the
+                # *nix network stack won't detect a dropped connection until the client tries to
+                # send something, so it's very possible that we're disconnected but don't know
+                # it yet.
```
BertKleewein: This is really easy to hit because the pipeline doesn't know that it's disconnected until a ping or a publish fails.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -133,6 +133,17 @@ def _run_op(self, op):
...
+                # some code must think we're still connected.  If we got an exception here,
+                # we're not conencted, and we need to notify upper layers. (Paho should do this,
+                # but it only causes a DisconnectedEvent on manual disconnect or if a PINGRESP
+                # failed, and it's possible to hit this code without either of those things
+                # happening.
+                # BKTODO: it("sends a DisconnectedEvent up if reauthorize_connection fails")
+                if isinstance(e, transport_exceptions.ConnectionDroppedError) or isinstance(
+                    e, transport_exceptions.ConnectionFailedError
+                ):
+                    self.send_event_up(pipeline_events_base.DisconnectedEvent())
```
BertKleewein: >self.send_event_up(pipeline_events_base.DisconnectedEvent()) [](start = 20, length = 60)

This is the third bug.  If reconnect/publish/subscribe/unsubscribe fails because of a connection failure, we need to notify upper stages.  Otherwise the pipeline is not connected when it thinks it is.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -420,7 +420,9 @@ def reauthorize_connection(self, password=None):
         try:
             rc = self._mqtt_client.reconnect()
         except Exception as e:
-            raise exceptions.ProtocolClientError(
+            logger.info("reconnect raised {}".format(e))
+            self._stop_automatic_reconnect()
```
BertKleewein: >_stop_automatic_reconnect [](start = 17, length = 25)

Note to self: need to call _stop_automatic_reconnect on publish/subscribe/unsubscribe failures also

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -252,6 +252,14 @@ def stage(self, mocker, cls_type, init_kwargs):
 )
 
 
+only_disconnected = [pytest.param(False, id="Pipeline Disconnected")]
```
YoDaMa: nit: don't think this is great for test readability... I can see how it helps when writing the tests.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -345,6 +354,133 @@ def test_connect_failure(self, mocker, stage, op, arbitrary_exception):
...
+        stage.run_op(op)
+
+        # if we weren't previously connected, connect and reset mocks
+        if not connected:
+            self.complete_connection_and_reset_mock(stage)
+
+        # make sure our call went down
+        assert stage.send_op_down.call_count == 1
+        assert stage.send_op_down.call_args[0][0] == op
+        stage.send_op_down.reset_mock()
```
YoDaMa: >send_op_down [](start = 14, length = 12)

is it necessary to reset_mock instead of getting the call_args for each call and making sure the call_count is incrementing?

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Is this coming in a future PR then? Why not now?

---
In reply to: [387699653](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r387699653) [](ancestors = 387699653)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -283,13 +283,41 @@ class AutoConnectStage(PipelineStage):
...
+                op.add_callback(check_for_connection_failure)
+                logger.info(
+                    "{}({}): Connected.  Sending down and adding callback to check result".format(
+                        self.name, op.name
+                    )
                 )
-            )
-            self._do_connect(op)
+                self.send_op_down(op)
+            else:
```
cartertinney: >else: [](start = 12, length = 5)

Consider adding a comment clarifying what this else block represents for readability (i.e. needs connection, but not connected)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -345,6 +354,133 @@ def test_connect_failure(self, mocker, stage, op, arbitrary_exception):
         assert op.completed
         assert op.error is arbitrary_exception
 
+    def complete_connection_and_reset_mock(self, stage):
```
olivakar: >complete_connection_and_reset_mock [](start = 8, length = 34)

can u put this somewhere in the beginning of the class

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -345,6 +354,133 @@ def test_connect_failure(self, mocker, stage, op, arbitrary_exception):
...
+        stage.run_op(op)
+
+        # if we weren't previously connected, connect and reset mocks
+        if not connected:
+            self.complete_connection_and_reset_mock(stage)
+
+        # make sure our call went down
+        assert stage.send_op_down.call_count == 1
+        assert stage.send_op_down.call_args[0][0] == op
+        stage.send_op_down.reset_mock()
```
olivakar: >stage.send_op_down.reset_mock() [](start = 8, length = 31)

so if we weren't connected this will happen twice ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -156,7 +166,14 @@ def on_published():
                 logger.debug("{}({}): PUBACK received. completing op.".format(self.name, op.name))
                 op.complete()
 
-            self.transport.publish(topic=op.topic, payload=op.payload, callback=on_published)
+            try:
+                self.transport.publish(topic=op.topic, payload=op.payload, callback=on_published)
+            except (
+                transport_exceptions.ConnectionDroppedError,
+                transport_exceptions.ConnectionFailedError,
```
cartertinney: >transport_exceptions.ConnectionFailedError, [](start = 16, length = 43)

According to documentation in `mqtt_transport.py`, this error is not one of the ones expected to be thrown. Is this a mistake here, or is there a mistake in `mqtt_transport.py`?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -166,7 +183,14 @@ def on_subscribed():
                 logger.debug("{}({}): SUBACK received. completing op.".format(self.name, op.name))
                 op.complete()
 
-            self.transport.subscribe(topic=op.topic, callback=on_subscribed)
+            try:
+                self.transport.subscribe(topic=op.topic, callback=on_subscribed)
+            except (
+                transport_exceptions.ConnectionDroppedError,
+                transport_exceptions.ConnectionFailedError,
+            ):
```
cartertinney: Same as above comment

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -178,7 +202,14 @@ def on_unsubscribed():
                 )
                 op.complete()
 
-            self.transport.unsubscribe(topic=op.topic, callback=on_unsubscribed)
+            try:
+                self.transport.unsubscribe(topic=op.topic, callback=on_unsubscribed)
+            except (
+                transport_exceptions.ConnectionDroppedError,
+                transport_exceptions.ConnectionFailedError,
+            ):
```
cartertinney: Same as above comment

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Also why did this become a ConnectionDroppedError instead of a ProtocolClientError? I don't quite follow

---
In reply to: [387968511](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r387968511) [](ancestors = 387968511,387699653)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
cartertinney: I agree with Yoseph, I did some stuff like this before and it ended up being more of a pain than it was worth to do single parametrization. Best off just breaking down the test cases IMO

---
In reply to: [387896996](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r387896996) [](ancestors = 387896996)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -345,6 +354,133 @@ def test_connect_failure(self, mocker, stage, op, arbitrary_exception):
...
+
+        stage.run_op(op)
+
+        # if we weren't previously connected, connect and reset mocks
+        if not connected:
+            self.complete_connection_and_reset_mock(stage)
+
+        # make sure our call went down
+        assert stage.send_op_down.call_count == 1
+        assert stage.send_op_down.call_args[0][0] == op
```
cartertinney: >== [](start = 50, length = 2)

Use 'is' operator

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -252,6 +252,14 @@ def stage(self, mocker, cls_type, init_kwargs):
...
+only_connected = [pytest.param(True, id="Pipeline Connected")]
+connected_and_disconnected = [
+    pytest.param(True, id="Pipeline Connected"),
+    pytest.param(False, id="Pipeline Disconnected"),
+]
+
+
 @pytest.mark.describe(
     "AutoConnectStage - .run_op() -- Called with an Operation that requires an active connection"
 )
```
cartertinney: Now that we've changed to create two unique flows in this circumstance, (one where we think we're connected, and one where we think we're not) it might be a good idea to split this up into two different testing units for clarity.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: OK, that makes sense.  I was just trying to avoid repeated code, plus I was having trouble phrasing the it.  e.g.  it("Sends down another connect operation if the pipeline was not connected before the operation, then successfully connected, then the op failed and the pipeline was no longer connected after the op failed")

---
In reply to: [388041552](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r388041552) [](ancestors = 388041552,387896996)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Good idea.  that might help with my it() phrasing problem.

---
In reply to: [388044172](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r388044172) [](ancestors = 388044172)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: I think you're right.  I'll go back and check.

---
In reply to: [388025540](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r388025540) [](ancestors = 388025540)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: Why not now?  because I'm not 100% sure this is even a real problem or just theoretical.  I don't want to delay a PR that solves real problems with the possibility that there might be other vaguely related problems.

Why ConnectionDroppedError?  It became ConnectionDroppedError because that matches the behavior we see.  If reconnect() raises, it's most likely because the connection was dropped.  "ProtocolClientError" is a catch-all that means "something went wrong, but we don't know what"

I'm going to go back and see if I can narrow the set of cases where we raise ConnectionDroppedError.  The problem is that we have no real recourse on ProtocolClientError, but we do have a recovery path for ConnectionDroppedError.

Our longer-term goal should be to get rid of ProtocolClientError entirely, or at least reserve it for cases where we know how to remedy the situation (perhaps by rebooting).  Right now, our ProtocolClientError handling goes along the lines of  "something went wrong, but we don't know what, so we're going to return a failure and hope everything is fine."  If we can actually fix the situation that lead to the error (by reconnecting, for instance), we should.  



---
In reply to: [388035883](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r388035883) [](ancestors = 388035883,387968511,387699653)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
cartertinney: Yeah, that's usually the giveaway that something needs to change.

---
In reply to: [388477549](https://github.com/Azure/azure-iot-sdk-python/pull/476#discussion_r388477549) [](ancestors = 388477549,388044172)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -291,12 +289,150 @@ def test_already_connected(self, mocker, stage, op):
...
+        op.complete(arbitrary_exception)
+
+        # assert that we got a new connect operation
+        assert stage.send_op_down.call_count == 2
+        connect_op = stage.send_op_down.call_args[0][0]
+        assert isinstance(connect_op, pipeline_ops_base.ConnectOperation)
+
+    @pytest.mark.it(
+        "Runs the operation a second time if the stage needs to reconnect after the op fails"
+    )
```
cartertinney: Could be good to have a test to show that this works more than just twice, i.e. n times, but not a huge deal

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -345,6 +482,148 @@ def test_connect_failure(self, mocker, stage, op, arbitrary_exception):
...
+
+        # complete the connection
+        connect_op = stage.send_op_down.call_args[0][0]
+        assert isinstance(connect_op, pipeline_ops_base.ConnectOperation)
+        stage.pipeline_root.connected = True
+        connect_op.complete()
+
+        # make sure our call went down
+        assert stage.send_op_down.call_count == 2
+        assert stage.send_op_down.call_args[0][0] is op
```
cartertinney: I'm not a huge fan of the fact that this test doens't do any kind of check for the first time the call goes down. Essentially, we're asserting it's been called twice, but only in order to validate the second one.

I understand that the first call happens as part of the setup, but it still might be good to check that it's already been called once as part of the precondition before the action that sends it into a second call. That's generally the pattern I try to use.

Not a huge deal though tbh.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -291,12 +289,150 @@ def test_already_connected(self, mocker, stage, op):
...
+        # assert that the op failure went back to the caller
+        assert op.completed
+        assert op.error == arbitrary_exception
+        assert callback.call_args == mocker.call(op=op, error=arbitrary_exception)
+
+
+@pytest.mark.describe(
+    "AutoConnectStage - .run_op() -- Called with an Operation that requires an active connection (pipeline not connected)"
+)
+class TestAutoConnectStageRunOpWithOpThatRequiresConnectionNotConnected(
```
olivakar: >TestAutoConnectStageRunOpWithOpThatRequiresConnectionNotConnected [](start = 6, length = 65)

Rename to similar previous class name `TestAutoConnectStageRunOpWithOpThatRequiresConnectionPipelineNotConnected`

olivakar:
>     fake_payload = "__fake_payload__"

does these need to be per class basis ?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:263 in 08c03e2. [](commit_id = 08c03e209836f7a44e8638a352237a13bb75330b, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -283,13 +283,42 @@ class AutoConnectStage(PipelineStage):
...
+                # stage again to connect.  This is more common than you might think because the
+                # *nix network stack won't detect a dropped connection until the client tries to
+                # send something, so it's very possible that we're disconnected but don't know
+                # it yet.
+
+                @pipeline_thread.runs_on_pipeline_thread
+                def check_for_connection_failure(op, error):
+                    if error and not self.pipeline_root.connected:
+                        logger.info(
+                            "{}({}): op failed with {} and we're not conencted.  Re-submitting.".format(
```
olivakar: >conencted [](start = 69, length = 9)

spell

PR #480, opened 2020-03-05
feature(azure-iot-hub): Update to latest swagger.

zolvarga:
Implement job manager.


azure-iot-hub/azure/iot/hub/constant.py
```diff
@@ -6,4 +6,4 @@
 """This module defines constants for use across the azure-iot-hub package
 """
 
-VERSION = "2.1.0"
+VERSION = "2.0.2"
```
jebrando: >2.0.2 [](start = 11, length = 5)

I believe the version needs to be 2.1.1

zolvarga: OK, fixing it

---
In reply to: [389027459](https://github.com/Azure/azure-iot-sdk-python/pull/480#discussion_r389027459) [](ancestors = 389027459)

PR #482, opened 2020-03-06
fix(azure-iot-device): Banned API bug from penetration testing, remove exec from patch, setup

olivakar:
- I have kept the `patch.py` file as of now. In case you guys want to see the output of the patch versus the code.
- I have not removed exec from the unit test file.
- I have removed it from `setup.py`


azure-iot-device/setup.py
```diff
@@ -27,13 +27,18 @@
...
-with open("azure/iot/device/constant.py") as fh:
-    exec(fh.read(), constant)
+
+version = None
+with open("azure/iot/device/constant.py", "r") as fh:
+    for line in fh:
+        if "VERSION" in line:
+            constant, value = line.strip().split("=")
+            version = value.strip()
+            break
```
jebrando: I like it, now what happens if for some reason we remove version from the Constant file.  Will it break or will it fail silently and we not know it?

azure-iot-device/azure/iot/device/abomination.py
```diff
@@ -0,0 +1,227 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module provides hard coded patches used to modify items from the libraries.
+Currently we have to do like this so that we don't use exec anywhere"""
+
+
+def execute_atrocity_for_sync():
```
jebrando: >execute_atrocity_for_sync [](start = 4, length = 25)

I actually Laughed Out Loud

azure-iot-device/setup.py
olivakar: it will fail silently in this case.....for the user agent case it may fail...i have to check that part though

---
In reply to: [389161964](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389161964) [](ancestors = 389161964)

azure-iot-device/azure/iot/device/abomination.py
olivakar: :D

---
In reply to: [389162134](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389162134) [](ancestors = 389162134)

azure-iot-device/setup.py
jebrando: Could we add a check to make sure that the version is not none.

azure-iot-device/azure/iot/device/aio/abomination.py
```diff
@@ -0,0 +1,265 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module provides hard coded patches used to modify items from the libraries.
+Currently we have to do like this so that we don't use exec anywhere"""
+
+
+def execute_atrocity_for_async():
```
cartertinney: I'd prefer if we could keep the atrocities and abominations to one file. The original patch.py worked for both due to the underscore being added to the class name (as you can see here, still copied from the original's output). I don't see why we couldn't do both here as well.

azure-iot-device/azure/iot/device/abomination.py
cartertinney: Where are the processes for using this documented? Before, patch.py documented what it did, but now that we're using it just to generate output, which we then use here, when and how does this get updated?

Is this (and the other abomination) manually maintained? Or do we regenerate from patch.py? And if we do generate from patch.py, could we add a script or something for doing so?

cartertinney: I also laughed, but...

<buzzkill>
this naming probably isn't a great look for us from an open source perspective. 
Just having these files alone is enough of an atrocity on the codebase.
</buzzkill>

---
In reply to: [389197704](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389197704) [](ancestors = 389197704,389162134)

azure-iot-device/setup.py
cartertinney: +1

---
In reply to: [389205255](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389205255) [](ancestors = 389205255)

azure-iot-device/setup.py
```diff
@@ -27,13 +27,18 @@
...
 
-constant = {}
-with open("azure/iot/device/constant.py") as fh:
-    exec(fh.read(), constant)
+
+version = None
+with open("azure/iot/device/constant.py", "r") as fh:
+    for line in fh:
+        if "VERSION" in line:
+            constant, value = line.strip().split("=")
```
cartertinney: This could turn up a false positive if the line contains "VERSION" i.e. we introduce some new variable that also contains the word, which would be bad.

I wonder if it might be better to consider having a separate `version.txt` file that maintains the version that we can read safely here (and not have to deal with any special logic to exclude other values), and then also read the same file into a variable `constant.py` for use elsewhere in the codebase.

Something similar to example 4 in the Single Source Packaging guidelines.
[https://packaging.python.org/guides/single-sourcing-package-version/](https://packaging.python.org/guides/single-sourcing-package-version/)

I don't love this solution either though because I hate the idea of having an extra file.

jebrando: Oliva did mention this earlier, but I didn't like have versions in two places.  There needs to be one source of truth for versions.

azure-iot-device/setup.py
olivakar: I have changed the code a little bit as i realized it would not have covered all bases. Tested with constant file content like :-

```
SOME_VERSION = "322"
IOTHUB_IDENTIFIER = "azure-iot-device-iothub-py"
PROVISIONING_IDENTIFIER = "azure-iot-device-provisioning-py"
VERSION = "2.1.0"
IOTHUB_API_VERSION = "2018-06-30"
TELEMETRY_MESSAGE_SIZE_LIMIT = 262144
```

---
In reply to: [389206423](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389206423) [](ancestors = 389206423,389205255)

olivakar: I have changed the code a little bit as i realized it would not have covered all bases. Tested with constant file content like :-

```
SOME_VERSION = "322"
IOTHUB_IDENTIFIER = "azure-iot-device-iothub-py"
PROVISIONING_IDENTIFIER = "azure-iot-device-provisioning-py"
VERSION = "2.1.0"
IOTHUB_API_VERSION = "2018-06-30"
TELEMETRY_MESSAGE_SIZE_LIMIT = 262144
```

---
In reply to: [389197616](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389197616) [](ancestors = 389197616,389161964)

azure-iot-device/azure/iot/device/abomination.py
olivakar: i feel we should keep patch around coz we need the output of patch to append onto these 2 files

---
In reply to: [389206044](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389206044) [](ancestors = 389206044)

azure-iot-device/azure/iot/device/aio/abomination.py
olivakar: yes precisely u guessed the reason. It was in 1 file but cant be..

---
In reply to: [389205328](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389205328) [](ancestors = 389205328)

azure-iot-device/setup.py
```diff
@@ -27,13 +28,27 @@
...
     _long_description = fh.read()
 
-constant = {}
-with open("azure/iot/device/constant.py") as fh:
-    exec(fh.read(), constant)
+
+filename = "azure/iot/device/constant.py"
+version = None
+
+with open(filename, "r") as fh:
```
BertKleewein: You're reading constant.py twice, once to make sure version is included and again to extract version.  Why not just check to make sure version is not None after you extract it using re.search?

azure-iot-hub/setup.py
```diff
@@ -27,13 +28,26 @@
 with open("README.md", "r") as fh:
     _long_description = fh.read()
 
-constant = {}
-with open("azure/iot/hub/constant.py") as fh:
-    exec(fh.read(), constant)
+filename = "azure/iot/hub/constant.py"
+version = None
+
+with open(filename, "r") as fh:
```
BertKleewein: >open [](start = 5, length = 4)

Same comment on reading the file twice.

BertKleewein:
I notice that you're not actually removing the exec() call from our code.  I know that it's not being called anymore, but will this satisfy the tool that flagged this change to begin with?  You may need to comment out the exec() calls from patch.py and just leave the code that prints the patch to the console.

azure-iot-device/azure/iot/device/__init__.py
```diff
@@ -10,16 +10,17 @@
...
 import sys
 
 if sys.version_info > (3, 5):  # This only works for python 3.5+ at present
     # Dynamically patch the clients to add shim implementations for all the inherited methods.
     # This is necessary to generate accurate online docs.
     # It SHOULD not impact the functionality of the methods themselves in any way.
-    patch.add_shims_for_inherited_methods(IoTHubDeviceClient)  # noqa: F405
-    patch.add_shims_for_inherited_methods(IoTHubModuleClient)  # noqa: F405
-    patch.add_shims_for_inherited_methods(ProvisioningDeviceClient)  # noqa: F405
+    # patch.add_shims_for_inherited_methods(IoTHubDeviceClient)  # noqa: F405
```
BertKleewein: >patch [](start = 6, length = 5)

Suggest adding a short comment on how to re-generate patch_documentation.py in both places.  Sure, it's easy to do now, but future-us might need some help.

BertKleewein:
marking as approved with the assumption that you'll remove these calls if you need to.  I strongly suspect that you'll need to since I assume their tool is just searching the code for offending calls and wouldn't know that the code isn't actually called.  

---
In reply to: [597667841](https://github.com/Azure/azure-iot-sdk-python/pull/482#issuecomment-597667841) [](ancestors = 597667841)

azure-iot-device/setup.py
cartertinney: +1

---
In reply to: [391009563](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r391009563) [](ancestors = 391009563)

azure-iot-device/setup.py
cartertinney: It would actually still be a single source of truth. Version would only be maintained in `version.txt`, and would be dynamically imported from that single source into `constant.py` (where the non-version constants are defined), and then all code references would refer to `constant.py`

---
In reply to: [389742007](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r389742007) [](ancestors = 389742007)

olivakar:
i will delete patch before merge. We have to also think how to keep patch around , so that we can get the output anytime we want

---
In reply to: [597669394](https://github.com/Azure/azure-iot-sdk-python/pull/482#issuecomment-597669394) [](ancestors = 597669394,597667841)

azure-iot-device/setup.py
olivakar: there should not be naming conflict right now as i shifted to regex...precisely for this reason...

---
In reply to: [391159201](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r391159201) [](ancestors = 391159201,389742007)

azure-iot-device/setup.py
olivakar: just wanted to fail fast in case not there...i originally had it together line by line...

---
In reply to: [391157717](https://github.com/Azure/azure-iot-sdk-python/pull/482#discussion_r391157717) [](ancestors = 391157717,391009563)

PR #490, opened 2020-03-13
fix(azure-iot-device): fixed parametrization

olivakar:
Let this PR go first to fix other stuff

### DIFF
--- a/azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
+++ b/azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
@@ -354,7 +354,7 @@ pipeline_stage_test.add_base_pipeline_stage_tests(
 
 @pytest.mark.describe("RegistrationStage - .run_op() -- called with RegisterOperation")
 class TestRegistrationStageWithRegisterOperation(StageRunOpTestBase, RegistrationStageConfig):
-    @pytest.fixture(params=[" ", fake_payload])
+    @pytest.fixture(params=[" ", fake_payload], ids=["empty payload", "some payload"])
     def request_payload(self, request):
         return request.param
 
@@ -404,7 +404,7 @@ class TestRegistrationStageWithArbitraryOperation(StageRunOpTestBase, Registrati
     "RegistrationStage - EVENT: RequestAndResponseOperation created from RegisterOperation is completed"
 )
 class TestRegistrationStageWithRegisterOperationCompleted(RegistrationStageConfig):
-    @pytest.fixture(params=[" ", fake_payload])
+    @pytest.fixture(params=[" ", fake_payload], ids=["empty payload", "some payload"])
     def request_payload(self, request):
         return request.param
 
@@ -640,7 +640,7 @@ class RetryStageConfig(object):
 
 @pytest.mark.describe("RegistrationStage - .run_op() -- retried again with RegisterOperation")
 class TestRegistrationStageWithRetryOfRegisterOperation(RetryStageConfig):
-    @pytest.fixture(params=[" ", fake_payload])
+    @pytest.fixture(params=[" ", fake_payload], ids=["empty payload", "some payload"])
     def request_payload(self, request):
         return request.param
 
--- a/azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
+++ b/azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
@@ -352,12 +352,12 @@ pipeline_stage_test.add_base_pipeline_stage_tests(
 )
 
 
-@pytest.mark.parametrize(
-    "request_payload",
-    [pytest.param(" ", id="empty payload"), pytest.param(fake_payload, id="some payload")],
-)
 @pytest.mark.describe("RegistrationStage - .run_op() -- called with RegisterOperation")
 class TestRegistrationStageWithRegisterOperation(StageRunOpTestBase, RegistrationStageConfig):
+    @pytest.fixture(params=[" ", fake_payload])
+    def request_payload(self, request):
+        return request.param
+
     @pytest.fixture
     def op(self, stage, mocker, request_payload):
         op = pipeline_ops_provisioning.RegisterOperation(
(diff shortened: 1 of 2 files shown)


jebrando:
>     # TODO: revist these tests after auth revision

Can you please remove the word TODO.  This gets flagged during PIN testing

---
Refers to: azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py:189 in 535e6ca. [](commit_id = 535e6cad1e4b7b90e6ec6f66071543acca010c2f, deletion_comment = False)

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
```diff
@@ -639,11 +639,11 @@ def stage(self, mocker, cls_type, init_kwargs):
 
 
 @pytest.mark.describe("RegistrationStage - .run_op() -- retried again with RegisterOperation")
-@pytest.mark.parametrize(
-    "request_payload",
-    [pytest.param(" ", id="empty payload"), pytest.param(fake_payload, id="some payload")],
-)
 class TestRegistrationStageWithRetryOfRegisterOperation(RetryStageConfig):
+    @pytest.fixture(params=[" ", fake_payload])
```
BertKleewein: >fixture [](start = 12, length = 7)

This change looks fine, but I'm curious about what inspired it and why you removed the ids.  At the surface, this looks like a style change, but I know that pytest can be tricky sometimes, especially with parametrization.

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -103,9 +87,11 @@ def assert_for_client_x509(x509):
     assert x509.pass_phrase == fake_pass_phrase
 
 
-@pytest.fixture(scope="function")
-def input_security_client(params_security_clients):
-    return params_security_clients["client_class"](**params_security_clients["init_kwargs"])
+@pytest.fixture(scope="function", params=[x509_security_client, symmetric_key_security_client])
+def input_security_client(request):
+    print(request)
+    print(request.param)
```
BertKleewein: this looks like debug logging.  please remove.

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -103,9 +87,11 @@ def assert_for_client_x509(x509):
...
+    return request.param
```
BertKleewein: >request.param [](start = 11, length = 13)

This change has a very subtle but important effect.  

The old fixture created a new security client object for each test, so each test started out with a blank slate.  

This new fixture creates a single global security client (actually one for x509 and one for symmetric key) and uses that one object for all tests.  This means that each test inherits the security client from the previous test (and all tests before), so tests are no longer independent.

This has 2 implications:
1. If a test fails, it could be because of something that a previous test did to affect the state of the security client.  (if test1 sets some security client attribute to an invalid value, test2 will fail if it runs after test 1)
2. If a test passes, it could be because of something that a previous test did to affect the state of the security client.  (If test 1 initializes the security client, and test 2 does not initialize the security client, test2 will fail if it doesn't run after test 1)

cartertinney: Yup, the security clients themselves ALSO need to be fixtures

---
In reply to: [392240669](https://github.com/Azure/azure-iot-sdk-python/pull/490#discussion_r392240669) [](ancestors = 392240669)

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
cartertinney: Yes, please add the ids back for correct formatting.

---
In reply to: [392234226](https://github.com/Azure/azure-iot-sdk-python/pull/490#discussion_r392234226) [](ancestors = 392234226)

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -209,32 +194,30 @@ def test_pipeline_configuration(self, input_security_client, pipeline_configurat
...
-    ):
+    def test_security_client_success(self, mocker, input_security_client, pipeline_configuration):
         mocker.spy(pipeline_stages_base.PipelineRootStage, "run_op")
         pipeline = ProvisioningPipeline(input_security_client, pipeline_configuration)
 
         op = pipeline._pipeline.run_op.call_args[0][1]
         assert pipeline._pipeline.run_op.call_count == 1
-        assert isinstance(op, params_security_clients["set_args_op_class"])
+        if isinstance(input_security_client, X509SecurityClient):
+            assert isinstance(op, pipeline_ops_provisioning.SetX509SecurityClientOperation)
```
cartertinney: In a normal circumstance I would suggest this is a problem - when you start seeing conditional statements in a test, that tends to be a sign something is wrong.

In this case, the reason is because of these two different security clients having different effects on program flow should probably indicate the test unit should be split to be one for each type of security client.

HOWEVER, because the next task I'm going to be taking on is to unify auths (and by extension, the security clients in provisioning), don't worry about it, fixing this here would be a lot of work to add something I'm just going to end up removing shortly.

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
```diff
@@ -103,9 +87,11 @@ def assert_for_client_x509(x509):
     assert x509.pass_phrase == fake_pass_phrase
 
 
-@pytest.fixture(scope="function")
-def input_security_client(params_security_clients):
-    return params_security_clients["client_class"](**params_security_clients["init_kwargs"])
+@pytest.fixture(scope="function", params=[x509_security_client, symmetric_key_security_client])
+def input_security_client(request):
```
cartertinney: Please add ids to this fixture

azure-iot-device/tests/provisioning/pipeline/test_pipeline_stages_provisioning.py
olivakar: pytest had changed something i am pretty sure for which the tests started failing. Mostly the reason is my prior parameters were tied to the class and even if the fixture using the params was not getting used in certain tests , it was asking me to declare it in those tests as well..this also led to the problem where the parent class needed the params/fixture even if there was no usage of those. now the params are tied to other fixtures , and not to class..

---
In reply to: [392234226](https://github.com/Azure/azure-iot-sdk-python/pull/490#discussion_r392234226) [](ancestors = 392234226)

azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py
olivakar: i could not directly use fixtures , so i have to see how to get them created here. 

---
In reply to: [392403164](https://github.com/Azure/azure-iot-sdk-python/pull/490#discussion_r392403164) [](ancestors = 392403164,392240669)

olivakar:
>     # TODO: revist these tests after auth revision

do u want me to remove right now ? these are TODO after auth revisions.

---
In reply to: [598539876](https://github.com/Azure/azure-iot-sdk-python/pull/490#issuecomment-598539876) [](ancestors = 598539876)

---
Refers to: azure-iot-device/tests/provisioning/pipeline/test_provisioning_pipeline.py:189 in 535e6ca. [](commit_id = 535e6cad1e4b7b90e6ec6f66071543acca010c2f, deletion_comment = False)

PR #491, opened 2020-03-14
feature(azure-iot-device): Added cipher functionality to HTTP pipeline

cartertinney:
* Cipher now works for HTTP operations

### DIFF
--- a/azure-iot-device/azure/iot/device/common/http_transport.py
+++ b/azure-iot-device/azure/iot/device/common/http_transport.py
@@ -27,11 +27,13 @@ class HTTPTransport(object):
 
         :param str hostname: Hostname or IP address of the remote host.
         :param str server_verification_cert: Certificate which can be used to validate a server-side TLS connection (optional).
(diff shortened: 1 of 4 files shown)


azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -45,8 +47,13 @@ def _create_ssl_context(self):
...
         else:
             ssl_context.load_default_certs()
-        ssl_context.verify_mode = ssl.CERT_REQUIRED
-        ssl_context.check_hostname = True
+
+        if self._cipher:
+            try:
+                ssl_context.set_ciphers(self._cipher)
+            except ssl.SSLError as e:
+                # TODO: custom error with more detail?
```
olivakar: >TODO [](start = 18, length = 4)

it seems TODO gets flagged in pen testing. So may be we should come up with different term. and use it consistently across

BertKleewein:
Looks simple enough.  How do we automate E2E testing of this?

PR #493, opened 2020-03-16
tests(azure-iot-device): Enhanced proxy tests

cartertinney:
* Added additional tests for covering proxy cases that were missing
* Adjusted how default kwargs are tested at the API in order to make tests easier to maintain

### DIFF
--- a/azure-iot-device/azure/iot/device/common/pipeline/config.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/config.py
@@ -33,7 +33,6 @@ class BasePipelineConfig(object):
         self.websockets = websockets
         self.cipher = self._sanitize_cipher(cipher)
 
-        # TODO: validate this deeper
         if isinstance(proxy_options, models.ProxyOptions) or proxy_options is None:
             self.proxy_options = proxy_options
         else:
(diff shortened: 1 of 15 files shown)


azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -26,10 +27,17 @@ def __init__(self, websockets=False, cipher="", proxy_options=None):
...
             "OpenSSL cipher list format" or as a list of cipher suite strings.
         :type cipher: str or list(str)
+        :param proxy_options: Details of proxy configuration
+        :type proxy_options: :class:`azure.iot.device.common.models.ProxyOptions`
         """
         self.websockets = websockets
         self.cipher = self._sanitize_cipher(cipher)
-        self.proxy_options = proxy_options
+
+        # TODO: validate this deeper
```
BertKleewein: ># TODO: validate this deeper [](start = 8, length = 28)

What specifically do you want to validate and why?  Doesn't this violate eafp?  (both the changes from this PR and also any additional validation you want to add)

cartertinney: Normally, overly checking something isn't good. However, in this case, there's a lot of weird errors that could happen if this item isn't a ProxyOptions object, since the user provided object gets directly used quite deeply in the stack, and thus we probably want an assurance that it really is a ProxyOptions object.

As far as validating the values in the object itself - you're probably right, we don't need to validate those, so I could remove this TODO

---
In reply to: [393421889](https://github.com/Azure/azure-iot-sdk-python/pull/493#discussion_r393421889) [](ancestors = 393421889)

PR #502, opened 2020-03-19
chore: add github guidelines

elhorton:


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
+**The Azure IoT SDKs team wants to hear from you!**
+
+- [Need Support?](#need-support)
+- [File a bug (code or documentation)](#file-a-bug-code-or-documentation)
+- [Contribute code and/or documentation](#contribute-code-andor-documentation)
+    - [Commit messages](#commit-messages)
+  - [Editing module requirements](#editing-module-requirements)
+  - [Adding new files](#adding-new-files)
+
+# Need Support?
```
jebrando: nitpick: I believe this is not md compliant, but I'm not exactly sure and it doesn't really matter.  we can fix it later after it's checked in (I will run it against the md linter in VsCode)..

PR #506, opened 2020-03-21
fix(azure-iot-device): Spaces can not be part of topic.

olivakar:
unfortunately urlencode does not encode spaces. And keeps them as `+`.
But `+` is not allowed in our topics.

An issue was raised regarding this.


olivakar:
Closing for duplicate.

PR #508, opened 2020-03-23
chore(azure-iot-device): certificate creation separate for python

olivakar:
# Description of the problem
Builds are failing for DPS end to end after each month consistently when the root certificate expires. That is because we use a script from C SDK that expires in 30 days. 

# Description of the solution
A separate script has been added for python which will at least generate certificate that will not expire for another year. This one is only used to create our build environment when we run end to end tests.


scripts/cert_gen.sh
```diff
@@ -0,0 +1,402 @@
...
+    echo "This script is only to help you understand Azure IoT Hub CA Certificates."
+    echo "Use your official, secure mechanisms for this cert generation."
+    echo "Also note that these certs will expire in ${days_till_expire} days."
+    tput sgr0
+}
+
+function generate_root_ca()
+{
+    local common_name="Azure IoT Hub CA Cert Test Only"
+    local password_cmd=" -aes256 -passout pass:${root_ca_password} "
```
tonygleck: Does this line mean you have to have a password on your cert.  If so do we want that?

scripts/cert_gen.sh
```diff
@@ -0,0 +1,402 @@
...
+# gets deleted after the tests.
+# NOTE : In case of 401 error during running DPS E2E tests for certificates
+# the certificates produced by this has expired. And the build environment
+# needs to run again.
+#
+# These certs MUST NOT be used in production.  It is expected that production
+# certificates will be created using a company's proper secure signing process.
+# These certs are intended only to help demonstrate and prototype CA certs.
+###############################################################################
+
```
tonygleck: if you put "set -e" then it will return immediately with the error code so you can call this from another script and on failure it will be able to catch it.

scripts/cert_gen.sh
olivakar: Right now it is very tightly coupled with original script. I did not change it. As of now I would let it be. I want to try something later.

---
In reply to: [396884105](https://github.com/Azure/azure-iot-sdk-python/pull/508#discussion_r396884105) [](ancestors = 396884105)

PR #512, opened 2020-03-25
chore(azure-iot-device) : User choice password

olivakar:
# Description of the problem
A hard coded password was used to create our test certs.

# Description of the solution
User can pass in the password now.


scripts/cert_gen.sh
```diff
@@ -61,6 +60,12 @@ function warn_certs_not_for_production()
 
 function generate_root_ca()
 {
+    if [ $# -ne 1 ]; then
+        echo "Usage: <password>"
+        exit 1
+    fi
+
+    root_ca_password="${1}"
```
jebrando: Good change!

PR #514, opened 2020-03-26
fix(azure-iot-device): MQTT Topic Updates

cartertinney:
* Added test suites for `mqtt_topic_iothub` and `mqtt_topic_provisioning`
* Renamed functions from within `mqtt_topic_iothub` and `mqtt_topic_provisioning` for clarity
* Fixed a bug where whitespace characters could not be sent in message properties
* Removed unused values from `Message`
* Removed usages of `mqtt_topic_iothub` from outside of the desired scope
* Added custom implementation of `urllib.parse.urlencode` for cross version support of key feature
* Added forced string conversion for values that go into MQTT topic strings


azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -185,32 +277,36 @@ def extract_properties_from_topic(topic, message_received):
...
             elif key == "$.uid":
                 message_received.user_id = value
-            elif key == "$.to":
-                message_received.to = value
             elif key == "$.ct":
                 message_received.content_type = value
             elif key == "$.ce":
                 message_received.content_encoding = value
+            elif key == "$.exp":
+                message_received.expiry_time_utc = value
```
cartertinney: We also could potentially get rid of this value if we see no purpose for it.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -60,14 +61,17 @@ def _run_op(self, op):
             # The customer user agent string would simply be appended to the end of this username, in URL Encoded format.
             query_param_seq = [
                 ("api-version", pkg_constant.IOTHUB_API_VERSION),
-                ("DeviceClientType", ProductInfo.get_iothub_user_agent()),
+                (
+                    "DeviceClientType",
+                    ProductInfo.get_iothub_user_agent()
+                    + str(self.pipeline_root.pipeline_configuration.product_info),
```
BertKleewein: >+ str(self.pipeline_root.pipeline_configuration.product_info), [](start = 20, length = 62)

Is there any behavior change when you change the code to append product_info _before_ URL encoding?  Your code looks correct, I'm just wondering if you changed this to make it more correct or did you also intend a change in behavior?

cartertinney: There is no change. This is just cleaner and clearer to the relationship between user agent and product info.

---
In reply to: [399580837](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399580837) [](ancestors = 399580837)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -7,35 +7,39 @@
...
 
+# NOTE: Whenever using standard URL encoding via the urllib.parse.quote() API
+# make sure to specify that there are NO safe values (e.g. safe=""). By default
+# "/" is skipped in encoding, and that is not desirable.
+#
+# DO NOT use urllib.parse.quote_plus(), as it turns ' ' characters into '+',
+# which is invalid for MQTT publishes.
+#
+# DO NOT use urllib.parse.unquote_plus(). as it turns '+' characters into ' ',
+# which is invalid.
```
BertKleewein: Great comment!  This explains so much.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -46,22 +50,65 @@ def get_input_topic_for_subscribe(device_id, module_id):
...
+    """
+    return _get_topic_base(device_id, module_id) + "/messages/events/"
+
+
 def get_method_topic_for_publish(request_id, status):
     """
     :return: The topic for publishing method responses. It is of the format
     "$iothub/methods/res/<status>/?$rid=<requestId>
+
+    Note that all inputs MUST be strings (not ints!)
```
BertKleewein: >Note that all inputs MUST be strings (not ints!) [](start = 4, length = 48)

Why?  What breaks?

azure-iot-device/azure/iot/device/common/version_compat.py
```diff
@@ -0,0 +1,37 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module defines functions intended for providing compatibility between
+different versions of Python"""
+
+from six.moves import urllib
+
+
+def urlencode(query, quote_via=urllib.parse.quote_plus, safe=""):
```
BertKleewein: >urlencode [](start = 4, length = 9)

style nit: if this exists because we want a consistent way to encode for MQTT topics, would it be better to call this something like mqtt_encode and remove the two optional parameters?  This way we're clearly spelling out that URL-encoding rules and MQTT-topic-encoding rules are very different and the fact that we're using a URL-encode helper for encoding MQTT topic names gets hidden in this one place.

olivakar: >urlencode [](start = 4, length = 9)

also a function of this name already exists in urllib right ? so better to rename for us.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -46,22 +50,65 @@ def get_input_topic_for_subscribe(device_id, module_id):
...
+# This is the only method that is shared. Would probably simplify code if it was split.
+# Please consider refactoring.
+def get_twin_topic_for_publish(method, resource_location, request_id):
     """
-    :return: The topic for ALL incoming methods. It is of the format
-    "$iothub/methods/POST/#"
+    :return: The topic for publishing twin requests / patches. It is of the format
+    "$iothub/twin/<method><resourceLocation>?$rid=<requestId>
+
+    Note that all inputs MUST be strings (not ints!)
```
olivakar: >not ints [](start = 42, length = 8)

are these being specified for the request ids ? should not these be then for device ids and module ids also ?

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -60,14 +61,17 @@ def _run_op(self, op):
             # The customer user agent string would simply be appended to the end of this username, in URL Encoded format.
             query_param_seq = [
                 ("api-version", pkg_constant.IOTHUB_API_VERSION),
-                ("DeviceClientType", ProductInfo.get_iothub_user_agent()),
+                (
+                    "DeviceClientType",
```
olivakar: >DeviceClientType [](start = 21, length = 16)

so previously user passed product info was not set here

azure-iot-device/azure/iot/device/provisioning/pipeline/mqtt_topic_provisioning.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    :return: The topic string used to send a registration. It is of the format
+    "$dps/registrations/PUT/iotdps-register/?$rid=<request_id>
+    """
+    return (_get_topic_base() + "PUT/iotdps-register/?$rid={request_id}").format(
+        request_id=urllib.parse.quote(request_id, safe="")
+    )
+
+
+def get_query_topic_for_publish(request_id, operation_id):
```
olivakar: >get_query_topic_for_publish [](start = 4, length = 27)

do you want to mention about the int vs string thing here as well ?

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+    @pytest.mark.it("Returns the topic for subscribing to Input messages from IoTHub")
+    def test_returns_topic(self):
+        device_id = "my_device"
+        module_id = "my_module"
+        expected_topic = "devices/my_device/modules/my_module/inputs/#"
+        topic = mqtt_topic_iothub.get_input_topic_for_subscribe(device_id, module_id)
+        assert topic == expected_topic
+
+    @pytest.mark.it("URL encodes the device_id and module_id when generating the topic")
+    @pytest.mark.parametrize(
```
BertKleewein: >parametrize [](start = 17, length = 11)

If we're testing + on decode, shouldn't we be testing + on encode also?  Here an everywhere,

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+class TestGetMethodNameFromTopic(object):
+    @pytest.mark.it("Returns the method name from a method topic")
+    def test_valid_method_topic(self):
+        topic = "$iothub/methods/POST/fake_method/?$rid=1"
+        expected_method_name = "fake_method"
+
+        assert mqtt_topic_iothub.get_method_name_from_topic(topic) == expected_method_name
+
+    @pytest.mark.it("URL decodes the returned method name")
+    def test_url_decodes_value(self):
```
BertKleewein: >test_url_decodes_value [](start = 8, length = 22)

this only tests $, do we need to test the others (space, slash, plus)?

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+    @pytest.mark.it("Raises a ValueError if the provided topic is not a method topic")
+    @pytest.mark.parametrize(
+        "topic",
+        [
+            pytest.param("not a topic", id="Not a topic"),
+            pytest.param(
+                "devices/fake_device/modules/fake_module/inputs/fake_input",
+                id="Topic of wrong type",
+            ),
+            pytest.param("$iothub/methdos/POST/fake_method/?$rid=1", id="Malformed topic"),
```
BertKleewein: >pytest.param("$iothub/methdos/POST/fake_method/?$rid=1", id="Malformed topic"), [](start = 12, length = 79)

This doesn't look particularly malformed to me.

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+    @pytest.mark.it("Raises a ValueError if the provided topic is not a twin response topic")
+    @pytest.mark.parametrize(
+        "topic",
+        [
+            pytest.param("not a topic", id="Not a topic"),
+            pytest.param(
+                "devices/fake_device/modules/fake_module/inputs/fake_input",
+                id="Topic of wrong type",
+            ),
+            pytest.param("$iothub/twn/res/200?rid=1", id="Malformed topic"),
```
BertKleewein: >Malformed [](start = 58, length = 9)

OK, I can see why this one is malformed.  Maybe add how it's malformed to the id (e.g. "Malformed topic (missing final slash)").  Here and everywhere.

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+        ],
+    )
+    def test_invalid_twin_response_topic(self, topic):
+        with pytest.raises(ValueError):
+            mqtt_topic_iothub.get_twin_request_id_from_topic(topic)
+
+
+# CT-TODO: add input message topic tests for all tests in this class
```
BertKleewein: You asked about example topics in teams.  Ping me again if you want some.

azure-iot-device/tests/provisioning/pipeline/test_mqtt_topic_provisioning.py
```diff
@@ -0,0 +1,257 @@
...
+import pytest
+import logging
+from azure.iot.device.provisioning.pipeline import mqtt_topic_provisioning
+
+logging.basicConfig(level=logging.DEBUG)
+
+# NOTE: All tests (that require it) are parametrized with multiple values for URL encoding.
+# This is to show that the URL encoding is done correctly - not all URL encoding encodes
+# the same way. We must always test the ' ' and '/' characters specifically, in addition
+# to a generic URL encoding value (e.g. $, #, etc.)
```
BertKleewein: Do we not test decode (or encode) of + in this file?  The other big test file has a comment in the header about + but it's missing in this file.

It feels like we should just always test + for encode and decode so future developers don't need to think about "do I need to test + here or not".  If we just have every encode/decode test run against some list (maybe ' ', '$', '#', '/', '+', and, just for yucks, '\')

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+# NOTE: All tests (that require it) are parametrized with multiple values for URL encoding.
+# This is to show that the URL encoding is done correctly - not all URL encoding encodes
+# the same way. We must always test the ' ' and '/' characters specifically, in addition
+# to a generic URL encoding value (e.g. $, #, etc.)
+#
+# For URL decoding, we must always test the '+' character speicifically, in addition to
+# a generic URL encoding value (e.g. $, #, etc.)
```
olivakar: >encoding [](start = 16, length = 8)

did u mean decoding ?

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+#
+# PLEASE DO THESE TESTS FOR EVEN CASES WHERE THOSE CHARACTERS SHOULD NOT OCCUR, FOR SAFETY.
+
+
+@pytest.mark.describe(".get_c2d_topic_for_subscribe()")
+class TestGetC2DTopicForSubscribe(object):
+    @pytest.mark.it("Returns the topic for subscribing to C2D messages from IoTHub")
+    def test_returns_topic(self):
+        device_id = "my_device"
+        expected_topic = "devices/my_device/messages/devicebound/#"
```
olivakar: >my_device [](start = 34, length = 9)

may use device id param defined above

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
```diff
@@ -0,0 +1,1014 @@
...
+                "fake?device",
+                id="Standard URL encoding required for device_id",
+            ),
+            pytest.param(
+                "devices/fake%20device/messages/devicebound/%24.mid=6b822696-f75a-46f5-8b02-0680db65abf5&%24.to=%2Fdevices%2Ffake%20device%2Fmessages%2Fdevicebound",
+                "fake device",
+                id="URL encoding of ' ' character required for device_id",
+            ),
+            # Note that this topic string is completely broken, even beyond the fact that device id's can't have a '/' in them.
+            # A device id with a '/' would not be possible to unencode correctly, because the '/' in the device name encoded in the
```
olivakar: >unencode [](start = 62, length = 8)

spell

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
olivakar: `methdos` and not `methods` so malformed 

---
In reply to: [399807652](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399807652) [](ancestors = 399807652)

azure-iot-device/azure/iot/device/common/version_compat.py
cartertinney: The idea here is that this is literally the same as urllib.parse.urlencode, it just allows for us to use the "quote_via" parameter even in Python 2. Kind of similar to how we reimplement certain functions with the same name for async compatibility purposes in `async_compat.py`, the idea was here that we could reimplement for compatibility purposes across versions.

The idea of changing this to be tied to MQTT semantically is an interesting one though.

---
In reply to: [399663014](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399663014) [](ancestors = 399663014)

azure-iot-device/tests/provisioning/pipeline/test_mqtt_topic_provisioning.py
cartertinney: No, counterintuitively we don't NEED to test for '+' for the following reason:

A '+' will always encode into %2B, and %2B will always decode to '+' no matter what style of URL encoding you do. 

HOWEVER, if using the plus style encoding, a ' ' will encode into a '+', and a '+' will decode back to a ' '.
In regular style encoding, a ' ' will encode into '%20', and then '%20' will decode back to a ' '

Furthermore, the extra consideration in Python is around the '/' due to the way that `quote_plus` by default considers '/' a "safety" character that is ignored. Wheras the standard `quote` does not, and so to ensure we are actually encoding as we desire, we must test it.

---
In reply to: [399808971](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399808971) [](ancestors = 399808971)

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
cartertinney: Yeah, it's subtle, but it is malformed with a typo.

---
In reply to: [399807652](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399807652) [](ancestors = 399807652)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
cartertinney: The specification here is due to the status code. Something like 200, 501, 404, etc. Naturally those are probably numbers instead of strings.

Better idea is probably just to type convert in here though, I think I'll change that.

---
In reply to: [399725688](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399725688) [](ancestors = 399725688)

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
cartertinney: Not really - it's important we test the weird values on encode, but on decode, things are a lot simpler. The only variance is on the '+' character, but since '+' is impossible in MQTT....... it's not really necessary IMO.

That said I could add it for completeness.

---
In reply to: [399805654](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399805654) [](ancestors = 399805654)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
cartertinney: Yeah, it was set differently down below. There has been no behavioural change, I just simplified the code a little bit so it does this all at once, and the relationship is clear.

---
In reply to: [399728411](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399728411) [](ancestors = 399728411)

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
cartertinney: It's a typo of "twin" as "twn"

---
In reply to: [399807979](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399807979) [](ancestors = 399807979)

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
cartertinney: I actually managed to get this myself. I'll remove this todo tho

---
In reply to: [399808130](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399808130) [](ancestors = 399808130)

azure-iot-device/azure/iot/device/common/version_compat.py
cartertinney: The idea here is that this is the exact same function as the one in urllib. Except this one is compatible with Python 2. Kind of like how we do aio compat in aio_compat.py

---
In reply to: [399724837](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399724837) [](ancestors = 399724837)

azure-iot-device/tests/iothub/pipeline/test_mqtt_topic_iothub.py
cartertinney: For now I'm going to pass on making these more detailed, because it opens the door to testing tons of different types of malformed topics.... which are quite honestly not really valid, because we receive them from service.

It also will then get into how poorly we really check malformed topics at present.

If this ever becomes relevant, we can expand it then, until such time, I think it's best to just keep it simple

---
In reply to: [400405654](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r400405654) [](ancestors = 400405654,399807979)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
cartertinney: `quote` needs a string or byte object, which an int is not.

I have now added forced string conversions to avoid this problem.

---
In reply to: [399660431](https://github.com/Azure/azure-iot-sdk-python/pull/514#discussion_r399660431) [](ancestors = 399660431)

PR #515, opened 2020-03-26
fix(azure-iot-device): file and not data for server verification cert

olivakar:
# Description of the problem
We are talking a path for the kwarg `server_verification_cert` and eventually that translates to `cadata`.  `cadata` actually represents the data in the file. 

# Description of the solution
It should be `cafile` which represents a path to a file.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/mqtt_transport.py
+++ b/azure-iot-device/azure/iot/device/common/mqtt_transport.py
@@ -107,7 +107,7 @@ class MQTTTransport(object):
         :param str client_id: The id of the client connecting to the broker.
         :param str hostname: Hostname or IP address of the remote broker.
         :param str username: Username for login to the remote broker.
-        :param str server_verification_cert: Certificate which can be used to validate a server-side TLS connection (optional).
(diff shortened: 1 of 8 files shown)


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -107,7 +107,7 @@ def __init__(
         :param str client_id: The id of the client connecting to the broker.
         :param str hostname: Hostname or IP address of the remote broker.
         :param str username: Username for login to the remote broker.
-        :param str server_verification_cert: Certificate which can be used to validate a server-side TLS connection (optional).
+        :param str server_verification_cert: The path to the certificate which can be used to validate a server-side TLS connection (optional).
```
jebrando: >path [](start = 49, length = 4)

The full path of the certificate file

cartertinney:
Are we sure this doesn't break anything else? I'm thinking Edge specifically?

olivakar:
Closing because discussion about different way of doing it.
We have to accept cert content to be compatible with  edge. 
Decision was to accept cert content, cert file and directory.

PR #516, opened 2020-03-27
Add watchdog timer for MQTT connect and reconnect operations

BertKleewein:
There are conditions where connect and reconnect operations can timeout and we have no control.  These are very uncommon, but reproducable with the right stress scripts.  

# How did I fix this?

To work around these conditions, I've added a watchdog timer to connect and reconnect operations.  If these operations don't complete within some interval (I picked 10 seconds), the watchdog cancels the operation and disconnects the transport.  This gives the transport a "clean slate" and the pipeline retry/reconnect logic is able to recover.

# Problem 1: Paho conect never returns

In some cases, the call into paho's mqtt_client.connect() never returns or takes a really long time to return.  (this has been seen in the field, but I haven't been able to reproduce this).  This is likely a bug in TLS  negotiation or the network stack.

We have no way to fix this.

# Problem 2: Paho connect succeeds, but CONACK never comes back.

In other cases, connect returns success, but the CONACK never returns and we never get an on_mqtt_disconneted event.  I can reproduce this under stress situations, and it has to do with the fact that paho's loop isn't really entirely 100% thread safe, especially if you're intermixing bad network conditions with sas token renewal and DeviceClient opreations.

We could fix this, but it would involve some combination of replacing Paho's thread with our own and submitting PRs to Paho.  Both of these are really expensive.

# Why add the code here?

I did this in the MQTT stage because I wanted to keep the recovery code as close to the transport as possible.  

I could have used TimeoutStage to detect the timeout, but I would have needed to add a recovery stage to disconnect when timeouts occur.  This is working around behaviors of Paho, so it seemed silly to put into a generic stage.

I could have created a new stage to add watchdog functionality to these 2 ops, but this seemed like overkill.  Stages aren't complex, the the logic of ordering stages is complex because of implied inter-stage dependencies.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/mqtt_transport.py
+++ b/azure-iot-device/azure/iot/device/common/mqtt_transport.py
@@ -210,7 +210,7 @@ class MQTTTransport(object):
             if rc:  # i.e. if there is an error
                 logger.debug("".join(traceback.format_stack()))
                 cause = _create_error_from_rc_code(rc)
-                this._stop_automatic_reconnect()
+                this._cleanup_transport_on_error()
 
             if this.on_mqtt_disconnected_handler:
                 try:
@@ -264,14 +264,20 @@ class MQTTTransport(object):
         mqtt_client.on_publish = on_publish
         mqtt_client.on_message = on_message
 
+        # Set paho automatic-reconnect delay to 2 hours.  Ideally we would turn
+        # paho auto-reconnect off entirely, but this is the best we can do.  Without
+        # this, we run the risk of our auto-reconnect code and the paho auto-reconnect
+        # code conflicting with each other.
+        mqtt_client.reconnect_delay_set(120 * 60)
+
         logger.debug("Created MQTT protocol client, assigned callbacks")
         return mqtt_client
 
-    def _stop_automatic_reconnect(self):
+    def _cleanup_transport_on_error(self):
         """
-        After disconnecting because of an error, Paho will attempt to reconnect (some of the time --
-        this isn't 100% reliable).  We don't want Paho to reconnect because we want to control the
-        timing of the reconnect, so we force the connection closed.
+        After disconnecting because of an error, Paho was designed to keep the loop running and
+        to try reconnecting after the reconnect interval. We don't want Paho to reconnect because
+        we want to control the timing of the reconnect, so we force the loop to stop.
 
         We are relying on intimite knowledge of Paho behavior here.  If this becomes a problem,
         it may be necessary to write our own Paho thread and stop using thread_start()/thread_stop().
@@ -281,7 +287,7 @@ class MQTTTransport(object):
 
         logger.info("Forcing paho disconnect to prevent it from automatically reconnecting")
 
-        # Note: We are calling this inside our on_disconnect() handler, so we are inside the
+        # Note: We are calling this inside our on_disconnect() handler, so we might be inside the
         # Paho thread at this point. This is perfectly valid.  Comments in Paho's client.py
         # loop_forever() function recomment calling disconnect() from a callback to exit the
         # Paho thread/loop.
@@ -297,8 +303,10 @@ class MQTTTransport(object):
         # Finally, because of a bug in Paho, we need to null out the _thread pointer.  This
         # is necessary because the code that sets _thread to None only gets called if you
         # call loop_stop from an external thread (and we're still inside the Paho thread here).
+        if threading.current_thread() == self._mqtt_client._thread:
+            logger.debug("in paho thread.  nulling _thread")
+            self._mqtt_client._thread = None
 
-        self._mqtt_client._thread = None
         logger.debug("Done forcing paho disconnect")
 
     def _create_ssl_context(self):
@@ -367,6 +375,8 @@ class MQTTTransport(object):
                     host=self._hostname, port=8883, keepalive=DEFAULT_KEEPALIVE
                 )
         except socket.error as e:
+            self._cleanup_transport_on_error()
+
             # Only this type will raise a special error
             # To stop it from retrying.
             if (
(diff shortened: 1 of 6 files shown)


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -264,14 +264,20 @@ def on_message(client, userdata, mqtt_message):
         mqtt_client.on_publish = on_publish
         mqtt_client.on_message = on_message
 
+        # Set paho automatic-reconnect delay to 2 hours.  Ideally we would turn
+        # paho auto-reconnect off entirely, but this is the best we can do.  Without
```
cartertinney: >but this is the best we can do. [](start = 44, length = 31)

Seriously? Paho doesn't let you do this??

The more I learn about Paho...

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -281,7 +287,7 @@ def _stop_automatic_reconnect(self):
 
         logger.info("Forcing paho disconnect to prevent it from automatically reconnecting")
 
-        # Note: We are calling this inside our on_disconnect() handler, so we are inside the
+        # Note: We are calling this inside our on_disconnect() handler, so we might be inside the
         # Paho thread at this point. This is perfectly valid.  Comments in Paho's client.py
         # loop_forever() function recomment calling disconnect() from a callback to exit the
```
cartertinney: >recomment [](start = 34, length = 9)

typo

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -264,14 +264,20 @@ def on_message(client, userdata, mqtt_message):
...
+        # this, we run the risk of our auto-reconnect code and the paho auto-reconnect
+        # code conflicting with each other.
+        mqtt_client.reconnect_delay_set(120 * 60)
+
         logger.debug("Created MQTT protocol client, assigned callbacks")
         return mqtt_client
 
-    def _stop_automatic_reconnect(self):
+    def _cleanup_transport_on_error(self):
         """
```
cartertinney: Why the name change?

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -780,6 +883,67 @@ def test_client_returns_failing_rc_code(
...
+        "Sets Paho's _thread to None if disconnect does not raise an exception while running in the Paho thread"
+    )
+    def test_sets_thread_to_none_on_success_in_paho_thread(
+        self, mocker, mock_mqtt_client, transport, mock_paho_thread_current
+    ):
+        transport.disconnect()
+        assert mock_mqtt_client._thread is None
+
+    @pytest.mark.it(
+        "Sets Paho's _thread to None if disconnect raises an exception while running in the Paho thread"
```
cartertinney: >_thread [](start = 21, length = 7)

I generally try to refer to things like this as a "._thread attribute" or something to indicate what the thing we're talking about is. Although depending on the test, it can get a bit too verbose to always do that.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -264,14 +264,20 @@ def on_message(client, userdata, mqtt_message):
         mqtt_client.on_publish = on_publish
         mqtt_client.on_message = on_message
 
+        # Set paho automatic-reconnect delay to 2 hours.  Ideally we would turn
+        # paho auto-reconnect off entirely, but this is the best we can do.  Without
+        # this, we run the risk of our auto-reconnect code and the paho auto-reconnect
+        # code conflicting with each other.
+        mqtt_client.reconnect_delay_set(120 * 60)
```
olivakar: >reconnect_delay_set [](start = 20, length = 19)

any reason for 2 hrs ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -53,12 +58,57 @@ def _cancel_pending_connection_op(self):
...
+        connection_op.watchdog = threading.Timer(WATCHDOG_INTERVAL, watchdog_function)
+        connection_op.watchdog.daemon = True
+        connection_op.watchdog.start()
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _cancel_connection_watchdog(self, op):
+        try:
+            if op.watchdog:
+                logger.debug("{}({}): cancelling watchdog".format(self.name, op.name))
+                op.watchdog.cancel()
```
olivakar: >watchdog [](start = 19, length = 8)

i like the name....but do u want to name it just as watchdog or wathcdog_timer following other conventions ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -53,12 +58,57 @@ def _cancel_pending_connection_op(self):
...
+            this = self_weakref()
+            op = op_weakref()
+            if this and op and this._pending_connection_op is op:
+                logger.info(
+                    "{}({}): Connection watchdog expired.  Cancelling op".format(this.name, op.name)
+                )
+                this.transport.disconnect()
+                if this.pipeline_root.connected:
+                    logger.info(
+                        "{}({}): Pipeline is still connected on watchdog expiration.  Sending DisconnectedEvent".format(
```
olivakar: >watchdog [](start = 64, length = 8)

This message is okay for us. But for other people they may not understand what this means. May be info statements are not relevant for other ppl, but soemthing to consider.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: It sort of does.  You can turn off auto-reconnect by making your own IO thread.  In a perfect world with infinite time, that's what I would have done.

---
In reply to: [399533440](https://github.com/Azure/azure-iot-sdk-python/pull/516#discussion_r399533440) [](ancestors = 399533440)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: 2 hours is completely arbitrary.

---
In reply to: [399582990](https://github.com/Azure/azure-iot-sdk-python/pull/516#discussion_r399582990) [](ancestors = 399582990)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: Usage of this function has expanded.  It no longer exists solely to stop the reconnect.  Most of that is handled by setting the interval to 2 hours.  Now this function is used to reset paho to a clean state (as much as we can, anyway).

---
In reply to: [399534294](https://github.com/Azure/azure-iot-sdk-python/pull/516#discussion_r399534294) [](ancestors = 399534294)

PR #520, opened 2020-03-30
(fix): resend RequestOperation on reconnect

BertKleewein:
If a connection drops after we send a RequestOperation but before we receive the corresponding response, we need to re-send the RequestOperation.


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -581,6 +585,26 @@ def _handle_pipeline_event(self, event):
...
+
+        elif isinstance(event, pipeline_events_base.ConnectedEvent):
+            """
+            If we're reconnecting, send all pending requests down again.  This is necessary
+            because any response that might have been sent by the service was likely lost
+            when the connection dropped.  The fact that the operation is still pending means
+            that we haven't received the response yet, and it's safe to re-send the request
+            because everything we do with this stage is allowed to repeat.  For instance, if
+            we update a twin twice with the same reported properties, the service will not
+            fail, likewise for getting a twin, or registering with the provisioning service.
```
cartertinney: Will it add spurious increments to the $version though?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -581,6 +585,26 @@ def _handle_pipeline_event(self, event):
                         self.name, event.name, event.request_id
                     )
                 )
+
+        elif isinstance(event, pipeline_events_base.ConnectedEvent):
+            """
+            If we're reconnecting, send all pending requests down again.  This is necessary
+            because any response that might have been sent by the service was likely lost
```
cartertinney: >likely [](start = 78, length = 6)

Would "possibly" be more accurate?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -494,58 +494,62 @@ def _run_op(self, op):
...
-                method=op.method,
-                resource_location=op.resource_location,
-                request_body=op.request_body,
-                request_id=request_id,
-                request_type=op.request_type,
-                callback=on_send_request_done,
-                query_params=op.query_params,
-            )
-            self.send_op_down(new_op)
+            self._send_request_down(request_id, op)
```
cartertinney: I don't quite understand why this was moved up, if you could explain that to me I'd appreciate it.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -1455,6 +1455,249 @@ def test_no_matching_request_id(self, mocker, stage, pending_op, event, fake_uui
...
+        stage.pipeline_root = pipeline_stages_base.PipelineRootStage(
+            pipeline_configuration=mocker.MagicMock()
+        )
+        stage.send_event_up = mocker.MagicMock()
+        stage.send_op_down = mocker.MagicMock()
+
+        return stage
+
+    @pytest.mark.it(
+        "Sends down a new RequestOperation if the previous RequestOperation never completed"
```
cartertinney: >previous [](start = 50, length = 8)

pending?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
olivakar: yes me as well

---
In reply to: [400437348](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400437348) [](ancestors = 400437348)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -1455,6 +1455,249 @@ def test_no_matching_request_id(self, mocker, stage, pending_op, event, fake_uui
...
 
 
+@pytest.mark.describe(
+    "CoordinateRequestAndResponseStage - .handle_pipeline_event() -- Called with ConnectedEvent"
+)
+class TestCoordinateRequestAndResponseStageHandlePipelineEventWithConnectedEvent(
+    CoordinateRequestAndResponseStageTestConfig, StageHandlePipelineEventTestBase
+):
+    @pytest.fixture
+    def event(self):
```
olivakar: >event [](start = 8, length = 5)

connected_event

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
olivakar: it is not a completely brand new request operation right ? it is one of the old ones that were already sent before...so may be `sends down a secondary request operation on being connected if the previous one was pending`...not sure of secondary is the right word here...

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -581,6 +585,26 @@ def _handle_pipeline_event(self, event):
...
+            fail, likewise for getting a twin, or registering with the provisioning service.
+            """
+
+            self.send_event_up(event)
+
+            for request_id in self.pending_responses:
+                logger.info(
+                    "{}: ConnectedEvent: re-publishing request {}".format(self.name, request_id)
+                )
+                self._send_request_down(request_id, self.pending_responses[request_id])
```
olivakar: >_send_request_down [](start = 21, length = 18)

oh i think u wanted to reuse this method here.....that is why u did the change above...unless i am wrong..then plz explain to me..

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -1455,6 +1455,249 @@ def test_no_matching_request_id(self, mocker, stage, pending_op, event, fake_uui
...
+        # simulate a connected event
+        stage.handle_pipeline_event(event)
+
+        # verify that it got sent again
+        assert stage.send_op_down.call_count == 2
+        assert isinstance(stage.send_op_down.call_args[0][0], pipeline_ops_base.RequestOperation)
+        assert stage.send_op_down.call_args[0][0].request_id == op1_guid
+
+    @pytest.mark.it(
+        "Sends down a new RequestOperation if the previous RequestOperation completed, but no corresponding ResponseEvent was receveived"
```
olivakar: >new [](start = 22, length = 3)

same comment here about the new

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -1455,6 +1455,249 @@ def test_no_matching_request_id(self, mocker, stage, pending_op, event, fake_uui
...
+        assert (
+            stage.send_op_down.call_args_list[2][0][0].request_id == op1_guid
+            and stage.send_op_down.call_args_list[3][0][0].request_id == op2_guid
+        ) or (
+            stage.send_op_down.call_args_list[2][0][0].request_id == op2_guid
+            and stage.send_op_down.call_args_list[3][0][0].request_id == op1_guid
+        )
+
+    @pytest.mark.it(
+        "Sends down multiple RequestOperations if previous RequestOperations completed, but the correspondig ResponseEvents were reveived"
```
olivakar: >reveived [](start = 129, length = 8)

spell

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
olivakar: spell

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
olivakar: >reveived [](start = 129, length = 8)

do u mean `never received` ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
cartertinney: No. It has to be "event" I believe for the inherited tests to work.

---
In reply to: [400444487](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400444487) [](ancestors = 400444487)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: I factored this into its own function because I needed to run the same code when we receive a ConnectedEvent

---
In reply to: [400440435](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400440435) [](ancestors = 400440435,400437348)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: Yes, Carter is correct.  I wanted it to be connected_event, but the inherited tests all broke.

---
In reply to: [400459206](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400459206) [](ancestors = 400459206,400444487)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: You are correct.

---
In reply to: [400450446](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400450446) [](ancestors = 400450446)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Yes, but that's a problem for a different layer.  At this layer, our goal is to ensure delivery.  We can do that, and we can verify it with tests.  Beyond that is uncharted territory.  I'm not sure if there's even a well-defined contract for the behavior of the $version attribute.  I know that it goes up by sometimes, but what happens if the server received a PATCH with an already used $rid?  Surely it _could_ know that this is a duplicate message and behave accordingly, but is it guaranteed to?  What about PATCH operations that don't actually change any twin properties?  Do those update $version even though nothing is actually changing?  I'm pretty sure that they do, but is this by coincidence or by design?  This is a rat's nest and we can't solve it here. 

---
In reply to: [400433963](https://github.com/Azure/azure-iot-sdk-python/pull/520#discussion_r400433963) [](ancestors = 400433963)

PR #521, opened 2020-04-01
fix(azure-iot-device):Move configuration files

olivakar:
# Description of the problem
Less dependence on configuration files from C

# Description of the solution
Moved files to python repo.


scripts/infra_tools/README.md
```diff
@@ -0,0 +1,6 @@
+### Contents of dir
```
jebrando, line 1: Should start with one #

olivakar: i made it heading 3 and not heading 1 for now

PR #527, opened 2020-04-02
chore(azure-iot-device): clean up upload to blob sample code

YoDaMa:
This is to address feedback from @BertKleewein on the upload sample. Should provide better error handling for a host of error responses that could happen with the Storage SDK calls, and slightly cleaner code.

### DIFF
--- a/azure-iot-device/samples/async-hub-scenarios/upload_to_blob.py
+++ b/azure-iot-device/samples/async-hub-scenarios/upload_to_blob.py
@@ -13,16 +13,17 @@ import http.client
 import pprint
 import json
 from azure.storage.blob import BlobServiceClient, BlobClient, ContainerClient
+from azure.core.exceptions import ResourceExistsError
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


azure-iot-device/samples/async-hub-scenarios/upload_to_blob.py
```diff
@@ -13,16 +13,17 @@
...
 from azure.storage.blob import BlobServiceClient, BlobClient, ContainerClient
+from azure.core.exceptions import ResourceExistsError
 import logging
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
 """
-Welcome to the Upload to Blob sample. To use this sample you must have azure.storage.blob installed in your python environment.
+Welcome to the Upload to Blob sample for the Azure IoT Python SDK. To use this sample you must have azure.storage.blob installed in your python environment.
```
cartertinney, line 23: >for the Azure IoT Python SDK [](start = 37, length = 28)

Azure IoT Device Library for Python

PR #535, opened 2020-04-08
chore(pnp): Adding a Environmental sample

olivakar:
PNP Sample


azure-iot-device/samples/pnp/pnp_helper.py
```diff
@@ -0,0 +1,40 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
cartertinney, line 5: Can pnp_helper and pnp_methods be combined?

azure-iot-device/samples/pnp/pnp_methods.py
```diff
@@ -0,0 +1,111 @@
...
+# license information.
+# --------------------------------------------------------------------------
+"""
+This module knows how to convert device SDK functionality into a PNP functionality.
+These methods formats the telemetry, methods, properties to PNP relevant telemetry,
+command requests and pnp properties.
+"""
+from azure.iot.device import Message, MethodResponse
+import json
+from pnp_helper import retrieve_values_dict_from_payload, create_command_response_payload
```
cartertinney, line 13: When importing a module with functions rather than classes, I think it probably best to just import the module and then access the functions via the module.

e.g. import pnp_helper
...

pnp_helper.retrieve_values_dict_from_payload(...)

azure-iot-device/samples/pnp/pnp_environmental_monitor.py
```diff
@@ -0,0 +1,131 @@
...
+# license information.
+# --------------------------------------------------------------------------
+import os
+import asyncio
+import random
+import logging
+
+from azure.iot.device.aio import IoTHubDeviceClient
+from pnp_methods import execute_listener, pnp_update_property, pnp_send_telemetry
+
```
cartertinney, line 13: When importing a module with functions rather than classes, I think it probably best to just import the module and then access the functions via the module. This makes it clearer what is defined in this file, vs what is imported.

e.g. import pnp_methods
...

pnp_methods.pnp_update_property(...)

azure-iot-device/samples/pnp/pnp_helper.py
olivakar: after trying to put everything in 1 file.....i decided to keep a distinction between 2 files...one is the thin layer over device SDK functionality....the other is pure helper methods...

---
In reply to: [408397435](https://github.com/Azure/azure-iot-sdk-python/pull/535#discussion_r408397435) [](ancestors = 408397435)

azure-iot-device/samples/pnp/pnp_methods.py
```diff
@@ -0,0 +1,109 @@
...
+    :param prop_kwargs: The user passed keyword arguments.
+    """
+    # Method 1 : Get only 1 top level dict with a complex class as value.
+    # Will NOT WORK now as internal implementation does not support it yet
+    # prop_dict = construct_properties_from_top_level_dict_and_classes(interface)
+    # await device_client.patch_twin_reported_properties(prop_dict)
+
+    # Method 2 : Get all multi level dictionary. WORKS.
+    # prop_dict = construct_properties_from_all_dict_and_no_classes(interface)
+    # await device_client.patch_twin_reported_properties(prop_dict)
```
BertKleewein, line 61: comments on methods 1+2 confuse things. i think you should remove this.

PR #555, opened 2020-04-29
refactor(azure-iot-device): Auth Revisions

cartertinney:
- Removed auth providers and refactored entire auth flow 
- Unified auth between provisioning and hub clients
- Refactored pipeline to store configuration details on the root
- Refactored translation tests to move to new infrastructure
- Removed all code no longer necessary due to these changes


cartertinney:
./azp run Azure.azure-iot-sdk-python-horton-e2e

cartertinney:
./azp run

cartertinney:
./azp run

azure-iot-device/azure/iot/device/common/auth/sastoken.py
```diff
@@ -0,0 +1,96 @@
...
+
+class SasToken(object):
+    """Shared Access Signature Token used to authenticate a request
+
+    Data Attributes:
+    expiry_time (int): Time that token will expire (in UTC, since epoch)
+    ttl (int): Time to live for the token, in seconds
+    """
+
+    _auth_rule_token_format = (
```
olivakar, line 30: >_auth_rule_token_format [](start = 4, length = 23)

why not  `keyname_token_format` ?

azure-iot-device/azure/iot/device/common/auth/sastoken.py
```diff
@@ -0,0 +1,96 @@
...
+            token = self._simple_token_format.format(
+                resource=url_encoded_uri,
+                signature=url_encoded_signature,
+                expiry=str(self.expiry_time),
+            )
+        return token
+
+    @property
+    def expiry_time(self):
+        "Expiry Time is READ ONLY"
```
olivakar, line 95: >" [](start = 8, length = 1)

are read only doc strings with single quotes instead of 3 ?

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -19,9 +19,28 @@ class BasePipelineConfig(object):
...
+    ):
         """Initializer for BasePipelineConfig
 
+        :param str hostname: The hostname being connected to
+        :param str gateway_hostname: The gateway hostname optionally being used
+        :param sastoken: SasToken to be used for authentication. Mutually exclusive with x509.
+        :type sastoken: :class:`azure.iot.device.common.auth.SasToken`
+        :param x509: X509 to be used for authentication. Mutually exclusive with sastoken.
+        :type x509: :class:`azure.iot.device.models.X509`
+        :param str server_verification_cert: The trusted certificate chain.
```
olivakar, line 41: >:param str server_verification_cert: The trusted certificate chain. [](start = 8, length = 67)

this has nothing to do with you PR...but i just wanted to mention should we update the doc string to say that it is the content of the cert chain...i have faced ppl with some confusion

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -273,6 +273,109 @@ def _handle_pipeline_event(self, event):
...
+            self.send_op_down(op)
+
+    @pipeline_thread.runs_on_pipeline_thread
+    def _cancel_token_renewal_timer(self):
+        """Cancel and delete any pending renewal timer"""
+        timer = self._token_renewal_timer
+        self._token_renewal_timer = None
+        if timer:
+            logger.debug("Cancelling SAS Token renewal timer")
+            timer.cancel()
```
olivakar, line 306: >timer.cancel() [](start = 12, length = 14)

can you also put `timer = None`? For reasons that I don't quite remember i see this pattern in all timers in our stack

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -11,9 +11,15 @@
 import logging
 import os
 import io
-from . import auth
 from . import pipeline
 
+# TODO: different package?
+from azure.iot.device.common.auth import connection_string as cs
```
olivakar, line 17: >cs [](start = 62, length = 2)

may be import as connection and token...just to make more meaningful in code.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub_mqtt.py
```diff
@@ -263,4 +220,4 @@ def _handle_pipeline_event(self, event):
 
         else:
             # all other messages get passed up
-            super(IoTHubMQTTTranslationStage, self)._handle_pipeline_event(event)
+            self.send_event_up(event)
```
olivakar, line 223: >self.send_event_up(event) [](start = 12, length = 25)

oh we missed this

azure-iot-device/azure/iot/device/product_info.py
```diff
@@ -3,6 +3,9 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+"""This module is for creating product identifiers or agent strings for IotHub as well as Provisioning."""
+
+# TODO: Should this be renamed to user_agent.py?
```
BertKleewein, line 8: yes

azure-iot-device/azure/iot/device/common/callable_weak_method.py
```diff
@@ -63,13 +64,18 @@ def __init__(self, object, method_name):
...
     def _get_method(self):
-        return getattr(self.object_weakref(), self.method_name)
+        return getattr(self.object_weakref(), self.method_name, None)
 
     def __call__(self, *args, **kwargs):
         return self._get_method()(*args, **kwargs)
 
     def __eq__(self, other):
         return self._get_method() == other
+        # if inspect.isfunction(other):
```
BertKleewein, line 74: ># if inspect.isfunction(other): [](start = 8, length = 31)

I've been debating this in my head.  I think you should leave the getattr change and revert the other 2 changes.

azure-iot-device/azure/iot/device/common/auth/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .signing_mechanism import SymmetricKeySigningMechanism
+
+# from .sastoken import SasToken
+# from .connection_string import ConnectionString
```
BertKleewein, line 4: please remove

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -273,6 +273,109 @@ def _handle_pipeline_event(self, event):
                 logger.warning("incoming pipeline event with no handler.  dropping.")
 
 
+# NOTE: This stage could be a candidate for being refactored into some kind of other
+# pipeline-related structure. What's odd about it as a stage is that it doesn't really respond
+# to operations or events so much as it spawns them on a timer.
+# Perhaps some kind of... Pipeline Daemon?
+class SasTokenRenewalStage(PipelineStage):
+    # Amount of time, in seconds, prior to token expiration, when the renewal process will begin
+    DEFAULT_TOKEN_RENEWAL_MARGIN = 120
```
BertKleewein, line 282: >DEFAULT_TOKEN_RENEWAL_MARGIN = 120 [](start = 4, length = 34)

why is this set here instead of setting this in the pipeline config with the renewal interval?  part of the purpose of this is to account for clock skew between the client and the server and it's not had to imagine needing a bigger margin if you know that your device clock is going to be inaccurate for some reason or another.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -273,6 +273,109 @@ def _handle_pipeline_event(self, event):
...
+    def _start_renewal_timer(self):
+        """Begin a renewal timer.
+        When the timer expires, and the token is renewed, a new timer will be set"""
+        self._cancel_token_renewal_timer()
+        # NOTE: The assumption here is that the SasToken has just been refreshed, so there is
+        # approximately 'TTL' seconds until expiration. In practice this could probably be off
+        # a few seconds given processing time. We could make this more accurate if SasToken
+        # objects had a live TTL value rather than a static one (i.e. "there are n seconds
+        # remaining in the lifespan of this token", rather than "this token was intended to live
+        # for n seconds")
```
BertKleewein, line 318: Given the possibility of clock skew, TTL is so vague that you're really splitting hairs here. :)

BertKleewein:
I love the clean relationships between sastoken, signing_mechanism, and the renewal logic!  This is a huge improvement!

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -273,6 +273,109 @@ def _handle_pipeline_event(self, event):
...
+                    this.send_op_down(
+                        pipeline_ops_base.ReauthorizeConnectionOperation(
+                            callback=on_reauthorize_complete
+                        )
+                    )
+                # Once again, start a renewal timer
+                this._start_renewal_timer()
+
+            self._token_renewal_timer = threading.Timer(seconds_until_renewal, renew_token)
+            self._token_renewal_timer.daemon = True
```
BertKleewein, line 375: >self._token_renewal_timer.daemon = True [](start = 12, length = 39)

Mental exercise: Consider different token lifetime rules.  why do we create a token on pipeline init and not on connect?  Why do we keep the token renewal timer going after we disconnect?  We don't need the token before we connect and we don't need it after we disconnect. 

This is why I'm asking the question: daemon=True protects us for shutdown scenarios.  In other words, if a daemon timer (thread) is still running when we shut down, it won't stop the app from exiting.  That's good.  

But, daemon=True doesn't stop the timer from living longer than the pipeline.  Consider an app that makes multiple clients, with each client sending a message then dying.  For each client instance, there would be a thread that lives for up to a full hour past the client.  Then, when each timer fires, it silently raises an exception when it tries to dereference `this` and dies.  No harm, but still sloppy and sort-of a memory leak.

azure-iot-device/azure/iot/device/common/auth/sastoken.py
```diff
@@ -0,0 +1,96 @@
...
+        self._signing_mechanism = signing_mechanism
+        self._key_name = key_name
+        self._expiry_time = None  # This will be overwritten by the .refresh() call below
+        self._token = None  # This will be overwritten by the .refresh() call below
+
+        self.ttl = ttl
+        self.refresh()
+
+    def __str__(self):
+        return self._token
```
BertKleewein, line 55: Mental exercise :#2 (written after the one in pipeline_stages_base.py).  

postulate: The notion that a SasToken has a value and that we need to periodically "refresh" that value is not quite right.  It's better to think about the SasToken object as a "password generator" object.  When a consumer of this class needs to talk to a network service, he can use this object to generate a password that gives him access for some period of time.  When that password is set to expire, the consumer is responsible for generating a new one.   This is how the code is written and I like it.

Saying `token.refresh(); password=str(token);` as we do, accomplishes this, but it's not quite semantically valid().  Rather, what we want is `password=token.generate_new_password()`

why this matters:  It doesn't _really_ matter that much, but, there an odd ownership relationship with the root owning the SasToken, object and 2 different pipelines controlling it.  Both MQTT and HTTP can refresh the token.  If we think of the SasToken object as having a value, having someone else change the value feels wrong.  But if we think of each pipeline as owning it's own password that it gets from the generator object, it just feels better.

azure-iot-device/azure/iot/device/common/auth/sastoken.py
cartertinney: Believe it or not.... according to official docs, "auth rule" is the official name. I was also kind of shocked.

azure-iot-device/azure/iot/device/common/auth/sastoken.py
cartertinney: Nope, that's just a mistake.

---
In reply to: [426946876](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426946876) [](ancestors = 426946876)

azure-iot-device/azure/iot/device/common/auth/sastoken.py
cartertinney: Because apparently for some reason the official name of of the `skn` field is not... `shared key name` as you might expect, and is for some reason... `auth rule`.

I don't get it either.

---
In reply to: [426946266](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426946266) [](ancestors = 426946266)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
BertKleewein: I have mixed feelings about renaming to cs and st.  Yes, it's shorter, but this is the only place we do this and I don't want it to get out of control.

---
In reply to: [426962229](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426962229) [](ancestors = 426962229)

cartertinney: So for context, the reason that we do this is that `connection_string` and `sastoken` are kind of abnormal modules (for us, not in general) in that there are associated values in the modules, such as the constants. Thus we need to import the whole module, not just the objects in it.

The problem is that `connection_string` easily gets overloaded, and I didn't want to have to worry about accidentally doing that, so I renamed the package to be something that is out of the way and not likely to be used in a variable name by accident.

---
In reply to: [427502200](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427502200) [](ancestors = 427502200,426962229)

azure-iot-device/azure/iot/device/iothub/edge_hsm.py
```diff
@@ -1,80 +1,28 @@
...
-        :return: The signature portion of the Sas Token.
-
-        :raises: IoTEdgeError if data cannot be signed
-        """
-        string_to_sign = quoted_resource_uri + "\n" + str(expiry)
-        return self.hsm.sign(string_to_sign)
-
-
-class IoTEdgeHsm(object):
+class IoTEdgeHsm(SigningMechanism):
```
BertKleewein, line 25: don't know the solution, but it seems odd that SigningMechanism is in the common.auth sub-package, but this one is not in an auth sub-package

azure-iot-device/azure/iot/device/common/callable_weak_method.py
cartertinney: I agree, especially since the getattr is the only thing required to fix the issue.

---
In reply to: [427012869](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427012869) [](ancestors = 427012869)

azure-iot-device/azure/iot/device/iothub/edge_hsm.py
cartertinney: Yeah, I'm not thrilled about it, but I didn't really want to make a package just for one module

---
In reply to: [427506098](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427506098) [](ancestors = 427506098)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I agree, that's why I didn't upend the implementation to account for this. Figured it would be a good information to document inline however.

---
In reply to: [427411587](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427411587) [](ancestors = 427411587)

azure-iot-device/azure/iot/device/common/pipeline/config.py
cartertinney: I'll make the change now

---
In reply to: [426949443](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426949443) [](ancestors = 426949443)

cartertinney: At some point we'll change the functionality to take both the contents or filepath

---
In reply to: [426949443](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426949443) [](ancestors = 426949443)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: This happens on line 303

---
In reply to: [426955668](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r426955668) [](ancestors = 426955668)

azure-iot-device/tests/common/pipeline/test_pipeline_ops_http.py
```diff
@@ -151,7 +78,7 @@ def test_reason(self, cls_type, init_kwargs):
 
 pipeline_ops_test.add_operation_tests(
     test_module=this_module,
-    op_class_under_test=pipeline_ops_http.SetHTTPConnectionArgsOperation,
+    op_class_under_test=pipeline_ops_http.HTTPRequestAndResponseOperation,
```
BertKleewein, line 81: >op_class_under_test=pipeline_ops_http.HTTPRequestAndResponseOperation, [](start = 4, length = 70)

huh.  apparently this attribute isn't that important

cartertinney: I mean.... it is important... it's just that it doesn't cause failure. If you pass the wrong one, you test the wrong one. 

---
In reply to: [427577323](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427577323) [](ancestors = 427577323)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
olivakar: to be clear are you saying this would be user input kwarg ?

---
In reply to: [427021277](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427021277) [](ancestors = 427021277)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
BertKleewein: I'm OK with that,  I've seen that problem before and never came up with a better answer.  I wouldn't be offended if you left this as-is

---
In reply to: [427504005](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427504005) [](ancestors = 427504005,427502200,426962229)

azure-iot-device/azure/iot/device/common/auth/sastoken.py
BertKleewein: I'm wondering if this was the name in the azure portal at one point.  Right now, the key names (iothubowner, etc) show up in the portal under "Shared Access Policy" as "Access policy name".  We had this same thing happen with one of the DPS names -- I think ProvisioningHostName shows up in the portal with one name and in the SDK as another.  

---
In reply to: [427500301](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r427500301) [](ancestors = 427500301,426946266)

azure-iot-device/tests/common/pipeline/config_test.py
```diff
@@ -0,0 +1,230 @@
...
+# license information.
+# --------------------------------------------------------------------------
+import pytest
+import abc
+import six
+from azure.iot.device import ProxyOptions
+
+
+@six.add_metaclass(abc.ABCMeta)
+class PipelineConfigInstantiationTestBase(object):
```
BertKleewein, line 13: >PipelineConfigInstantiationTestBase [](start = 6, length = 35)

<eyes crossing at 7 AM> Man, this "InstantiationTestBase" convention makes for some long identifiers.  No change requested.  Just murmuring to myself.  Get off my lawn.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: yes, margin should be a config kwarg that the user can supply

---
In reply to: [428953125](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r428953125) [](ancestors = 428953125,427021277)

azure-iot-device/tests/common/pipeline/config_test.py
cartertinney: Some of these test class names could be confused for parody of Java tbh :joy: 

---
In reply to: [429295058](https://github.com/Azure/azure-iot-sdk-python/pull/555#discussion_r429295058) [](ancestors = 429295058)

PR #560, opened 2020-05-08
chore(pnp):read from json file

olivakar:
Read component names from model files


azure-iot-device/samples/pnp/pnp_sample_device.py
```diff
@@ -27,10 +27,26 @@
 # and the name user wants to call the pnp device
 model_id = "dtmi:my_company:com:SampleDevice;1"
```
jebrando, line 28: Technically you can read the model_id from the file also.

olivakar: so i already thought about this but right now i need some kind of knowledge to decide that i am reading a sensor model/component name vs a device information model/component vs a sdk information model/component....on reading both from the file i would have no control over which component is only for telemetry and which one is for twin etc....one thing that can be done eventually is to decide telemetry and properties and commands from the file as well...

PR #574, opened 2020-05-26
chore: script for certificate creation using cryptography as library

olivakar:
Checked for both python 3 and python 2. 
DPS E2E now uses this script.

TODO : Update Wiki for usage with this. Then delete the other file


BertKleewein:
I'm confused about the purpose of this tool.  Is this for our testing pipeline or is this for end-user consumption?  It looks very pipeline-specific, but the comment about updating the wiki makes me think that you designed it to be used from outside the pipeline.

scripts/create_x509_chain_crypto.py
```diff
@@ -0,0 +1,558 @@
...
+    :param common_name: The common name to be used in the subject. This is a single common name which would be applied to all certs created. Since this common name is meant for all,
+    this common name will be prepended by the words "root", "inter" and "device" for root, intermediate and device certificates.
+    For device certificates the common name will be further appended with the index of the device.
+    :param ca_password: The password for the root certificate which is going to be referenced by the intermediate.
+    :param intermediate_password: The password for the intermediate certificate
+    :param device_password: The password for the device certificate
+    :param device_count: The number of leaf devices for which that many number of certificates will be generated.
+    :param key_size: The key size to use for encryption. The default is 4096.
+    :param days: The number of days for which the certificate is valid. The default is 1 year or 365 days.
+    For the root cert this value is multiplied by 10. For the device certificates this number will be divided by 10.
```
BertKleewein, line 69: >For the root cert this value is multiplied by 10. For the device certificates this number will be divided by 10. [](start = 4, length = 112)

why?  i know this is for testing only, but I'm not sure I follow the logic.

scripts/create_x509_chain_crypto.py
```diff
@@ -0,0 +1,558 @@
...
+
+    :param common_name: The common name to be used in the subject. This is a single common name which would be applied to all certs created. Since this common name is meant for all,
+    this common name will be prepended by the words "root", "inter" and "device" for root, intermediate and device certificates.
+    For device certificates the common name will be further appended with the index of the device.
+    :param ca_password: The password for the root certificate which is going to be referenced by the intermediate.
+    :param intermediate_password: The password for the intermediate certificate
+    :param device_password: The password for the device certificate
+    :param device_count: The number of leaf devices for which that many number of certificates will be generated.
+    :param key_size: The key size to use for encryption. The default is 4096.
+    :param days: The number of days for which the certificate is valid. The default is 1 year or 365 days.
```
BertKleewein, line 68: >The default is 1 year or 365 day [](start = 72, length = 32)

doesn't match code

scripts/create_x509_chain_crypto.py
```diff
@@ -0,0 +1,558 @@
...
+            csr_file=device_csr_file,
+            subject=device_common_name + str(i),
+            is_ca=False,
+        )
+
+        builder = create_cert_builder(
+            subject=device_csr.subject,
+            issuer_name=inter_cert_subject,
+            public_key=device_csr.public_key(),
+            days=int(days / 100),
```
BertKleewein, line 184: >days=int(days / 100) [](start = 12, length = 20)

this on is divided by 100.  this all seems very arbitrary and random.

scripts/create_x509_chain_crypto.py
```diff
@@ -0,0 +1,558 @@
...
+    )
+
+    verification_cert = verification_builder.sign(
+        private_key=issuer_private_key, algorithm=hashes.SHA256(), backend=default_backend()
+    )
+    with open(verfication_cert_file, "wb") as f:
+        f.write(verification_cert.public_bytes(serialization.Encoding.PEM))
+
+
+def create_cert_builder(subject, issuer_name, public_key, days=30, is_ca=False):
```
BertKleewein, line 244: This one is 30.  Hmmm.  I'm only paying attention to the days because the inconsistency stands out, but also because cert expiration is an important part of security.  If this is only for tests, why do we have any certs that last longer than one day?  If this is for other uses, then maybe we need to pay extra attention because certs that last 10 years seem dangerously long-lived.

scripts/create_x509_chain_crypto.py
olivakar: now removed and everything is default to 1 year...i think i took a suggestion of creating root for 10 years..not a good idea.....

---
In reply to: [432499532](https://github.com/Azure/azure-iot-sdk-python/pull/574#discussion_r432499532) [](ancestors = 432499532)

scripts/create_x509_chain_crypto.py
olivakar: not actually random...also this exactly does what the last script used to do.....so this is how i go with it..roots are created for 1 year, intermediate for almost a  month, device for almost some days...

---
In reply to: [432503675](https://github.com/Azure/azure-iot-sdk-python/pull/574#discussion_r432503675) [](ancestors = 432503675)

scripts/create_x509_chain_crypto.py
olivakar: changed to 365...the default of everything is 365 or 1 year....there is no 10 year one now...also intermediates get crated for almost a month and device ones get created for some days

---
In reply to: [432557761](https://github.com/Azure/azure-iot-sdk-python/pull/574#discussion_r432557761) [](ancestors = 432557761)

olivakar:
i  should give a little bit of history here......there were 2 scripts before....absolutely similar in logic...one for the wiki/samples  and one for the e2e tests....mostly because the tests one were not working for all versions of python....this one can be used for creating local certificates and can be used for DPS e2e tests....the DPS e2e certs do not use the root cert from here as that is created from build environment script....and any certificates created also get deleted after tests....sso there are no long running scripts...now for the wiki if you feel the durations needs to be reduced...then i can reduce all to 30...right now even if used from wiki only the root will be 1 year, intermediate will be month and device will be days

---
In reply to: [635983269](https://github.com/Azure/azure-iot-sdk-python/pull/574#issuecomment-635983269) [](ancestors = 635983269)

PR #576, opened 2020-05-27
fix (azure-iot-device): use paho disconnect/reconnect to refresh sas token

BertKleewein:
# Thank you for helping us improve the Azure IoT Python SDK!

## Need Support

- **Have a feature request for SDKs?** Please post it on [User Voice](https://feedback.azure.com/forums/321918-azure-iot) to help us prioritize
- **Have a technical question?** Ask on [Stack Overflow with tag "azure-iot-hub"](https://stackoverflow.com/questions/tagged/azure-iot-hub)
- **Need Support?** Every customer with an active Azure subscription has access to [support](https://docs.microsoft.com/en-us/azure/azure-supportability/how-to-create-azure-support-request) with guaranteed response time.  Consider submitting a ticket and get assistance from Microsoft support team
- **Found a bug?** Please help us fix it by thoroughly documenting it and filing an issue on GitHub (See below).

Here's a little checklist of things that will help it make its way to the repository: Note that you don't have to check all the boxes, we can help you with that.
This being said, the more you do, the quicker it'll go through our gated build!
-->

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-python/.github/CONTRIBUTING.md).

## Description of the problem

Please be as precise as possible: what issue you experienced, how often...

How you solved the issue and the other things you considered and maybe rejected


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -217,7 +217,9 @@ def on_disconnect(client, userdata, rc):
             if not this:
                 # Paho will sometimes call this after we've been garbage collected,  If so, we have to
                 # stop the loop to make sure the Paho thread shuts down.
-                logger.info("disconnected after garbage collection. stopping loop")
+                logger.info(
+                    "on_disconnect called with this=None. Transport must have been garbage collected. stopping loop"
```
olivakar, line 221: >this=None [](start = 47, length = 9)

Do you want to say this specifically ? or say `on_disconnect called with transport being None..`

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -897,28 +897,33 @@ class ReconnectState(object):
...
 class ReconnectStage(PipelineStage):
     def __init__(self):
         super(ReconnectStage, self).__init__()
         self.reconnect_timer = None
-        self.state = ReconnectState.NEVER_CONNECTED
+        self.state = ReconnectState.LOGICALLY_DISCONNECTED
+        self.never_connected = True
         # connect delay is hardcoded for now.  Later, this comes from a retry policy
-        self.reconnect_delay = 10
+        self.delayed_reconnect_delay = 10
```
olivakar, line 925: >delayed_reconnect_delay [](start = 13, length = 23)

could you put some help docs over the two types of delay ? when do we choose one over the other ?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -992,60 +1015,52 @@ def _send_new_connect_op_down(self):
         def on_connect_complete(op, error):
             this = self_weakref()
             if this:
+                logger.debug(
+                    "{}({}): on_connect_complete error={} state={} never_conencted={} connected={} ".format(
```
olivakar, line 1019: >never_conencted [](start = 67, length = 15)

spell

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -966,18 +980,27 @@ def _run_op(self, op):
...
+                # For now, we use immediate_reconnect_delay to mean "close enough to immediate"
+                # because it saves us from refactoring this code when we know that another change
+                # is coming.  Specifically, we know that this stage is broken in cases where the
+                # user calls disconnect().  In that case, this stage doesn't currently know that
+                # it shouldn't reconnect automatically.
                 self.state = ReconnectState.WAITING_TO_RECONNECT
-                self._start_reconnect_timer()
+                self._start_reconnect_timer(self.immediate_reconnect_delay)
+
             else:
```
olivakar, line 1001: >else [](start = 12, length = 4)

for an user disconnect the code path will turn up here right ? where the reconnect timer will not fire ?

olivakar:
> class ReconnectStageInstantiationTests(ReconnectStageTestConfig):

initialize test for the never connected attribute to be set to true ?

---
Refers to: azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py:2666 in 123c045. [](commit_id = 123c0456376be586c73f63daa54604f3a1a3ad0a, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -897,28 +897,33 @@ class ReconnectState(object):
     """
     Class which holds reconenct states as class variables.  Created to make code that reads like an enum without using an enum.
 
-    NEVER_CONNECTED: Ttransport has never been conencted.  This state is necessary because some errors might be fatal or transient,
-    depending on wether the transport has been connceted.  For example, a failed conenction is a transient error if we've connected
-    before, but it's fatal if we've never conencted.
+    WAITING_TO_RECONNECT: This stage is in a waiting period before reconnecting.  This state implies
+    that the user wants the pipeline to be connected.  ie. After a successful connection, the
+    state will change to LOGICALLY_CONNECED
```
olivakar, line 902: >LOGICALLY_CONNECED [](start = 25, length = 18)

spell

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -966,18 +980,27 @@ def _run_op(self, op):
...
-                        self.name, event.name, self.state
-                    )
+            logger.debug(
+                "{}({}): State is {} Connected is {}.".format(
+                    self.name, event.name, self.state, self.pipeline_root.connected
                 )
+            )
+
+            if self.pipeline_root.connected and self.state == ReconnectState.LOGICALLY_CONNECTED:
+                # when we get disconnected, we try to immediatly reconnect.  If that fails,
```
olivakar, line 990: >immediatly [](start = 54, length = 10)

spell

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -992,60 +1015,52 @@ def _send_new_connect_op_down(self):
...
+                    )
+                )
                 if error:
-                    if this.state == ReconnectState.NEVER_CONNECTED:
-                        logger.info(
-                            "{}({}): error on first connection.  Not triggering reconnection".format(
-                                this.name, op.name
-                            )
-                        )
+                    if this.never_connected:
```
olivakar, line 1029: >if this.never_connected: [](start = 20, length = 24)

this can also be put in the else bock right ? this does the same as the else block ... a comment can be added that the else block is executed on if `never_connected =True ` as well as for other non-transient errors

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -3170,67 +3239,55 @@ def stage(self, mocker, cls_type, init_kwargs, state, reconnect_timer, mock_time
...
-
-    @pytest.mark.it("If previously connected, clears the previous reconnect timer if there was one")
-    def test_clears_reconnect_timer(self, stage, event):
-        old_timer = stage.reconnect_timer
-        stage.pipeline_root.connected = True
+    @pytest.mark.parametrize("state", [pipeline_stages_base.ReconnectState.LOGICALLY_CONNECTED])
+    @pytest.mark.it(
+        "If and only if connected and logically connected, changes the state to WAITING_TO_RECONNECT"
+    )
+    def test_changes_state(self, stage, event, state, currently_connected):
```
olivakar, line 3246: >test_changes_state [](start = 8, length = 18)

can u separate these tests for the `currently_connected` part ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -3170,67 +3239,55 @@ def stage(self, mocker, cls_type, init_kwargs, state, reconnect_timer, mock_time
...
-    def test_sets_new_reconnect_timer(self, stage, event, mock_timer):
-        stage.pipeline_root.connected = True
-        stage.handle_pipeline_event(event)
-        assert mock_timer.call_count == 1
-        assert stage.reconnect_timer == mock_timer.return_value
-        assert stage.reconnect_timer.start.call_count == 1
+        if currently_connected and state == pipeline_stages_base.ReconnectState.LOGICALLY_CONNECTED:
+            if old_timer:
+                assert old_timer.cancel.call_count == 1
+            assert stage.reconnect_timer != old_timer
```
olivakar, line 3265: >!=  [](start = 41, length = 3)

are we trying to assert that these 2 objects are different ? then we should probably use `is not` for unequal identities.

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -3513,47 +3605,37 @@ def test_completes_waiting_connect_ops(
...
             assert op.original_callback.call_args == mocker.call(op=op, error=None)
 
-    @pytest.mark.parametrize(
-        "state",
-        [
-            pipeline_stages_base.ReconnectState.NEVER_CONNECTED,
-            pipeline_stages_base.ReconnectState.CONNECTED_OR_DISCONNECTED,
-        ],
+    @pytest.mark.it(
+        "Changes the state to LOGICALLY_DISCONNECTED if the connection fails with an arbitrary error"
```
olivakar, line 3609: >arbitrary [](start = 85, length = 9)

all these arbitrary errors are non transient errors right ? for me if the statement mentions non transient error it makes more sense..

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -3586,9 +3677,8 @@ def test_completes_waiting_connect_ops_on_arbitrary_exception(
             assert op.original_callback.call_count == 1
             assert op.original_callback.call_args == mocker.call(op=op, error=arbitrary_exception)
 
-    @pytest.mark.parametrize("state", [pipeline_stages_base.ReconnectState.NEVER_CONNECTED])
     @pytest.mark.it(
-        "Completes all waiting ops with the transient failure if the connection fails with a transient error"
+        "Completes all waiting ops with the transient failure if a first-time connection fails with a transient error"
```
olivakar, line 3681: >transient error [](start = 102, length = 15)

for a first connection case , does this matter if it is transient error or it includes transient and non transient error of all kinds ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
```diff
@@ -3610,57 +3701,37 @@ def test_completes_all_waiting_connect_ops_on_transient_connect_exception(
                 op=op, error=transient_connect_exception
             )
 
-    @pytest.mark.parametrize("state", [pipeline_stages_base.ReconnectState.NEVER_CONNECTED])
     @pytest.mark.it(
-        "Does not create a reconnect timer if the connection fails with a transient error"
+        "Does not create a reconnect timer if the connection fails with a transient error on a first-time connection"
```
olivakar, line 3705: >transient error on a first-time connection [](start = 74, length = 42)

same comment as the first time connection case

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -392,76 +392,29 @@ def test_pending_operation_cancelled(self, mocker, stage, op, pending_connection
...
-        assert mock_timer.call_count == 1
-        assert mock_timer.call_args == mocker.call(WATCHDOG_INTERVAL, mocker.ANY)
-        assert mock_timer.return_value.daemon is True
-        assert mock_timer.return_value.start.call_count == 1
-
-    @pytest.mark.it(
-        "Performs an MQTT reconnect via the MQTTTransport, using the pipeline root's SasToken as a password"
-    )
+    @pytest.mark.it("Performs an MQTT disconnect via the MQTTTransport")
     def test_mqtt_connect(self, mocker, stage, op):
```
olivakar, line 396: >test_mqtt_connect [](start = 8, length = 17)

change test name

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -763,7 +716,9 @@ def test_completes_pending_connect_op(self, mocker, stage):
         assert op.error is None
         assert stage._pending_connection_op is None
 
-    @pytest.mark.it("Completes a pending ReauthorizeConnectionOperation successfully")
+    @pytest.mark.it(
+        "Ignores a pending ReathorizeConnectionOperation when the transport connected event fires"
+    )
     def test_completes_pending_reconnect_op(self, mocker, stage):
```
olivakar, line 722: >test_completes_pending_reconnect_op [](start = 8, length = 35)

`pending_reauthorize_op` ?

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -816,23 +770,22 @@ def test_cancels_watchdog_on_pending_connect(self, mocker, stage, mock_timer):
         assert mock_timer.return_value.cancel.call_count == 1
 
     @pytest.mark.it(
-        "Cancels the connection watchdog if the pending operation is a ReauthorizeConnectionOperation"
+        "Does not cancel the connection watchdog if the pending operation is a ReauthorizeConnectionOperation"
     )
-    def test_cancels_watchdog_on_pending_reauthorize(self, mocker, stage, mock_timer):
+    def test_does_not_cancelcel_watchdog_on_pending_reauthorize(self, mocker, stage, mock_timer):
```
olivakar, line 775: >test_does_not_cancelcel_watchdog_on_pending_reauthorize [](start = 8, length = 55)

spell cancel

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -763,7 +716,9 @@ def test_completes_pending_connect_op(self, mocker, stage):
         assert op.error is None
         assert stage._pending_connection_op is None
 
-    @pytest.mark.it("Completes a pending ReauthorizeConnectionOperation successfully")
+    @pytest.mark.it(
+        "Ignores a pending ReathorizeConnectionOperation when the transport connected event fires"
```
olivakar, line 720: >Ignores [](start = 9, length = 7)

does not complete sounds more clear..

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -1121,24 +1057,12 @@ def test_comletes_with_cause_as_error_if_cause(
...
-            pytest.param(
-                pipeline_ops_base.ConnectOperation(callback=fake_callback),
-                id="Pending ConnectOperation",
-            ),
-            pytest.param(
-                pipeline_ops_base.ReauthorizeConnectionOperation(callback=fake_callback),
-                id="Pending ReauthorizeConnectionOperation",
-            ),
-        ],
+        "Completes (unsuccessfully) a ConnectOoperation with a ConnectionDroppedError if no cause is provided for the disconnection"
```
olivakar, line 1060: >ConnectOoperation [](start = 38, length = 17)

spell

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -1121,24 +1057,12 @@ def test_comletes_with_cause_as_error_if_cause(
...
     )
     def test_comletes_with_connection_dropped_error_as_error_if_no_cause(
```
olivakar, line 1062: >comletes [](start = 13, length = 8)

spell

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
```diff
@@ -1209,23 +1133,24 @@ def test_cancels_watchdog_on_pending_connect(self, mocker, stage, mock_timer, ca
         assert mock_timer.return_value.cancel.call_count == 1
 
     @pytest.mark.it(
-        "Cancels the connection watchdog if the pending operation is a ReauthorizeConnectionOperation"
+        "Does not cancel the connection watchdog if the pending operation is a ReauthorizeConnectionOperation"
     )
-    def test_cancels_watchdog_on_pending_reauthorize(self, mocker, stage, mock_timer, cause):
+    def test_does_not_cancel_watchdog_on_pending_reauthorize(
```
olivakar, line 1138: >test_does_not_cancel_watchdog_on_pending_reauthorize [](start = 8, length = 52)

i understand what you want to test , but if you just look at the code of `_on_mqtt_disconnected` , regardless of what type of op it is we are always calling  `_cancel_connection_watchdog` , so when the test becomes does not cancel , it is not matching if we just read the code.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -897,28 +897,33 @@ class ReconnectState(object):
...
+    LOGICALLY_DISCONNECTED: The client does not want the pipeline to be connected or the pipeline had
+    a fatal error and was forced to disconnect.  If the state is LOGICALLY_DISCONNECTED, then the pipeline
+    should be physically disconnected since there is no reason to leave the pipeline connected in this state.
     """
 
-    NEVER_CONNECTED = "NEVER_CONNECTED"
     WAITING_TO_RECONNECT = "WAITING_TO_RECONNECT"
-    CONNECTED_OR_DISCONNECTED = "CONNECTED_OR_DISCONNECTED"
+    LOGICALLY_CONNECTED = "LOGICALLY_CONNECTED"
+    LOGICALLY_DISCONNECTED = "LOGICALLY_DISCONNECTED"
```
cartertinney, line 915: Big improvement here on the names

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: +1

---
In reply to: [437135351](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r437135351) [](ancestors = 437135351)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -937,6 +942,14 @@ def _run_op(self, op):
                         self.name, op.name, self.state
                     )
                 )
+                self.state = ReconnectState.LOGICALLY_CONNECTED
+                # We don't send this op down.  Instead, we send a new connect op down.  This way,
+                # we can distinguish between connect ops that we're handling (they go into the
+                # queue) and connect ops that we are sending down.
```
cartertinney, line 948: I wonder if this indicates perhaps there should be a concrete separate op so this isn't just by convention. Something to consider going forth

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -966,18 +980,27 @@ def _run_op(self, op):
...
+                # then we wait a while before retrying.
+                #
+                # For now, we use immediate_reconnect_delay to mean "close enough to immediate"
+                # because it saves us from refactoring this code when we know that another change
+                # is coming.  Specifically, we know that this stage is broken in cases where the
+                # user calls disconnect().  In that case, this stage doesn't currently know that
+                # it shouldn't reconnect automatically.
                 self.state = ReconnectState.WAITING_TO_RECONNECT
-                self._start_reconnect_timer()
+                self._start_reconnect_timer(self.immediate_reconnect_delay)
```
cartertinney, line 999: >self.immediate_reconnect_delay [](start = 44, length = 30)

Why not just directly initiate the action rather than starting a timer for 0.01 seconds that then triggers the action?

I assume it's to do with code readability and cleanliness?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -947,17 +960,18 @@ def _run_op(self, op):
                         self.name, op.name, self.state
                     )
                 )
+                self.state = ReconnectState.LOGICALLY_DISCONNECTED
                 self._clear_reconnect_timer()
                 self._complete_waiting_connect_ops(
                     pipeline_exceptions.OperationCancelled("Explicit disconnect invoked")
```
cartertinney, line 966: >pipeline_exceptions.OperationCancelled("Explicit disconnect invoked") [](start = 20, length = 69)

Unrelated to the rest of this PR I guess, but I don't get what's going on here, and why this gets set as a waiting connect op

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -992,60 +1015,52 @@ def _send_new_connect_op_down(self):
...
                     else:
-                        logger.info(
-                            "{}({}): State is {}.  Connection failed. Not triggering reconnection".format(
-                                this.name, op.name, this.state
-                            )
-                        )
+                        # all others are fatal
+                        this.state = ReconnectState.LOGICALLY_DISCONNECTED
+                        this._clear_reconnect_timer()
                         this._complete_waiting_connect_ops(error)
```
cartertinney, line 1042: >this._complete_waiting_connect_ops(error) [](start = 24, length = 41)

This seems like it would include, as mentioned in a comment above, OperationCancelled operations, which represent a disconnect somehow.

So, reaching a disconnect state by connection failure causes the disconnect operation to fail?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -937,6 +942,14 @@ def _run_op(self, op):
                         self.name, op.name, self.state
                     )
                 )
+                self.state = ReconnectState.LOGICALLY_CONNECTED
+                # We don't send this op down.  Instead, we send a new connect op down.  This way,
```
cartertinney, line 946: What if we already were Logically Connected? Should we perhaps have an additional conditional block to prevent us from trying to connect when we already are (logically) connected?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -947,17 +960,18 @@ def _run_op(self, op):
...
                 self._complete_waiting_connect_ops(
                     pipeline_exceptions.OperationCancelled("Explicit disconnect invoked")
                 )
-                self.state = ReconnectState.CONNECTED_OR_DISCONNECTED
                 op.complete()
 
             else:
                 logger.info(
                     "{}({}): State is {}.  Sending op down.".format(self.name, op.name, self.state)
                 )
```
cartertinney, line 973: Here (and in the above blocks as well) we log the old state right before setting a new state. That's probably going to really confuse up the log statements, I assume it would be better for us to log AFTER adjusting the state

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -897,28 +897,33 @@ class ReconnectState(object):
...
 
 
 class ReconnectStage(PipelineStage):
     def __init__(self):
         super(ReconnectStage, self).__init__()
         self.reconnect_timer = None
-        self.state = ReconnectState.NEVER_CONNECTED
+        self.state = ReconnectState.LOGICALLY_DISCONNECTED
+        self.never_connected = True
         # connect delay is hardcoded for now.  Later, this comes from a retry policy
```
cartertinney, line 924: I also would like some doc surrounding this value, and it's significance. Clearly it is quite important to the logic, but just from the code here I can't tell why we have to care if something was never connected.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -992,60 +1015,52 @@ def _send_new_connect_op_down(self):
...
+                )
                 if error:
-                    if this.state == ReconnectState.NEVER_CONNECTED:
-                        logger.info(
-                            "{}({}): error on first connection.  Not triggering reconnection".format(
-                                this.name, op.name
-                            )
-                        )
+                    if this.never_connected:
+                        # any error on a first connection is fatal
```
cartertinney, line 1030: >fatal [](start = 61, length = 5)

Is it? It doesn't look fatal

BertKleewein: "fatal" is maybe a bad error.  I changed it to "transient error" or "permanent error".  I _think_ this is better.

---
In reply to: [439846621](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439846621) [](ancestors = 439846621)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: In short, 2 reasons: 
1. When this is called, we are still "disconnecting" and not "completely disconnected", so it's better to let this event flow up before we try again.
2. When I tried to call it immediately, the volume of test cases exploded.  I'm normally against changing code to make testing easier, but this is a situation where the tests showed me that  calling connect here had more implications than you would think.

---
In reply to: [439685542](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439685542) [](ancestors = 439685542)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: I went back to one delay variable and hardcoded the other.  There's no reason to ever adjust the immediate delay and it's only used in a single case, so it seems less confusing if we only have a single delay variable.


---
In reply to: [439658750](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439658750) [](ancestors = 439658750,437135351)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: corrrect

---
In reply to: [437156650](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r437156650) [](ancestors = 437156650)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: I like this train of thought, but it's taking me in a different direction.  For now, ConnectionLockStage is above this one, so we only get a ConnectOperation here if we're physically disconnected.  
In the longer term, your comment somewhere else in here makes me wonder if we want to have different types of connect ops:
PhysicalConnectOp - for calling into transport
LogicalConnectOp - for setting LogicallyConnected
GuaranteedPhysicalConnectOp - physical connect op with retry

or something like that.  Myabe PhysicalConnectOp is guaranteed and NonGuaranteedConnectOp is the one that isn't guaranteed.  I'm not sure.  

Either way, that's rambing and not well thouht out.  The code in here is good for now.

---
In reply to: [439846074](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439846074) [](ancestors = 439846074)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: That's correct.  The tests for permanent (arbitrary, non-transient) errors are above.

---
In reply to: [437760479](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r437760479) [](ancestors = 437760479)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_base.py
BertKleewein: same as above.

---
In reply to: [437760653](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r437760653) [](ancestors = 437760653)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Not sure if you selected the wrong line here.  This code (DisconnectOperation, WAITING_TO_RECONNECT) covers the case where the connection was dropped and it's waiting to retry when the user calls disconnect().  In that case, any operations that were waiting for the reconnect to succeed need to be cancelled.  By calling disconnect(), the user was effectively cancelling all operations that were pending. 

---
In reply to: [439686461](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439686461) [](ancestors = 439686461)

BertKleewein: I'm not sure because the line you selected is the one that completes (cancels) all of the waiting connection ops.

---
In reply to: [440422365](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r440422365) [](ancestors = 440422365,439686461)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Related to my previous response.  The DisconnectOperation is never set as a waiting_connectoin_op.  The disconnect causes waiting ops to be cancelled, but it's never a waiting op itself.

---
In reply to: [439687473](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r439687473) [](ancestors = 439687473)

azure-iot-device/tests/common/pipeline/test_pipeline_stages_mqtt.py
BertKleewein: For disconnect and reauthorize, there is no connection watchdog, so there's nothing to cancel.  The call to `_cancle_connection_watchdog` just doesn't do anything in this case.  I changed the test name to be a little more clear and, unfortunately, also a little big more awkward.

---
In reply to: [437786906](https://github.com/Azure/azure-iot-sdk-python/pull/576#discussion_r437786906) [](ancestors = 437786906)

PR #578, opened 2020-05-28
fix(azure-iot-device): DPS samples allow sending telemetry

olivakar:
Change all DPS samples to send telemetry


azure-iot-device/samples/async-hub-scenarios/provision_symmetric_key.py
```diff
@@ -27,6 +32,34 @@ async def main():
...
+        await device_client.connect()
+
+        async def send_test_message(i):
+            print("sending message #" + str(i))
+            msg = Message("test wind speed " + str(i))
+            msg.message_id = uuid.uuid4()
+            msg.correlation_id = "correlation-1234"
+            msg.custom_properties["count"] = i
+            msg.custom_properties["tornado-warning"] = "yes"
+            await device_client.send_message(msg)
```
jebrando, line 52: You don't need to set properties here (and other places), we want to make it as concise as possible

PR #579, opened 2020-05-29
feature (azure-iot-device): Add 'sastoken_ttl' kwarg option to factory methods

cartertinney:
* Added support for custom SasToken TTL via the 'sastoken_ttl' kwarg on SAS-based factory methods of the IoTHub clients
* Added support for custom SasToken TTL via the 'sastoken_ttl' kwarg on SAS-based facotry methods of the Provisioning clients

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/abstract_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/abstract_clients.py
@@ -21,7 +21,7 @@ from . import edge_hsm
 logger = logging.getLogger(__name__)
 
 
-def _validate_kwargs(**kwargs):
+def _validate_kwargs(exclude=[], **kwargs):
     """Helper function to validate user provided kwargs.
     Raises TypeError if an invalid option has been provided"""
     valid_kwargs = [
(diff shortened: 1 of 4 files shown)


azure-iot-device/tests/iothub/shared_client_tests.py
```diff
@@ -530,6 +595,13 @@ def test_client_returned(
         assert client._mqtt_pipeline is mock_mqtt_pipeline_init.return_value
         assert client._http_pipeline is mock_http_pipeline_init.return_value
 
+    @pytest.mark.it("Raises a TypeError if the 'sastoken_ttl' kwarg is supplied by the user")
+    def test_sastoken_ttl(self, client_class, x509):
+        with pytest.raises(TypeError):
+            client_class.create_from_x509_certificate(
+                x509=x509, hostname=self.hostname, device_id=self.device_id, sastoken_ttl=1000
+            )
+
```
cartertinney, line 604: There is a better pattern for handling excluded kwargs from a subset of factory methods, but I'll wait to implement it until we have more than one instance of this problem

azure-iot-device/tests/iothub/shared_client_tests.py
```diff
@@ -410,6 +445,34 @@ def test_sastoken(self, mocker, client_class):
...
+    @pytest.mark.it(
+        "Uses 3600 seconds (1 hour) as the default SasToken TTL if no custom TTL is provided"
+    )
+    def test_sastoken_default(self, mocker, client_class):
+        sksm_mock = mocker.patch.object(auth, "SymmetricKeySigningMechanism")
+        sastoken_mock = mocker.patch.object(st, "SasToken")
+        expected_uri = "{hostname}/devices/{device_id}".format(
+            hostname=self.hostname, device_id=self.device_id
+        )
```
cartertinney, line 474: We test SasToken config here, as opposed in the shared user option tests, because this isn't a kwarg that goes to the PipelineConfig. I'll probably rename `SharedIoTHubClientCreateMethodUserOptionTests` down the road to reflect that

cartertinney:
./azp run Azure.azure-iot-sdk-python-dps-e2e

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -235,8 +259,10 @@ def create_from_symmetric_key(cls, symmetric_key, hostname, device_id, **kwargs)
             arbitrary product info which is appended to the user agent string.
         :param proxy_options: Options for sending traffic through proxy servers.
         :type proxy_options: :class:`azure.iot.device.ProxyOptions`
+        :param int sastoken_ttl: The time to live (in seconds) for the created SasToken used for
+            authentication. Default is 3600 seconds (1 hour)
```
BertKleewein, line 263: random comment: I like to see the level of detail in our docstrings go up.  If you haven't looked at the public-facing docstrings in the other Azure SDKs, you should (I think I'm thinking of eventhub, but maybe uamqp).

cartertinney: Yeah I agree, we could do a bit more. Once we get docgen occurring correctly, there's a number of improvements I think could be made.

PR #592, opened 2020-06-15
chore: remove unused transitions library

BertKleewein:


azure-iot-device/setup.py
```diff
@@ -84,7 +84,6 @@
         # Actual project dependencies
         "six>=1.12.0,<2.0.0",
         "paho-mqtt>=1.4.0,<2.0.0",
-        "transitions>=0.6.8,<1.0.0",
```
jebrando, line 87: If we are no longer dependent on this can you please remove it from the ThirdPartyNotice.txt

cartertinney: +1

---
In reply to: [440493584](https://github.com/Azure/azure-iot-sdk-python/pull/592#discussion_r440493584) [](ancestors = 440493584)

PR #600, opened 2020-06-18
fix: always retry on name resolution error

BertKleewein:


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -1028,6 +1029,17 @@ def _handle_pipeline_event(self, event):
         else:
             self.send_event_up(event)
 
+    def _always_try_reconnecting(self, error):
+        try:
+            return (
+                type(error) == transport_exceptions.ConnectionFailedError
+                and error.__cause__
+                and type(error.__cause__) == socket.gaierror
```
cartertinney, line 1037: >type(error.__cause__) == socket.gaierror [](start = 20, length = 40)

Would isinstance() be more accurate here?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -1028,6 +1029,17 @@ def _handle_pipeline_event(self, event):
         else:
             self.send_event_up(event)
 
+    def _always_try_reconnecting(self, error):
```
cartertinney, line 1032: >_always_try_reconnecting(self, error): [](start = 8, length = 38)

I'd like a little text documentation explaining this logic because it's not clear on it's face why this is necessary

PR #615, opened 2020-07-06
fix(pnp) : Added pure helper

olivakar:
So that there is very little difference between the non-PNP and pnp device.

### DIFF
--- a/azure-iot-device/samples/pnp/pnp_temp_controller_with_thermostats.py
+++ b/azure-iot-device/samples/pnp/pnp_temp_controller_with_thermostats.py
@@ -10,9 +10,10 @@ import logging
 import json
 
 from azure.iot.device.aio import IoTHubDeviceClient
-import pnp_methods
-from azure.iot.device import constant
+from azure.iot.device import Message, MethodResponse
 from datetime import date
+
+import pnp_helper_summer_refresh
 
 
 logging.basicConfig(level=logging.ERROR)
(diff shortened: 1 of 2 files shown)


azure-iot-device/samples/pnp/pnp_temp_controller_with_thermostats.py
```diff
@@ -135,66 +213,59 @@ async def main():
...
 
-    asyncio.create_task(
-        pnp_methods.pnp_update_property(
-            device_client,
-            thermostat_1_component_name,
-            targetTemperature={"value": 56.78, "ac": 200, "ad": "wingardium leviosa", "av": 1},
-            maxTempSinceLastReboot=67.89,
-        )
+    pnp_properties_root = pnp_helper_summer_refresh.create_reported_properties(
+        serialNumber="alohomora"
```
jebrando, line 219: You should make a const for this so the user can change it easily.

azure-iot-device/samples/pnp/pnp_temp_controller_with_thermostats.py
```diff
@@ -92,16 +93,92 @@ def create_max_min_report_response(values):
...
+        elif method_name:
+            command_name = method_name
+        else:
+            command_name = None
+
+        command_request = await device_client.receive_method_request(command_name)
+        print("Command request received with payload")
+        print(command_request.payload)
+
+        # TODO should npt need this once the command request envelope is removed.
```
jebrando, line 139: I'm not sure if we need this comment, and I'm not sure what it means.

olivakar: this is the TODO I was referring to...

PR #620, opened 2020-07-11
feature(azure-iot-device): Async IoTHub Client Handler Infrastructure

cartertinney:
* Added a handler manager to the async client, which will manage tasks and threads related to handlers
* Added additional client infrastructure to support future handler properties on async clients
* Updated InboxManager to allow for unified messages (input + C2D) in order to be compatible with the proposed handler properties
* Updated Message to allow for the user to see what input the message was sent to


azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -85,6 +90,46 @@ def _on_disconnected(self):
         self._inbox_manager.clear_all_method_requests()
         logger.info("Cleared all pending method requests due to disconnect")
 
+    async def _enable_feature(self, feature_name):
+        """Enable an Azure IoT Hub feature
```
olivakar, line 94: these enable and disable are not new methods only the else part is new right ?

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -350,6 +384,26 @@ async def notify_blob_upload_status(
         await handle_result(callback)
         logger.info("Successfully notified blob upload status")
 
+    @property
+    def on_method_request_received(self):
+        return self._on_method_request_received
+
+    @on_method_request_received.setter
+    def on_method_request_received(self, value):
```
olivakar, line 392: >on_method_request_received [](start = 8, length = 26)

similar setter and getters will eventually exist for the message received and twin patches right ?

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
cartertinney: Actually there's nothing added at all - they're just moved to a different part of the file to better organize it

---
In reply to: [454010147](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454010147) [](ancestors = 454010147)

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
cartertinney: Yes. In fact, this shouldn't even be here yet - this is only going to be added in a separate PR - I will remove before merging.

---
In reply to: [454011418](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454011418) [](ancestors = 454011418)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -79,6 +85,40 @@ def __init__(self, mqtt_pipeline, http_pipeline):
...
 
+        self._inbox_manager = None  # this will be overriden in child class
+        self._receive_type = RECEIVE_TYPE_NONE_SET
+
+        # Generic Client handlers
+        self._on_method_request_received = None
+        self._on_twin_desired_properties_patch_received = None
+        self._on_message_received = None
+
+    def _validate_receive_api_invoke(self):
```
BertKleewein, line 96: I hate it when people open this can of worms, especially since a problem is extremely unlikely, but this function isn't thread safe and there's nothing protecting it on the calling side.  Same for the other _validate function.

azure-iot-device/azure/iot/device/iothub/inbox_manager.py
```diff
@@ -25,11 +25,24 @@ def __init__(self, inbox_type):
...
+        # deprecated on the client. However we need to keep them functional for backwards
+        # compatibility
         self.c2d_message_inbox = self._create_inbox()
         self.input_message_inboxes = {}
-        self.generic_method_request_inbox = self._create_inbox()
         self.named_method_request_inboxes = {}
-        self.twin_patch_inbox = self._create_inbox()
+
+        # Set this value to True if want to only use unified message mode
+        self.use_unified_msg_mode = False
```
BertKleewein, line 40: since use_unified_msg_mode is True if and only if self.unified_message_inbox is set, can you remove this variable and just check that unified_msg_inbox is truthy instead?  Otherwise there's a risk that use_unified_msg_mode will be True, but unified_msg_inbox will be None (or vise-versa).

BertKleewein: Shoot, you always allocate unified_msg_inbox even if you don't use it.  Maybe this isn't so easy. 

---
In reply to: [454407805](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454407805) [](ancestors = 454407805)

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -350,6 +384,26 @@ async def notify_blob_upload_status(
...
+        self._validate_receive_handler_setter()
+        # Enable the feature if necessary
+        if value is not None and not self._mqtt_pipeline.feature_enabled[constant.METHODS]:
+            loop = asyncio.get_event_loop()
```
BertKleewein, line 396: >loop = asyncio.get_event_loop() [](start = 12, length = 31)

huh, so I guess we can't have async setters.  Random comment: Longer term, we'll probably need to scrub the code to remove all `import asyncio` in favor of asyncio_compat.

azure-iot-device/azure/iot/device/iothub/models/message.py
```diff
@@ -43,7 +43,9 @@ def __init__(
         self.user_id = None
         self.content_encoding = content_encoding
         self.content_type = content_type
+        # Open Question: could input_name and output_name fields be combined somehow (destination?)
```
BertKleewein, line 46: ># Open Question: could input_name and output_name fields be combined somehow (destination?) [](start = 8, length = 91)

personal opinion: I don't see an advantage and you'd actually remove some of the meaning if you did that.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_events_iothub.py
```diff
@@ -21,6 +21,8 @@ def __init__(self, message):
         self.message = message
 
 
+# TODO: refactor to not need separate input name
+# The message already contains the input name, it's not necessary to maintain here as well
```
BertKleewein, line 25: yes!

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
```diff
@@ -812,6 +812,18 @@ def test_message(self, event, stage, mock_mqtt_topic):
...
         assert message.data == event.payload
 
+    @pytest.mark.it(
+        "Extracts the input name from the topic, and attaches it to the newly created Message"
+    )
+    def test_input_name_on_message(self, event, stage, mock_mqtt_topic, input_name):
+        stage.handle_pipeline_event(event)
+
+        # Get Message that was created
+        assert mock_mqtt_topic.extract_message_properties_from_topic.call_count == 1
```
BertKleewein, line 822: >mock_mqtt_topic [](start = 15, length = 15)

is mock_mqtt_topic a mock or is it a spy?  I don't know if "spy" is a common term, but that's what the node mocking framework calls something that wraps an object so you can spy on calls to that object.  It looks like this test is using the real extract_message_properties_from_topic function, and it's just spying on the calls.  I think if it were a mock, the real function wouldn't be called.  
(probably don't want to change this now, but I did get confused reading this test, but I might just be confused in general.)

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -472,6 +477,88 @@ def test_no_message_in_inbox_nonblocking_mode(self, client, method_name):
         result = client.receive_method_request(method_name, block=False)
         assert result is None
 
+    @pytest.mark.it("Locks the client to API Receive Mode if the receive mode has not yet been set")
```
BertKleewein, line 480: observation: users can't mix api receive mode and handlers in the same app.  this is a great decision.  we need to make sure it's documented.

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -374,6 +377,62 @@ async def test_called_with_method_name_returns_method_request_from_named_method_
...
 
+    @pytest.mark.it("Locks the client to API Receive Mode if the receive mode has not yet been set")
+    @pytest.mark.parametrize(
+        "method_name",
+        [pytest.param(None, id="Generic Method"), pytest.param("method_x", id="Named Method")],
+    )
+    async def test_receive_mode_not_set(self, mocker, client, method_name):
+        # patch this so receive_method_request won't block
+        mocker.patch.object(
+            AsyncClientInbox, "get", return_value=(await create_completed_future(None))
```
BertKleewein, line 388: >create_completed_future [](start = 57, length = 23)

this is a mind-bender.  having create_completed_future be a coroutine seems silly -- I would expect `await create_completed_future()` to await on the completed future.
No action requested.  Just pointing out my confusion.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        self._on_twin_desired_properties_patch_received = None
+
+        # Other handlers
+        # TODO: add
+
+    async def _run_inbox_handler(self, inbox, handler_name):
+        """Run infinite loop that waits for an inbox to receive an object from it, then calls
+        the handler with that object
+        """
+        # Run the handler in a threadpool, so that it cannot block other handlers (from a different task),
```
BertKleewein, line 52: ># Run the handler in a threadpool, so that it cannot block other handlers (from a different task), [](start = 8, length = 98)

I'm having a hard time convincing myself that this is the right decision.  mixing threads and coroutines seems dangerous one of the big advantages of coroutines is that you don't have to worry about multiple threads so you don't have to worry about synchronizing those threads.  you're breaking that assumption and I don't know if that is cool.

I read (part of) a thread on this.  I'll forward it because there's a lot of good info & discussions in it.

BertKleewein: that being said, I don't think this needs to block your PR.  This particular decision can always change because the rest of your code isn't tightly bound to this.    

---
In reply to: [454548409](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454548409) [](ancestors = 454548409)

azure-iot-device/azure/iot/device/iothub/inbox_manager.py
cartertinney: Yeah, and you can't choose which it is at instantiation, because the mode is chosen by user action after the client is created. It's a very unfortunate situation.

---
In reply to: [454410693](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454410693) [](ancestors = 454410693,454407805)

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub_mqtt.py
cartertinney: No, it is a mock. A mock's attributes are all also mocks. So we are looking at the `extract_message_properties_from_topic` attribute of the mock, which is itself a mock, and thus we can check how many times it has been called.

---
In reply to: [454521437](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454521437) [](ancestors = 454521437)

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
cartertinney: Async setters aren't supported by the @property decorator, no. There's actually a bit of an issue that this creates that I want to address before we start exposing these to users.

---
In reply to: [454511914](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454511914) [](ancestors = 454511914)

azure-iot-device/tests/iothub/test_sync_clients.py
cartertinney: Yeah, this will go in the docstrings and docs when the functionality is exposed

---
In reply to: [454524064](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454524064) [](ancestors = 454524064)

azure-iot-device/tests/iothub/aio/test_async_clients.py
cartertinney: I agree - fundamentally we have this problem in a lot of async tests - mocks just don't play well with coroutines

---
In reply to: [454526462](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454526462) [](ancestors = 454526462)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
cartertinney: I also tried to avoid this, but in order to get the performance we want, for various reasons, it has to be this way. Let's discuss further on a call.

---
In reply to: [454549029](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454549029) [](ancestors = 454549029,454548409)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        if handler_name == METHOD:
+            return self._inbox_manager.get_method_request_inbox()
+        elif handler_name == TWIN_DP_PATCH:
+            return self._inbox_manager.get_twin_patch_inbox()
+        elif handler_name == MESSAGE:
+            return self._inbox_manager.get_unified_message_inbox()
+        else:
+            return None
+
+    def _add_handler_task(self, handler_name):
```
BertKleewein, line 87: >_add_handler_task [](start = 8, length = 17)

Don't we want to store the handler _coroutine_ instead of the handler _task_?  If it's a handler _task_, I think that means it's a piece of _running_ code -- the coroutine has already been invoked..  Once the task it complete, you can't run it again.  A coroutine, on the other hand, is like a function.  you can invoke it over and over.  (disclaimer: me no good with words.  I might have this wrong.)

azure-iot-device/tests/iothub/aio/test_async_handler_manager.py
```diff
@@ -0,0 +1,264 @@
...
+        def __init__(self):
+            self.handler_called = False
+            self.handler_call_count = 0
+            self.handler_call_arg = None
+
+    return HandlerChecker()
+
+
+@pytest.fixture(params=["Handler function", "Handler coroutine"])
+def handler(mocker, request, handler_checker):
```
olivakar, line 52: >mocker [](start = 12, length = 6)

possibly don't need the mocker here

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
cartertinney: No. The task is a "while True" loop that forever 1) waits for an inbox to get an item, then 2) calls the latest handler. By doing it this way it gives us a few abilities that are very useful, such as being able to update a handler without affecting the flow of the program, or be able to cancel the task if the handler is set back to None.

Because it's a while True loop, the task never completes. It runs forever until we cancel it

---
In reply to: [454574346](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454574346) [](ancestors = 454574346)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Good catch

---
In reply to: [454403531](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454403531) [](ancestors = 454403531)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
cartertinney: We have discussed to potentially change this design in a future PR that address widespread client performance issues related to customer event loops

---
In reply to: [454567833](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r454567833) [](ancestors = 454567833,454549029,454548409)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -79,6 +86,45 @@ def __init__(self, mqtt_pipeline, http_pipeline):
...
+        if self._receive_type is RECEIVE_TYPE_NONE_SET:
+            # Lock the client to ONLY use receive APIs (no handlers)
+            self._receive_type = RECEIVE_TYPE_API
+        elif self._receive_type is RECEIVE_TYPE_HANDLER:
+            raise exceptions.ClientError(
+                "Cannot use receive APIs - receive handler(s) have already been set"
+            )
+        else:
+            pass
+        self._client_lock.release()
```
BertKleewein, line 110: >self._client_lock.release() [](start = 8, length = 27)

this will not release the lock if an exception is raised.  Recommend using the lock as a context manager.  e..g `with self._client_lock:`

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+            # NOTE: we MUST use getattr here using the handler name, as opposed to directly passing
+            # the handler in order for the handler to be able to be updated without cancelling
+            # the running task created for this coroutine
+            handler = getattr(self, handler_name)
+            if inspect.iscoroutinefunction(handler):
+                # Wrap the coroutine in a function so it can be run in ThreadPool
+                def coro_wrapper(coro, arg):
+                    asyncio_compat.run(coro(arg))
+
+                tp.submit(coro_wrapper, handler, retval)
```
BertKleewein, line 67: >tp.submit(coro_wrapper, handler, retval) [](start = 16, length = 40)

Please don't change this since this code is going to be updated, but this line will swallow any errors that get raised from inside coro_wrapper.  It's a real hassle to get exceptions that are raised from threads.  Fortunately, the threadpool makes this easier by returning a concurrent.futures.Future object (which is not an asyncio.Future object.  what a wonderful world we live in.)   [https://stackoverflow.com/questions/2829329/catch-a-threads-exception-in-the-caller-thread-in-python](https://stackoverflow.com/questions/2829329/catch-a-threads-exception-in-the-caller-thread-in-python)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+    async def _run_inbox_handler(self, inbox, handler_name):
+        """Run infinite loop that waits for an inbox to receive an object from it, then calls
+        the handler with that object
+        """
+        # Run the handler in a threadpool, so that it cannot block other handlers (from a different task),
+        # or the main client thread. The number of worker threads forms an upper bound on how many instances
+        # of the same handler can be running simultaneously.
+        tp = ThreadPoolExecutor(max_workers=4)
+        while True:
+            retval = await inbox.get()
```
BertKleewein, line 57: >retval [](start = 12, length = 6)

better name please.  this is an argument, not a return value.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+            if inspect.iscoroutinefunction(handler):
+                # Wrap the coroutine in a function so it can be run in ThreadPool
+                def coro_wrapper(coro, arg):
+                    asyncio_compat.run(coro(arg))
+
+                tp.submit(coro_wrapper, handler, retval)
+            else:
+                # Run function directly in ThreadPool
+                tp.submit(handler, retval)
+
```
BertKleewein, line 71: same comment from 4 lines up on exceptions being swallowed.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+
+        inbox = self._get_inbox_for_handler(handler_name)
+
+        if inbox:
+            coro = self._run_inbox_handler(inbox, handler_name)
+        else:
+            coro = self._run_event_handler(handler_name)
+        task = asyncio_compat.create_task(coro)
+        self._handler_tasks[handler_name] = task
```
BertKleewein, line 103: >self._handler_tasks[handler_name] = task [](start = 8, length = 40)

What happens if the task raises an exception?  From what I've seen so far, it dies and becomes a dead husk that occupies the self._handler_tasks dictionary, creating exceptions and havoc for all eternity.  The task object has an add_done_callback that might be useful.

azure-iot-device/tests/iothub/aio/test_async_handler_manager.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+@pytest.fixture
+def handler_manager(inbox_manager):
+    return AsyncHandlerManager(inbox_manager)
+
+
+# ----------------------
+# We have to do some unfortunate things here in order to manually mock out handlers, to test
+# tha they have been called. We can't use MagicMocks because not only do they not work well
```
BertKleewein, line 34: >MagicMocks [](start = 42, length = 10)

Not sure if this helps or not, but you might look at the unittest.mock.AsyncMock class.

BertKleewein:
My comments on exceptions being swallowed are significant and should be addressed, but I think it's safe to pull as-is since it's not exposed to the user.

azure-iot-device/tests/iothub/aio/test_async_handler_manager.py
cartertinney: That could somewhat help, but doesn't really address the second point - these handlers are themselves the things under test, so mocking them out inherently takes away from the value of the test itself

---
In reply to: [456686506](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r456686506) [](ancestors = 456686506)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Excellent catch

---
In reply to: [456676459](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r456676459) [](ancestors = 456676459)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
cartertinney: Good catch. I'm going to do something with this in the next PR, because a similar problem (although with a different solution) exists for sync as well + i'm unifying some of the architecture between the two

---
In reply to: [456684624](https://github.com/Azure/azure-iot-sdk-python/pull/620#discussion_r456684624) [](ancestors = 456684624)

Issue #604, opened 2020-06-24
[feature request] Configurable connection keep-alive setting in python.


mcisoc:
I'm developing an IoT device with a global reach in a constrained network environment. This means we have strict requirements on network usage. Previously we were using python v1 SDK, but after moving to v2 we see a spike in our network traffic due to increased keep-alive activity on the MQTT connection.

As the v1 SDK was based on the C SDK, we had a 4 minute keepalive (which is configurable, see https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support) now we have a 1 minute keepalive which is not configurable. This will lead to a 25-50% network usage increase. For comparison with a 4 minute keepalive over a 15 minute timespan you will use ~500 bytes of traffic. With 60 seconds keepalive the usage is ~2000 bytes of data (depending on IP/TCP implementation overhead). 

Ideally we would want to be able to configured a keepalive of 16 minutes. We close our connection after 15 minutes to conserve network traffic. With a configurable keep alive we can reduce our network traffic significantly. 

If we don't have this option, we will not meet our data usage requirements, actually we will use our total usage just for keepalive.  We would then have use the C SDK with a python wrapper, stop using a MS sdk and use our own MQTT broker, or stop using IoT Hub and use another MQTT broker which supports are data requirements.

AB#7526607

jebrando:
We have added this feature on the backlog and we will be keep you updated on the progress.

olivakar:
Released at 2.2.0

az-iot-builder-01:
@mcisoc, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

olivakar:
https://github.com/Azure/azure-iot-sdk-python/pull/635

PR #635, opened 2020-07-24
feat(azure-iot-device) : Add user configurable keep alive

olivakar:


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -49,7 +49,7 @@
 
 # Default keepalive.  Paho sends a PINGREQ using this interval
 # to make sure the connection is still open.
-DEFAULT_KEEPALIVE = 60
+DEFAULT_KEEPALIVE = 180
```
cartertinney, line 52: In my view, this default value should be maintained within the config object, so all our defaults are in one place.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -381,12 +383,16 @@ def connect(self, password=None):
...
+                    port=443,
+                    keepalive=self._keep_alive if self._keep_alive else DEFAULT_KEEPALIVE,
                 )
             else:
                 logger.info("Connect using port 8883 (TCP)")
                 rc = self._mqtt_client.connect(
-                    host=self._hostname, port=8883, keepalive=DEFAULT_KEEPALIVE
+                    host=self._hostname,
+                    port=8883,
+                    keepalive=self._keep_alive if self._keep_alive else DEFAULT_KEEPALIVE,
```
cartertinney, line 395: >keepalive=self._keep_alive if self._keep_alive else DEFAULT_KEEPALIVE, [](start = 20, length = 70)

Another advantage to moving the Default keepalive to the config is that we get rid of this conditional logic here. MQTTTransport will take whatever value is in the config, and doesn't need branching logic.

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -77,3 +82,23 @@ def _sanitize_cipher(cipher):
             raise TypeError("Invalid type for 'cipher'")
 
         return cipher
+
+    @staticmethod
+    def _rectify_keep_alive(keep_alive):
```
cartertinney, line 87: >rectify [](start = 9, length = 7)

Let's call it "_santize_keep_alive" for consistency with the other, similar method

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -77,3 +82,23 @@ def _sanitize_cipher(cipher):
...
+            # Not allowing a keep alive of 0 as this would mean frequent ping exchanges.
+            logger.error(
+                "'keep alive' can not be zero or negative. A default value of 'keep alive' will be used by the protocol."
+            )
+            keep_alive = None
+        elif keep_alive and keep_alive > constant.LOAD_BALANCER_LIMIT_SECS:
+            # This is Hub's maximum Load Balancer Limit before
+            logger.error(
+                "'keep alive' can not be more than 29 minutes. 'keep alive' will be set to max value of 29 minutes to continue."
+            )
```
cartertinney, line 101: IMO, raise a ValueError here instead of just setting the value as the ceiling

azure-iot-device/tests/common/test_mqtt_transport.py
```diff
@@ -400,24 +400,49 @@ def test_calls_paho_connect(
             host=fake_hostname, port=port, keepalive=mocker.ANY
         )
 
-    @pytest.mark.it("Passes DEFAULT_KEEPALIVE to paho connect function")
+    @pytest.mark.it("Passes keep alive to paho connect function")
     @pytest.mark.parametrize(
```
cartertinney, line 404: Actually, now that we have dynamic keepalive values, we can fold this test into the one above with a simple addition of the keep_alive parametrization (and an adjustment to the check on the keepalive call value on line 400)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -381,12 +383,16 @@ def connect(self, password=None):
             if self._websockets:
                 logger.info("Connect using port 443 (websockets)")
                 rc = self._mqtt_client.connect(
-                    host=self._hostname, port=443, keepalive=DEFAULT_KEEPALIVE
+                    host=self._hostname,
+                    port=443,
+                    keepalive=self._keep_alive if self._keep_alive else DEFAULT_KEEPALIVE,
```
BertKleewein, line 388: >self._keep_alive if self._keep_alive else DEFAULT_KEEPALIVE [](start = 30, length = 59)

You can also write this as `keepalive = self._keepalive or DEFAULT_KEEPALIVE`.

azure-iot-device/azure/iot/device/constant.py
```diff
@@ -13,6 +13,7 @@
 PROVISIONING_API_VERSION = "2019-03-31"
 SECURITY_MESSAGE_INTERFACE_ID = "urn:azureiot:Security:SecurityAgent:1"
 TELEMETRY_MESSAGE_SIZE_LIMIT = 262144
+LOAD_BALANCER_LIMIT_SECS = 1740
```
BertKleewein, line 16: I assume this is some max keepalive value.  In that case, I would recommend calling it MAX_KEEPALIVE and then put a comment that says the value comes from the limitations of the load balancer.

cartertinney: +1

---
In reply to: [461195404](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461195404) [](ancestors = 461195404)

azure-iot-device/setup.py
```diff
@@ -90,6 +90,7 @@
         "futures;python_version == '2.7'",
         "PySocks",
         "win-inet-pton;python_version == '2.7'",
+        "future",
```
BertKleewein, line 93: >future [](start = 9, length = 6)

are we actually importing this?

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: Or, does this come from constant.py?  the max keepalive is in there, why not the default?

---
In reply to: [461186383](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461186383) [](ancestors = 461186383)

azure-iot-device/azure/iot/device/common/pipeline/config.py
BertKleewein: If we raise an exception when the value is out of bounds, then we're not sanitizing it.  Maybe `_validate_keep_alive`?

---
In reply to: [461187069](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461187069) [](ancestors = 461187069)

azure-iot-device/azure/iot/device/common/pipeline/config.py
BertKleewein: yes,

---
In reply to: [461187294](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461187294) [](ancestors = 461187294)

azure-iot-device/azure/iot/device/common/pipeline/config.py
cartertinney: Good point - I had thought we were doing some type conversions as well, but upon closer look we aren't - maybe we should

---
In reply to: [461197405](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461197405) [](ancestors = 461197405,461187069)

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -77,3 +82,23 @@ def _sanitize_cipher(cipher):
             raise TypeError("Invalid type for 'cipher'")
 
         return cipher
+
+    @staticmethod
+    def _rectify_keep_alive(keep_alive):
+        if keep_alive and not isinstance(keep_alive, (int, float)):
```
cartertinney, line 88: >if keep_alive and not isinstance(keep_alive, (int, float)): [](start = 8, length = 59)

Why not just convert this to int, always? That would presumably get rid of the need for the builtins import as well

azure-iot-device/setup.py
cartertinney: Yeah, not sure what this is

---
In reply to: [461195945](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461195945) [](ancestors = 461195945)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Config objects have default values for if the value is not provided with the argument syntax of a function. So I would think the default value would be there

---
In reply to: [461196312](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461196312) [](ancestors = 461196312,461186383)

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -29,6 +32,7 @@ def __init__(
         websockets=False,
         cipher="",
         proxy_options=None,
+        keep_alive=None,
```
cartertinney, line 35: >None [](start = 19, length = 4)

This should probably be the default keepalive instead of None, as I mentioned in another comment

azure-iot-device/azure/iot/device/common/pipeline/config.py
olivakar: do you mean just try to convert to int and throw an error if conversion goes wrong ? the built in int is only used to support long for python 2.7.  Long for python 3 falls within the int range but for python 2 if someone passes 9L it will fail if its just the normal int. 

---
In reply to: [461199086](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461199086) [](ancestors = 461199086)

azure-iot-device/azure/iot/device/common/pipeline/config.py
olivakar: i am actually just logging errors and assigning correct values , i am not throwing errors as such.. i thought of validation in case of throwing errors but since i am not throwing errors i called it rectify...sanitize also can be used but i was not cleaning up value...i was actually correcting them...so if it is much larger use max , if it negative or zero  use default.

---
In reply to: [461198869](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461198869) [](ancestors = 461198869,461197405,461187069)

azure-iot-device/azure/iot/device/common/pipeline/config.py
olivakar: is it necessary for the user to know the bounds ?  I actually went back and forth with this as technically these are all correct values for keep alive but we have bounds simply because we have a load balancer... it is not a wrong value as such considering only the keep alive param but rather our load balancer will go wrong with these values...

---
In reply to: [461197621](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461197621) [](ancestors = 461197621,461187294)

azure-iot-device/setup.py
olivakar: this is to make the built ins import work...if we don't have this it does not work...this was surprising to me as well, but it complains that it does not have builtins

---
In reply to: [461199185](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461199185) [](ancestors = 461199185,461195945)

BertKleewein: The future module is 2.3 MB and it's only needed for python 2.7.  You should just convert to an int and catch ValueError. 

---
In reply to: [461941297](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r461941297) [](ancestors = 461941297,461199185,461195945)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -101,6 +97,7 @@ def __init__(
         websockets=False,
         cipher=None,
         proxy_options=None,
+        keep_alive=None,
```
BertKleewein, line 100: >keep_alive=None, [](start = 8, length = 16)

As you have this written, the connection will fail if the user doesn't provide a keep_alive value, so it's not _really_ optional.  (please don't make it required).

azure-iot-device/azure/iot/device/common/pipeline/config.py
```diff
@@ -77,3 +84,23 @@ def _sanitize_cipher(cipher):
             raise TypeError("Invalid type for 'cipher'")
 
         return cipher
+
+    @staticmethod
+    def _rectify_keep_alive(keep_alive):
```
BertKleewein, line 89: >_rectify_keep_alive [](start = 8, length = 19)

This function just feels wrong to me.  _sanitize_cipher is needed because there is more than one way to specify cipher and we want to normalize it.  You're doing something different.  You're silently fixing out-of-bounds errors, and I can't think of may examples of that. 
Python talks about LBYL and EAFP and this is neither.  It's like you're adding a third option: LMFTFY (Let Me Fix That For You)

azure-iot-device/azure/iot/device/common/mqtt_transport.py
olivakar: the value is defaulted at the config level , so it will never be None rather it will be always 180 secs...

---
In reply to: [462526084](https://github.com/Azure/azure-iot-sdk-python/pull/635#discussion_r462526084) [](ancestors = 462526084)

olivakar:
https://github.com/Azure/azure-iot-sdk-python/issues/604

PR #641, opened 2020-08-03
fix (azure-iot-device): Update DPS error messages

BertKleewein:


azure-iot-device/azure/iot/device/provisioning/provisioning_device_client.py
```diff
@@ -24,13 +24,17 @@ def handle_result(callback):
     try:
         return callback.wait_for_completion()
     except pipeline_exceptions.ConnectionDroppedError as e:
-        raise exceptions.ConnectionDroppedError(message="Lost connection to IoTHub", cause=e)
+        raise exceptions.ConnectionDroppedError(
+            message="Lost connection to the provisioning server", cause=e
```
olivakar, line 28: will `device provisioning service` be better ?

PR #644, opened 2020-08-15
feature(azure-iot-device): Handler Manager (part 2)

cartertinney:
* Added a HandlerManager to the synchronous client
* Created shared infrastructure for sync/async HandlerManagers
* Added rough event loop management infrastructure to protect async client processes from invoking code with poor performance (could be expanded in the future to be more robust, but it works just fine as is)
* Better error handling and logging for both sync and async HandlerManagers
* Modified how handler removal works - handlers now will be invoked for all pending items in the inbox at time of handler removal request, prior to the actual removal of the handler. Items added to the inbox AFTER handler removal request will NOT be invoked.
* Ensured iothub client is threadsafe in async contexts

Note that customer-provided handlers are all still invoked in their own threads (even coroutines). Further optimization there (specifically with regard to coroutines) could come in a separate PR whenever necessary.


azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -384,25 +384,31 @@ async def notify_blob_upload_status(
         await handle_result(callback)
         logger.info("Successfully notified blob upload status")
 
-    # @property
-    # def on_method_request_received(self):
-    #     return self._on_method_request_received
-
-    # @on_method_request_received.setter
-    # def on_method_request_received(self, value):
+    # def _generic_handler_setter(self, handler_name, feature_name, new_handler):
```
cartertinney, line 387: ># def _generic_handler_setter(self, handler_name, feature_name, new_handler): [](start = 4, length = 77)

This commented code will be refined and made available in the next PR - I left it in here because while it works, it's not something I wish to expose yet.

BertKleewein:
> def run(coro):

I'm hoping that we get rid of this function after you change the loop that customer coroutines get called in.  I didn't know just how dangerous this function was until very recently.

---
Refers to: azure-iot-device/azure/iot/device/common/asyncio_compat.py:60 in 6929799. [](commit_id = 6929799bfafde872ab341b3017e1ed1232e75966, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+logger = logging.getLogger(__name__)
+
+MESSAGE = "_on_message_received"
+METHOD = "_on_method_request_received"
+TWIN_DP_PATCH = "_on_twin_desired_properties_patch_received"
+# TODO: add more for "event"
+
+
+class HandlerManagerException(ChainableException):
```
BertKleewein, line 25: >HandlerManagerException [](start = 6, length = 23)

add doc please and pass a message into the base constructor to explain what this signifies.

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
```diff
@@ -0,0 +1,232 @@
...
+    pass
+
+
+class HandlerRunnerKillerSentinel(object):
```
BertKleewein, line 29: >HandlerRunnerKillerSentinel [](start = 6, length = 27)

doc please

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+@six.add_metaclass(abc.ABCMeta)
+class AbstractHandlerManager(object):
+    """Partial class that defines handler manager functionality shared between sync/async"""
+
+    def __init__(self, inbox_manager):
+        self._inbox_manager = inbox_manager
+
+        self._handler_runners = {
```
BertKleewein, line 40: >_handler_runners [](start = 13, length = 16)

I think this needs a better name.  "handler runner" sounds like to words that mean almost the same thing pasted together.  I don't know what that name would be.  handler_invocation_object?  words are not working for me right now.

cartertinney: I'm open to an alternative, but I've yet to think of something better. The thing is it's not an object - it's a function/coroutine that loops, calling the handler. i.e. it is a runner for a handler.

---
In reply to: [473419040](https://github.com/Azure/azure-iot-sdk-python/pull/644#discussion_r473419040) [](ancestors = 473419040)

cartertinney:
> def run(coro):

Yeah this really shouldn't be called except at the top level of customer code. Ideally it's not in here anywhere (although this implementation is almost exactly how asyncio.run() actually works). It won't be necessary once we call customer code in their original loop, but I'm not sure how high of a priority that's going to be - there's probably a couple things that are of higher priority.

---
In reply to: [676789006](https://github.com/Azure/azure-iot-sdk-python/pull/644#issuecomment-676789006) [](ancestors = 676789006)

---
Refers to: azure-iot-device/azure/iot/device/common/asyncio_compat.py:60 in 6929799. [](commit_id = 6929799bfafde872ab341b3017e1ed1232e75966, deletion_comment = False)

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
cartertinney: I believe the message will automatically go into the base constructor due to inheritance. However, yes I will add some doc.

---
In reply to: [473407912](https://github.com/Azure/azure-iot-sdk-python/pull/644#discussion_r473407912) [](ancestors = 473407912)

azure-iot-device/azure/iot/device/iothub/aio/async_inbox.py
```diff
@@ -4,10 +4,24 @@
...
 from azure.iot.device.iothub.sync_inbox import AbstractInbox
 
+# This logic could potentially be encapsulated in another module...
+# Say... some kind of... loop_manager.py?
+INBOX_LOOP = asyncio.new_event_loop()
+INBOX_THREAD = threading.Thread(target=INBOX_LOOP.run_forever)
+INBOX_THREAD.daemon = True
+INBOX_THREAD.start()
+
+# IMPLEMENTATION NOTE: The janus Queue exists entirely on the above mentioned INBOX_LOOP,
```
BertKleewein, line 19: >The janus Queue exists entirely on the above mentioned INBOX_LOOP [](start = 23, length = 65)

Nice hack.  Like you say, not pretty, but it's simple and it's the right size solution.  Adding a loop manager would be overkill IMO.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -5,150 +5,166 @@
...
         """Run infinite loop that waits for an inbox to receive an object from it, then calls
         the handler with that object
         """
+        logger.debug("HANDLER RUNNER ({}): Starting runner".format(handler_name))
+
+        # Define a callback that can handle errors in the ThreadPoolExecutor
+        def _handler_callback(future):
+            try:
+                e = future.exception()
+            except concurrent.futures.CancelledError as raised_e:
```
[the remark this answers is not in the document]
BertKleewein: I was wondering if you actually needed this exception handler so I wrote some test cases and I only found one case where this handler is needed. 

In order to hit this code, you need to submit a function to the thread pool, and then cancel the future (by calling future.cancel) __before it starts executing__.  

I could only think of 6 cases to test:
1 function starts running and  future.cancel is called before function is complete: function is not cancelled and no exception is raised 
2. function stars running and future.result(timeout=1) is called before the function completes: function is not cancelled and no exception is raised
3. function starts running and future.exception(timeout=1) is called before the function completes: function is not cancelled and no exception is raised.
4. function doesn't start running (because all threads are busy) and future.cancel is called.  function is cancelled, callback is called, future.exception raises a CancelledError.
5. function doesn't start running and future.result(timeout=1) is called.  future.result() raises CancelledError.  function is cancelled, and callback is not even called.
6. function doesn't start running and future.exception(timeout=1) is called.  future.exception() raises CancelledError.  function is cancelled, and callback is not even called.

This behavior is consistent across 2.7 and 3.5-3.8.

This is annoying and lame.



---
In reply to: [474007224](https://github.com/Azure/azure-iot-sdk-python/pull/644#discussion_r474007224) [](ancestors = 474007224)

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -5,150 +5,166 @@
...
+                new_err = HandlerManagerException(
+                    message="HANDLER ({}): Invocation unexpectedly ended with cancellation".format(
+                        handler_name
+                    ),
+                    cause=raised_e,
+                )
+                handle_exceptions.handle_background_exception(new_err)
```
BertKleewein, line 50: >handle_exceptions.handle_background_exception(new_err) [](start = 16, length = 54)

If, by some chance, future.exception raises some other error, (which it _shouldn't_  AFAIK), that error will be swallowed.  Suggest adding a catch-all here instead of just catching CancelledError.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -5,150 +5,166 @@
...
-    def on_twin_desired_properties_patch_received(self, value):
-        self._generic_handler_setter(TWIN_DP_PATCH, value)
+            coro = self._event_handler_runner(handler_name)
+        future = asyncio.run_coroutine_threadsafe(coro, RUNNER_LOOP)
+
+        # Define a callback for the future (in order to handle any errors)
+        def _handler_runner_callback(completed_future):
+            try:
+                e = completed_future.exception()
+            except asyncio.CancelledError:
```
BertKleewein, line 124: >except asyncio.CancelledError: [](start = 12, length = 30)

comments above also apply here.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -5,150 +5,166 @@
...
+                handle_exceptions.handle_background_exception(new_err)
+            else:
+                if e:
+                    new_err = HandlerManagerException(
+                        message="HANDLER RUNNER ({}): Unexpected error during task".format(
+                            handler_name
+                        ),
+                        cause=e,
+                    )
+                    handle_exceptions.handle_background_exception(new_err)
```
BertKleewein, line 140: >handle_exceptions.handle_background_exception(new_err) [](start = 20, length = 54)

This is a very underwhelming response to a very serious error.  if the handler runner stops running, then the handler will never be called again.  App developers have no way to detect this condition unless they happen to notice the background exception or the lack of handlers being called.  This feels fatal to me.  It would be better to crash spectacularly or try to restart the handler runner.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -5,150 +5,166 @@
...
-MESSAGE = "_on_message_received"
-METHOD = "_on_method_request_received"
-TWIN_DP_PATCH = "_on_twin_desired_properties_patch_received"
-# TODO: add more for "event"
+# This logic could potentially be encapsulated in another module...
+# Say... some kind of... loop_manager.py?
+RUNNER_LOOP = asyncio.new_event_loop()
+RUNNER_THREAD = threading.Thread(target=RUNNER_LOOP.run_forever)
+RUNNER_THREAD.daemon = True
+RUNNER_THREAD.start()
```
BertKleewein, line 27: this makes me nervous.  importing a module should not start a thread.

azure-iot-device/azure/iot/device/iothub/aio/async_inbox.py
BertKleewein: That being said, the more I think about this, the more nervous I become.  I like the idea of an INBOX_LOOP, but I don't like the thread being created as a side-effect of importing the module.  Can't you wait to crate the thread/loop until you actually instantiate something?  Also, if INBOX_THREAD crashes (or otherwise stops running), can the application detect this and recover?  If INBOX_LOOP is a global like this, the only recovery is a reboot.  I'm not sure the app can even detect that this needs to happen though.

---
In reply to: [473995318](https://github.com/Azure/azure-iot-sdk-python/pull/644#discussion_r473995318) [](ancestors = 473995318)

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
BertKleewein: Let's drop it for now.  I can't think of anything better.

---
In reply to: [473481960](https://github.com/Azure/azure-iot-sdk-python/pull/644#discussion_r473481960) [](ancestors = 473481960,473419040)

BertKleewein:
Signing off even though I suspect you are going to change things based on feedback, I think it's good enough to go in because it's not exposed.  My comment with "underwhelming response" is maybe the most consequential .  Others less so.

azure-iot-device/azure/iot/device/iothub/sync_handler_manager.py
```diff
@@ -0,0 +1,232 @@
...
+                    "HANDLER RUNNER ({}): HandlerRunnerKillerSentinel found in inbox. Exiting.".format(
+                        handler_name
+                    )
+                )
+                tpe.shutdown()
+                break
+            # NOTE: we MUST use getattr here using the handler name, as opposed to directly passing
+            # the handler in order for the handler to be able to be updated without cancelling
+            # the running task created for this coroutine
+            handler = getattr(self, handler_name)
```
olivakar, line 184: this handler is constant for the inbox and name right ? we are only in a loop to handle various items in the same inbox ? so any reason for doing this inside the loop

Issue #633, opened 2020-07-24
IoTHubModuleClient.invoke_method incorrect function parameter documentation


dapisani:
The method_params parameter for invoke_method is stated as
        :param dict method_params: Should contain a method_name, payload, connect_timeout_in_seconds, response_timeout_in_seconds.

When invoking a method in which the method_params dictionary has the aforementioned keys you will receive a 402 error.

By changing to camel case such that
        :param dict method_params: Should contain a methodName, payload, connectTimeoutInSeconds, responseTimeoutInSeconds.

the invocation works as expected.

Also the description should state what the value types should be in the method_parms dictionary. It was unclear if the payload value should be an object, a json string or some other type.

AB#7788584

cartertinney:
Fixed in #655

az-iot-builder-01:
@dapisani, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #649, opened 2020-08-25
chore (azure-iot-device): change message severity for more consistent logging

BertKleewein:


BertKleewein:
# Guidelines for picking levels
Please debate me if these seem wrong.  This gets very fuzzy after a while

## `logger.critcal`
(very rare)

Something bad has happened and we can no longer guarantee that things are going to work.

Examples
* threading.Thread() call raises an Exception

## `logger.error`
(rare)

Something bad has happened, and we either couldn't report it to you or it's so bad that you need to investigate it.

Examples:
* Unexpected state
    * operation still not handled at last stage
    * attempting to complete an already-completed op.
    * No onConnected callback is registered (error becasue our code always registers this).
* Our code raises an exception inside of a transport callback, we catch it and use logger.error to report it, but don't want to crash the transport.


## `logger.warning`
(very rare)

Something happened that we couldn't report to you, but we think it might not be _that_ bad.

Examples:
* Message received on a module input, but user is not subscribed for that input.

## `logger.info`
(common)

Something significant happened, but you only get the high level info.

Examples:
* User called into our code and we started something
* We returned an error to the user
* We called into the transport
* The transport calls into us
* An error was created inside our code out of something that wasn't an error.  e.g. if queue_size == 0: raise Exception("queue should not be empty").
* Errors that come from another layer do not warrant a logger.info call.
* A new operation was initiated as a side-effect of another.  e.g. starting a connection on publishing
* A new event is crated as a side-effect of something unrelated.
* Completion of an operation is cut-short.  e.g.  ConnectionOp() completes early because transport is already connected.
* We receive a message on a topic that isn't registered or the MID or RID is unexpected
* An op is put into a queue.
* The queue unblocks and ops start again
* A timer caused something new to start.  e.g. "starting sas token renewal".

## `logger.debug`
(super-duper common)

Anything else.

Examples:
* Calling into some internal object.
* An error is returned from some internal object
* An internal object got a call from some other internal object
* An internal object returned an error to some other internal object.
* An operation is completed or ignored
* Details on some internal structure or some internal processing.  e.g. "topic parsed.  method name=blah"
* Progress report on some internal flow.  (e.g. Calling into SSL, done initializing pipeline)
* State changes.  (e.g. "CONACK received.  Setting connected=True")
* Internal handoff (e.g. "incoming message put into message inbox")
* Internal object created or destroyed.
* Timer start, stop, run

olivakar:
I agree with most except the point here of returning an error to a user via logger.info. Second point on logger.info

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -706,7 +714,7 @@ def _handle_pipeline_event(self, event):
             self.send_event_up(event)
 
             for request_id in self.pending_responses:
-                logger.info(
+                logger.debug(
```
olivakar, line 717: this is the one i possibly changed...conflict may arise when merging..

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -42,7 +42,7 @@ def _run_op(self, op):
             # sees this -1, it will send a GetTwinOperation to refresh desired properties.
 
             if op.feature_name == constant.TWIN_PATCHES:
-                logger.info(
+                logger.debug(
```
olivakar, line 45: i feel this should be info as it contains information about enabling a certain feature

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -108,7 +107,7 @@ def _handle_pipeline_event(self, event):
         if isinstance(event, pipeline_events_iothub.TwinDesiredPropertiesPatchEvent):
             # remember the $version when we get a patch.
             version = event.patch["$version"]
-            logger.info(
+            logger.debug(
```
olivakar, line 110: may be this should be info as well

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -115,7 +115,7 @@ def do_retry_after():
             original_provisioning_op.completed = False
             this.run_op(original_provisioning_op)
 
-        logger.warning(
+        logger.debug(
```
olivakar, line 118: this retry is happening coz DPS said so and not from our side..should this be an info then ?

azure-iot-device/azure/iot/device/common/async_adapter.py
```diff
@@ -70,9 +70,9 @@ def wrapping_callback(*args, **kwargs):
                 result = None
 
             if exception:
-                # Do not use exc_info parameter on logger.error.  This casuses pytest to save the traceback which saves stack frames which shows up as a leak
-                logger.error("Callback completed with error {}".format(exception))
-                logger.error(traceback.format_exception_only(type(exception), exception))
+                # Do not use exc_info parameter on logger.* calls.  This casuses pytest to save the traceback which saves stack frames which shows up as a leak
+                logger.info("Callback completed with error {}".format(exception))
```
olivakar, line 74: do we return this error ? if so info is fine...asking otherwise

azure-iot-device/azure/iot/device/common/evented_callback.py
```diff
@@ -44,9 +44,9 @@ def wrapping_callback(*args, **kwargs):
                     )
 
             if self.exception:
-                # Do not use exc_info parameter on logger.error.  This casuses pytest to save the traceback which saves stack frames which shows up as a leak
-                logger.error("Callback completed with error {}".format(self.exception))
-                logger.error(traceback.format_exc())
+                # Do not use exc_info parameter on logger.* calls.  This casuses pytest to save the traceback which saves stack frames which shows up as a leak
+                logger.info("Callback completed with error {}".format(self.exception))
```
olivakar, line 48: do we return this error ? if so info is fine...asking otherwise

azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1517,7 +1517,7 @@ class TestIoTHubModuleClientReceiveTwinDesiredPropertiesPatch(
 class TestIoTHubModuleClientInvokeMethod(IoTHubModuleClientTestsConfig):
     @pytest.mark.it("Begins a 'invoke_method' HTTPPipeline operation where the target is a device")
     async def test_calls_pipeline_invoke_method_for_device(self, mocker, client, http_pipeline):
-        method_params = "__fake_method_params__"
+        method_params = {"methodName": "__fake_method_name__"}
```
cartertinney, line 1520: >"methodName" [](start = 25, length = 12)

In the docstring we actually say it's "method_name". Not sure which is correct

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
cartertinney: This process of enabling is hidden to the user though due to being implicit though. They aren't really exposed to the concept of features being enabled/disabled

---
In reply to: [477604104](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r477604104) [](ancestors = 477604104)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -182,16 +182,16 @@ def _run_op(self, op):
             try:
                 self.transport.connect(password=password)
             except Exception as e:
-                logger.error("transport.connect raised error")
-                logger.error(traceback.format_exc())
+                logger.info("transport.connect raised error")
```
cartertinney, line 185: >info [](start = 23, length = 4)

I feel like this could probably remain debug - The user has no insight into what transport.connect is. It's not particularly "informative" - plus when the error gets raised they'll get something that's actually useful at high level

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -601,7 +601,8 @@ def establish_operation(self, mid, callback=None):
                     logger.error("Unexpected error calling callback for MID: {}".format(mid))
                     logger.error(traceback.format_exc())
             else:
-                logger.exception("No callback for MID: {}".format(mid))
+                # Not entirely unexpected becuase of QOS=1
+                logger.info("No callback for MID: {}".format(mid))
```
cartertinney, line 606: I feel this could probably be debug. Other MID related logs are all debug.

cartertinney:
I don't have a lot of strong opinions on logging levels. Left a few comments on things that stood out to me though.

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -69,15 +69,15 @@ def _form_complete_result(operation_id, decoded_response, status):
...
+        logger.info(
             "{stage_name}({op_name}): Received error with status code {status_code} for {prov_op_name} request operation".format(
                 stage_name=self.name,
                 op_name=request_response_op.name,
                 prov_op_name=request_response_op.request_type,
                 status_code=request_response_op.status_code,
             )
         )
-        logger.error(
+        logger.info(
```
cartertinney, line 80: >info [](start = 15, length = 4)

I feel this (and the one above) could probably be debug - In general I feel like info should probably be higher level stuff than this, which means very little to the user.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -128,11 +127,6 @@ def complete(self, error=None):
                 if self.completed:
                     # This block should never be reached - this is an invalid state.
                     # If this block is reached, there is a bug in the code.
```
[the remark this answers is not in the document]
BertKleewein: handle_background_exception will log the error for us. I removed a few cases like this which were effectively logging the same error twice.

---
In reply to: [477576652](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r477576652) [](ancestors = 477576652)

azure-iot-device/tests/iothub/aio/test_async_clients.py
BertKleewein: methodName is what the implementation has.  Changing it to method_name at this point would be a breaking change.  This is why we have [https://github.com/Azure/azure-iot-sdk-python/issues/633](https://github.com/Azure/azure-iot-sdk-python/issues/633) 

---
In reply to: [478767548](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r478767548) [](ancestors = 478767548)

azure-iot-device/azure/iot/device/common/evented_callback.py
BertKleewein: Yes, we raise this exception when the caller calls wait_for_completion (line 70 below)

---
In reply to: [477605946](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r477605946) [](ancestors = 477605946)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
BertKleewein: I'm following the convention that errors that get returned from external code get logged at info level.  Yes,m the user doesn't know what transport.connect is, but this is useful in understanding why the code is changing behavior.  There's also a certain amount of "expectedness" that I'm factoring in.   We don't expect `transport.connect` to raise an exception.  If it does, we need to "call it out" using `logger.info`

---
In reply to: [478833073](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r478833073) [](ancestors = 478833073)

azure-iot-device/azure/iot/device/common/async_adapter.py
BertKleewein: Yes, it gets raised when the user awaits the future.  

---
In reply to: [477605792](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r477605792) [](ancestors = 477605792)

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
BertKleewein: No, but I think the one below should be.

---
In reply to: [477604528](https://github.com/Azure/azure-iot-sdk-python/pull/649#discussion_r477604528) [](ancestors = 477604528)

PR #651, opened 2020-08-27
feature(azure-iot-device): Handlers Part 3

cartertinney:
* Enabled handlers via the clients (sync and async)
* Added client tests for handlers
* Added initial loop management infrastructure

NOTE: Samples will be updated in a forthcoming PR, as will formal deprecation of the old receive-style APIs
NOTE ALSO: Using handlers currently will create some warnings upon program exit. This will be fixed later, it does not affect performance.

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
@@ -128,7 +128,7 @@ class GenericIoTHubClient(AbstractIoTHubClient):
             logger.info("Successfully disabled feature: {}".format(feature_name))
         else:
             # This branch shouldn't be reached, but in case it is, log it
-            logger.info("Feature ({}) already enabled - skipping".format(feature_name))
+            logger.info("Feature ({}) already disabled - skipping".format(feature_name))
 
     async def connect(self):
         """Connects the client to an Azure IoT Hub or Azure IoT Edge Hub instance.
@@ -384,25 +384,31 @@ class GenericIoTHubClient(AbstractIoTHubClient):
         await handle_result(callback)
         logger.info("Successfully notified blob upload status")
 
+    # def _generic_handler_setter(self, handler_name, feature_name, new_handler):
+    #     self._validate_receive_handler_setter()
+    #     # Set the handler on the handler manager
+    #     setattr(self._handler_manager, handler_name, new_handler)
+
+    #     # Enable the feature if necessary
+    #     if new_handler is not None and not self._mqtt_pipeline.feature_enabled[feature_name]:
+
+    #         # CT-TODO: fix this to work better
+    #         loop = asyncio.get_event_loop()
+    #         asyncio.ensure_future(self._enable_feature(feature_name), loop=loop)
+
+    #     # Disable the feature if necessary
+    #     elif new_handler is None and self._mqtt_pipeline.feature_enabled[feature_name]:
+    #         # this works in async context, but not in sync
+    #         loop = asyncio.get_event_loop()
+    #         asyncio.ensure_future(self._disable_feature(feature_name), loop=loop)
+
     # @property
     # def on_method_request_received(self):
     #     return self._on_method_request_received
 
     # @on_method_request_received.setter
     # def on_method_request_received(self, value):
-    #     self._validate_receive_handler_setter()
-    #     # Enable the feature if necessary
-    #     if value is not None and not self._mqtt_pipeline.feature_enabled[constant.METHODS]:
-    #         loop = asyncio.get_event_loop()
-    #         asyncio.ensure_future(self._enable_feature(constant.METHODS), loop=loop)
-
-    #     # Disable the feature if necessary
-    #     elif value is None and self._mqtt_pipeline.feature_enabled[constant.METHODS]:
-    #         loop = asyncio.get_event_loop()
-    #         asyncio.ensure_future(self._disable_feature(constant.METHODS), loop=loop)
-
-    #     # Set the handler on the handler manager
-    #     self._handler_manager.on_method_request_received = value
+    #     self._generic_handler_setter("on_method_request_received", constant.METHODS, value)
 
 
 class IoTHubDeviceClient(GenericIoTHubClient, AbstractIoTHubDeviceClient):
--- a/azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
@@ -384,31 +384,40 @@ class GenericIoTHubClient(AbstractIoTHubClient):
         await handle_result(callback)
         logger.info("Successfully notified blob upload status")
 
-    # def _generic_handler_setter(self, handler_name, feature_name, new_handler):
-    #     self._validate_receive_handler_setter()
-    #     # Set the handler on the handler manager
-    #     setattr(self._handler_manager, handler_name, new_handler)
+    def _generic_handler_setter(self, handler_name, feature_name, new_handler):
+        self._validate_receive_handler_setter()
+        # Set the handler on the handler manager
+        setattr(self._handler_manager, handler_name, new_handler)
 
-    #     # Enable the feature if necessary
-    #     if new_handler is not None and not self._mqtt_pipeline.feature_enabled[feature_name]:
+        # Enable the feature if necessary
+        if new_handler is not None and not self._mqtt_pipeline.feature_enabled[feature_name]:
+            # CT-TODO: Make this callable outside coroutine by adding loop management
+            loop = asyncio.get_event_loop()
+            asyncio.ensure_future(self._enable_feature(feature_name), loop=loop)
 
-    #         # CT-TODO: fix this to work better
-    #         loop = asyncio.get_event_loop()
-    #         asyncio.ensure_future(self._enable_feature(feature_name), loop=loop)
+        # Disable the feature if necessary
+        elif new_handler is None and self._mqtt_pipeline.feature_enabled[feature_name]:
+            # CT-TODO: Make this callable outside coroutine by adding loop management
+            loop = asyncio.get_event_loop()
+            asyncio.ensure_future(self._disable_feature(feature_name), loop=loop)
 
-    #     # Disable the feature if necessary
-    #     elif new_handler is None and self._mqtt_pipeline.feature_enabled[feature_name]:
-    #         # this works in async context, but not in sync
-    #         loop = asyncio.get_event_loop()
-    #         asyncio.ensure_future(self._disable_feature(feature_name), loop=loop)
+    @property
+    def on_twin_desired_properties_patch_received(self):
+        return self._handler_manager.on_twin_desired_properties_patch_received
 
-    # @property
-    # def on_method_request_received(self):
-    #     return self._on_method_request_received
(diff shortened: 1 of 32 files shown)


azure-iot-device/tests/iothub/test_sync_handler_manager.py
```diff
@@ -164,33 +164,43 @@ def test_handler_invoked_multiple(self, mocker, handler_name, handler_manager, i
         "Is invoked for every item already in the corresponding Inbox at the moment of handler removal"
     )
     def test_handler_resolve_pending_items_before_handler_removal(
-        self, mocker, handler_name, handler_manager, inbox
+        self, mocker, handler_name, handler_manager, handler, inbox
```
cartertinney, line 167: This change here is unrelated to the rest of the PR - it's a fix for a flaky test that was giving me trouble upon trying to check in

azure-iot-device/tests/iothub/aio/test_async_handler_manager.py
```diff
@@ -229,30 +229,38 @@ async def test_handler_invoked_multiple(
     async def test_handler_resolve_pending_items_before_handler_removal(
         self, mocker, handler_name, handler_manager, handler, handler_checker, inbox
     ):
-        # Set the handler
-        setattr(handler_manager, handler_name, handler)
-        # Add 100 items to the associated inbox (enough to build up a backlog)
+        assert inbox.empty()
```
cartertinney, line 232: This change here is unrelated to the rest of the PR - it's a fix for a flaky test that was giving me trouble upon trying to check in

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
```diff
@@ -0,0 +1,72 @@
...
+    loop_thread.daemon = True
+    loop_thread.start()
+    # Store the loop
+    loops[loop_name] = new_loop
+
+
+# def get_user_loop():
+#     """Return the loop that was active during client instantiation"""
+#     # TODO: this is insufficient
+#     return loops["USER_LOOP"]
```
cartertinney, line 58: This feature to be used in async client for sending handlers back to the user's event loop will come later in a different PR. It's going to take a bit of finessing.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -369,23 +369,45 @@ def receive_twin_desired_properties_patch(self, block=True, timeout=None):
         logger.info("twin patch received")
         return patch
 
-    # @property
-    # def on_method_request_received(self):
-    #     return self._on_method_request_received
+    def _generic_handler_setter(self, handler_name, feature_name, new_handler):
+        self._validate_receive_handler_setter()
```
BertKleewein, line 373: >_validate_receive_handler_setter [](start = 13, length = 32)

slight name confusion: _validate_receive_handler_setter in this context makes it sound like you're validating the setter itself.  maybe _validate_use_of_receive_handlers and _validate_use_of_receive_apis?

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -576,3 +610,15 @@ def invoke_method(self, method_params, device_id, module_id=None):
         invoke_method_response = handle_result(callback)
         logger.info("Successfully invoked method")
         return invoke_method_response
+
+    @property
+    def on_message_received(self):
+        """The handler function that will be called when a message is received.
```
BertKleewein, line 616: >The handler function that will be called when a message is received. [](start = 11, length = 68)

should say "when an input message is received".  If this is going to be public doc, then you should mention the name of the message attribute that holds the input name.

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -547,3 +583,15 @@ async def invoke_method(self, method_params, device_id, module_id=None):
         method_response = await handle_result(callback)
         logger.info("Successfully invoked method")
         return method_response
+
+    @property
+    def on_message_received(self):
```
BertKleewein, line 588: >on_message_received [](start = 8, length = 19)

The doc comments from sync on_message_received property also apply here.

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
```diff
@@ -0,0 +1,72 @@
...
+from azure.iot.device.common import asyncio_compat
+
+logger = logging.getLogger(__name__)
+
+loops = {
+    # Store whatever loop the user has on their thread the client was created in.
+    # We can use this to schedule tasks for their handler/callback code so that any
+    # poor performance in their provided code doesn't slow down the client.
+    "USER_LOOP": asyncio.get_event_loop(),
+    "CLIENT_INTERNAL_LOOP": None,
```
BertKleewein, line 20: You need to defer the call to `get_event_loop`. Calling this here mean that the event loop is created when this file gets imported before any customer code has a chance to run.  

Here's one bug that this introduces: We sneakily create the event loop before the user has a chance to set a custom event loop policy.  

```
from azure.iot.device import AzureIoTDeviceClient    # creates an event loop according to the default policy
import asyncio

asyncio.set_event_loop_policy(my_custom_policy)    # sets the policy after the current thread's loop has already been created

asyncio.run(my_main_coroutine)    # user expects code to run on loop created by my_custom_policy, but it's running on the default loop.
```

This kind of bug is __extremely__ hard to track down because you don't expect the unrelated `import` statement at the top of the file to change the behavior of code in such a subtly destructive way.

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
```diff
@@ -0,0 +1,72 @@
...
+def _cleanup():
+    """Clear all running loops and end respective threads.
+    Does not clear the USER_LOOP.
+    ONLY FOR TESTING USAGE
+    By using this function, you can wipe all global loops.
+    DO NOT USE THIS IN PRODUCTION CODE
+    """
+    for loop_name, loop in loops.items():
+        if loop_name == "USER_LOOP":
+            # Do not clean up the USER_LOOP since it wasn't made by us
```
BertKleewein, line 34: >since it wasn't made by us [](start = 44, length = 26)

but it was..... see above comment :)

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
```diff
@@ -0,0 +1,72 @@
...
+            # NOTE: Stopping the loop will also end the thread, because the only thing keeping
+            # the thread alive was the loop running
+            loops[loop_name] = None
+
+
+def _make_new_loop(loop_name):
+    logger.debug("Creating new event loop - {}".format(loop_name))
+    # Create the loop on a new Thread
+    new_loop = asyncio.new_event_loop()
+    loop_thread = threading.Thread(target=new_loop.run_forever)
```
BertKleewein, line 47: I think this is in the wrong order.  Instead of crating an event loop in the current thread and then running it in a second thread, I think you want to crate a second thread and then create and run the event loop inside the second thread.

I was just reading the implementations for 3 of the 4 "included" loop policies, and I wasn't able to find a any problems with doing it your way, but it just feels like you should create the loop from inside the thread that it's running in.

But -- and I don't think we're going to get this right for all possible cases -- this code is built on the notion of one loop per thread, but that is not necessarily always the case.  The documentation for event loop policies ([https://docs.python.org/3/library/asyncio-policy.html](https://docs.python.org/3/library/asyncio-policy.html)) says "A policy defines the notion of context and manages a separate event loop per context. The default policy defines context to be the current thread."  This means that 1:1 mapping between threads and event loops is the _default_ but it's not guaranteed.  The asyncio implementation talks about theoretical policies that might use different contexts:

```
class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
    """Default policy implementation for accessing the event loop.

    In this policy, each thread has its own event loop.  However, we
    only automatically create an event loop by default for the main
    thread; other threads by default have no event loop.

    Other policies may have different rules (e.g. a single global
    event loop, or automatically creating an event loop per thread, or
    using some other notion of context to which an event loop is
    associated).
    """
```

This gets into brain-twisting territory.  If there's only one event loop for all threads, then how does `new_event_loop` behave?  It might be safer to call `get_event_loop` instead because that one will either return the current one (if it exists for the current context) or create a new one (if there is no event loop for the current context), and this would work for the default case and the "one event loop shared between all threads" case.  But, if there's a singleton event loop that's shared between all threads, what happens if you call `run_forever` on it when it's already running?

I know that pytest-asyncio made some accommodations for custom event loop policies, but I don't know if they fiddle with the policy or if they're just careful about how they use asyncio because there might be a custom policy in play.  

I think I just bumped my head on one of the rocks down here in this asyncio rabbit hole.  damn.

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
cartertinney: Yeah, this is why I didn't complete the functionality - this is gonna take a bit more work.

---
In reply to: [480443287](https://github.com/Azure/azure-iot-sdk-python/pull/651#discussion_r480443287) [](ancestors = 480443287)

azure-iot-device/azure/iot/device/iothub/sync_clients.py
cartertinney: I'm not sure - technically this function could return any type of message the client receives. Right now that's only input messages, sure, but an input message IS just a message.

Also I think documentation re: details of the message should be captured on the doc for the Message object rather than here.

---
In reply to: [480408673](https://github.com/Azure/azure-iot-sdk-python/pull/651#discussion_r480408673) [](ancestors = 480408673)

azure-iot-device/azure/iot/device/iothub/aio/loop_management.py
cartertinney: A few things:

1) I don't see a benefit to creating the loop in another thread, because that just makes things more complicated. After all, a Thread simply runs a single target function until execution is completed. If we had to build a new loop in a separate thread, and then run it, that's two operations, which would mean a custom function would need to be defined in order to do those two things (+ store it in the dictionary), and I just don't see the benefit to doing that. I think simply running it in a separate thread is simpler.

2) Given the low-level manual loop creation we're doing here, I don't think loop policy should make too much of an impact. Calling `get_event_loop` will return the default loop based on policy, but we're operating at a lower level here by making our own. And since `new_event_loop()` is part of the abstract policy, it will always be implemented no matter what custom policy is in effect. It also cannot return None. I also don't know what this means for the case of a policy enforcing a singular global loop. My feeling based on reading the docs is that as long as we're manually dealing with loops ourselves, we should probably be fine, a lot of the policy stuff only comes into effect when we try to use default loops for a context, which isn't something that happens here.

---
In reply to: [480468732](https://github.com/Azure/azure-iot-sdk-python/pull/651#discussion_r480468732) [](ancestors = 480468732)

cartertinney:
./azp run Azure.azure-iot-sdk-python-dps-e2e

cartertinney:
azp run Azure.azure-iot-sdk-python-dps-e2e

olivakar (approved):

olivakar:
could you add at least 1 sample since it is going to work now ?

PR #668, opened 2020-09-15
feature(azure-iot-device): Handlers Part 4 (final)

cartertinney:
- Updated samples to use receive handlers
- Formally deprecated old receive method APIs
- .disconnect() will now allow for a graceful program exit
- Misc. other organizational fixes


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -362,6 +377,16 @@ def create_from_symmetric_key(cls, symmetric_key, hostname, device_id, **kwargs)
     def receive_message(self):
         pass
 
+    @abc.abstractmethod
```
cartertinney, line 380: Unrelated to this PR, but these abstract definitions were missing, so I added them

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -560,3 +585,7 @@ def send_message_to_output(self, message, output_name):
     @abc.abstractmethod
     def receive_message_on_input(self, input_name):
         pass
+
+    @abc.abstractmethod
```
cartertinney, line 589: Again, unrelated to this PR, but I decided to add a missing abstract definition

azure-iot-device/tests/iothub/shared_client_tests.py
```diff
@@ -433,19 +433,6 @@ def test_raises_value_error_on_sastoken_failure(self, mocker, client_class, conn
 
 
 class SharedIoTHubClientPROPERTYHandlerTests(object):
-    @pytest.fixture(autouse=True)
-    def teardown(self, client, handler_name):
```
cartertinney, line 437: Not necessary now that we are cleaning up the client in the client fixture

azure-iot-device/tests/iothub/test_sync_clients.py
```diff
@@ -192,14 +221,6 @@ def test_raises_error_on_pipeline_op_error(
         assert e_info.value.__cause__ is my_pipeline_error
 
 
-class SharedClientDisconnectEventTests(object):
```
cartertinney, line 195: Moved to shared_client_tests.py

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -173,9 +167,30 @@ def disconnect(self):
...
+        callback = EventedCallback()
+        self._mqtt_pipeline.disconnect(callback=callback)
+        handle_result(callback)
+        logger.debug("Successfully executed secondary disconnect")
+
+        # It's also possible that in the (very short) time between stopping the handlers and
+        # the second disconnect, additional items were received (e.g. C2D Message)
+        # Currently, this isn't really possible to accurately check due to a race condition.
+        # It has always been true of this client, even before handlers.
+        # TODO: Fix that
```
cartertinney, line 193: This actually may not be completely fixable on our end - may need to wait for MQTT 5 to completely resolve this issue (although there are things we can do to mitigate it in the meantime)

azure-iot-device/azure/iot/device/aio/patch_documentation.py
```diff
@@ -22,28 +22,12 @@ async def disconnect(self):
     disconnect.__doc__ = IoTHubDeviceClient_.disconnect.__doc__
     setattr(IoTHubDeviceClient_, "disconnect", disconnect)
 
-    async def get_storage_info_for_blob(self, blob_name):
```
cartertinney, line 25: The changes in this file are a result of fixing that issue where get_storage_blob() and notfiby_blob_upload_status() were maintained in the wrong spot

BertKleewein: Was this file generated?  I know we had a plan to codegen this at one point, but I don't know what the final decision was.  If this is code-generated, it should be marked as such in header comments, including instructions for re-generating it.  If this file is generated, then I assume you re-generated it to get these results?  If not, you should have.

---
In reply to: [489763278](https://github.com/Azure/azure-iot-sdk-python/pull/668#discussion_r489763278) [](ancestors = 489763278)

azure-iot-device/azure/iot/device/iothub/aio/async_inbox.py
```diff
@@ -70,6 +70,8 @@ async def get(self):
     def empty(self):
         """Returns True if the inbox is empty, False otherwise
 
+        Note that there is a race condition here, and this may not be accurate.
```
BertKleewein, line 73: >Note that there is a race condition here, and this may not be accurate. [](start = 8, length = 71)

please add some text that explains the race condition.  It doesn't need to be too explicit, but it should give people somewhere to look.  "race condition here" implies that the async inbox has a race condition, but I don't think that's the case.  I think you need to say something like "this function doesn't always return an accurate value because a message might arrive while the client is in process of disconnecting.  If the message arrives after the blorg is schmotzed, then the schenzer might shmear without mauering."  Same for the sync version

azure-iot-device/samples/async-hub-scenarios/receive_message_x509.py
```diff
@@ -27,26 +27,25 @@ async def main():
...
+    # Define behavior for receiving a message
+    # NOTE: this could be a function or a coroutine
+    def message_received_handler(message):
+        print("the data in the message received was ")
+        print(message.data)
+        print("custom properties are")
+        print(message.custom_properties)
+
+    # Set the message received handler on the client
+    device_client.on_message_received = message_received_handler
```
BertKleewein, line 39: >device_client.on_message_received = message_received_handler [](start = 4, length = 60)

I notice that you almost always set handlers _after_ you call connect, and I know that this is significant because setting a handler can cause a connection to happen.  This sample doesn't call connect, yet it still works and that's confusing.  Since many developers copy samples and build off of them, I would recommend adding a comment to each handler assignment in the samples to say something like "If the client isn't connected, this assignment will cause the client to connect." or something like that.  (Have I mentioned that I don't like having an assignment causing a network connection?  Yes, I'm a broken record....)

BertKleewein (approved):
:shipit:

olivakar (approved):
🐋

PR #684, opened 2020-09-29
fix(azure-iot-device): Fix flaky HandlerManager tests

cartertinney:
* Fixed bugs in tests due to MagicMocks not being threadsafe
* Optimized tests for better performance

This PR has passed 10 consecutive full pipeline runs with no errors, as well as 1000 consecutive runs on a local machine


azure-iot-device/tests/iothub/test_sync_handler_manager.py
```diff
@@ -278,26 +300,24 @@ def test_handler_resolve_pending_items_before_handler_removal(
...
-        # result in an inability for me to write this test in a way that would work on both
-        # versions. So we'll just test that it's less than 100.
+        # it's pretty difficult to make the timing work
         assert mock_handler.call_count < 100
+
         # Immediately remove the handler
         setattr(handler_manager, handler_name, None)
         # Wait to give a chance for the handler runner to finish calling everything
-        time.sleep(0.5)
+        time.sleep(0.2)
```
BertKleewein, line 310: >time.sleep(0.2) [](start = 8, length = 15)

<self-righteous soapboxing>Adding sleeps to a test is just begging for problems. <s/elf-righteous soapboxing> 
Even though you're waiting for things to happen in other threads, there are ways to test without using sleeps (in most cases -- testing for things that don't happen still require some time component).  

Here is one way you could do this without sleep calls:
```
class ThreadsafeMock(object):
    """ This class provides (some) Mock functionality in a threadsafe manner, specifically, it
    ensures that the 'call_count' attribute will be accurate when the mock is called from another
    thread.

    It does not cover ALL mock functionality, but more features could be added to it as necessary
    """

    def __init__(self, trigger_count=1):
        self.call_count = 0
        self.lock = threading.Lock()
        # >> NEW
        # add an event that can fire when we reach a certain number of calls (defaults to one call)
        self.trigger_count = trigger_count
        self.trigger = threading.Event()
        # << NEW
        
        # >> NEW
    def reset_trigger(trigger_count):
        # reset the trigger
        self.trigger_count = trigger_count
        self.trigger.clear()
        # << NEW

    def __call__(self, *args, **kwargs):
        with self.lock:
            self.call_count += 1
            # >> NEW
            # if we hit our call count, set the trigger event
            if self.call_count == self.trigger_count:
                self.trigger.set()
            # << NEW


    def test_handler_resolve_pending_items_before_handler_removal(
        self, mocker, handler_name, handler_manager, inbox
    ):
        # Use a threadsafe mock to ensure accurate counts
        mock_handler = ThreadsafeMock(trigger_count=100)
        assert inbox.empty()
        # Queue up a bunch of items in the inbox
        for _ in range(100):
            inbox._put(mocker.MagicMock())
        # The handler has not yet been called
        assert mock_handler.call_count == 0
        # Items are still in the inbox
        assert not inbox.empty()
        # Set the handler
        setattr(handler_manager, handler_name, mock_handler)
        # The handler has not yet been called for everything that was in the inbox
        # NOTE: I'd really like to show that the handler call count is also > 0 here, but
        # it's pretty difficult to make the timing work
        assert mock_handler.call_count < 100

        # Immediately remove the handler
        setattr(handler_manager, handler_name, None)
        # >> NEW
        # Wait for 100 calls.  A longer timeout is OK here because this will only wait as long as it takes to call 100 times.  
        mock_handler.trigger.wait(timeout=.5)
        # << NEW
        # Despite removal, handler has been called for everything that was in the inbox at the
        # time of the removal
        assert mock_handler.call_count == 100
        assert inbox.empty()

        # >> NEW
        # Reset the trigger.  Do this before we add more items.
        mock_handler.reset_trigger(101)
        # << NEW
        # Add some more items
        for _ in range(100):
            inbox._put(mocker.MagicMock())
        # Wait to give a chance for the handler to be called (it won't)
        # >> NEW
        # This is still sort-of a timer because we're testing that the handler _doesn't_ get called in .2 seconds
        # If the handler gets called in .2001 seconds, this test would pass
        mock_handler.trigger.wait(timeout=.2)
        # << NEW
        # Despite more items added to inbox, no further handler calls have been made beyond the
        # initial calls that were made when the original items were added
        assert mock_handler.call_count == 100
```

BertKleewein (approved):
:shipit:

azure-iot-device/tests/iothub/test_sync_handler_manager.py
cartertinney: This is a very good idea that I will pivot to if we end up having to revisit these tests

PR #693, opened 2020-10-16
fix(azure-iot-device): Fixed errors when using device/module ids with special characters

cartertinney:
- Removed URL encoding from module id and device id in topic strings
- Left all other URL encoding in topic strings in place for now, since this issue only really affected device and module ids. When Hub makes a new API, we will be able to return to URL encoding all values for consistency


BertKleewein (approved):
:shipit:

PR #701, opened 2020-10-30
fix(azure-iot-device): Fixed .update_sastoken() to reauthorize connection

cartertinney:
- .update_sastoken() will now issue a reauthorization on the pipeline
- because there is now I/O, on the async clients, .update_sastoken() is now a coroutine instead of a method


BertKleewein (approved):
:shipit:

PR #706, opened 2020-11-09
fix(azure-iot-device): SAS renewal is not broken by system sleep

cartertinney:
* Added a new class called 'Alarm' that will trigger an event at a specific date-time
* Modified the SasTokenRenewalStage of the pipeline to use Alarms instead of Timers


azure-iot-device/azure/iot/device/common/alarm.py
```diff
@@ -0,0 +1,36 @@
...
+        Thread.__init__(self)
+        self.alarm_time = alarm_time
+        self.function = function
+        self.args = args if args is not None else []
+        self.kwargs = kwargs if kwargs is not None else {}
+        self.finished = Event()
+
+    def cancel(self):
+        """Stop the alarm if it hasn't finished yet."""
+        self.finished.set()
```
BertKleewein, line 23: >self.finished.set() [](start = 8, length = 19)

Do you want to null out function, args, and kwargs since the function will never be called at this point?  GC _should_ clean them up, but I don't necessarily trust GC.
(note: be careful if you do this -- the easy answer has a race condition.)

azure-iot-device/azure/iot/device/common/alarm.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def run(self):
```
olivakar, line 25: >run [](start = 8, length = 3)

can u say what it does like if the current time is greater then something otherwise something else

cartertinney:
This PR is currently blocked pending #721

azure-iot-device/azure/iot/device/common/alarm.py
cartertinney: If the alarm has already expired, it immediately resolves the associated function. I will add a test to show this.

azure-iot-device/azure/iot/device/common/alarm.py
cartertinney: This pattern is copied from the other similar classes (e.g. Timer) from the builtin threading module. I'm inclined to keep the similarities since the builtins are known to be reliable.

BertKleewein (approved):
:shipit:

PR #710, opened 2020-11-18
docs(azure-iot-device): Updating samples to support running as IoT Edge modules

niswitze:
## Description of the problem

Running any sample under azure-iot-device that prompts for user input from stdin directly as modules in IoT Edge will throw an EOF error. Since these modules are executed as non-interactive containers, stdin will not be attached which causes the input method to throw an EOF error and the module to exit prematurely.

Placing a try except around the input method for the EOF error and forcing the stdin_listener method to sleep instead fixes this issue. Please let me know if there is a better way to get around this issue or if I should make any updates to this PR.


danhellem:
@cartertinney is this something we should add to the samples or should we just close this out?

PR #714, opened 2020-11-21
Em tempcontroller v2

ericmitt:
Make this sample compatible with Central (uses different models for each component)


anthonyvercolano (approved):
:shipit:

olivakar (approved):

PR #721, opened 2020-12-31
feature(azure-iot-device): Client shutdown

cartertinney:
* Added a .shutdown() method to the IoTHub clients that cleans up client resources. This method should be called at the end of client use instead of .disconnect()
* Added implicit shutdown after a successful registration via the .register() method on the Provisioning client
* Not all resources are cleaned up at this time as a result, but the most important ones are. A more robust implementation could be extended in the future, leveraging this structure
* Updated samples to reflect the correct usage
* Various associated refactoring


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -353,6 +353,19 @@ def _create_ssl_context(self):
...
+        """Shut down the transport. This is (currently) irreversible."""
+        # Remove the disconnect handler from Paho. We don't want to trigger any events in response
+        # to the shutdown and confuse the higher level layers of code. Just end it.
+        self._mqtt_client.on_disconnect = None
+        # Now disconnect and do some additional cleanup.
+        self._force_transport_disconnect_and_cleanup()
+        # Paho can't be fully shut down until it is garbage collected. So we delete the reference.
+        # Due to the GC algorithm, we don't know exactly when it will get garbage collected, and
+        # thus we don't know when it will truly be shut down. But we do know it will happen (soon).
+        del self._mqtt_client
```
BertKleewein, line 366: >del self._mqtt_client [](start = 8, length = 21)

Do you know with certainty that this is required?  All of the code in this file assumes that `self._mqtt_client` is valid.  It seems like you're taking a big risk in assuming that none of the code in this module is going to be called after `shutdown()` is called.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
```diff
@@ -167,6 +167,17 @@ def _run_op(self, op):
 
             op.complete()
 
+        elif isinstance(op, pipeline_ops_base.ShutdownPipelineOperation):
+            # TODO: Make sure the transport is disconnected. This is trickier than it seems.
```
BertKleewein, line 171: ># TODO: Make sure the transport is disconnected. This is trickier than it seems. [](start = 12, length = 80)

Aren't you already doing this at a higher level?  There's code in sync_client.py to disconnect before shutting down.

Also, what does "Make sure" mean?  It could mean a few different things.  A few that come to mind:
1. review the transport code to make sure all the shutdown code paths also cause the disconnect.
2. fail this operation if the transport is still connected.
3. cause the transport to disconnect if it's still connected
4. wait for the transport to disconnect if it's still connected. 

For what it's worth, there isn't much state management in this level and below.  e.g. if you try to disconnect, there's no code in here to verify that you're actually connected.  Likewise if you try to publish.  The "make sure the transport is in the correct state" logic happens above this.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -155,26 +156,66 @@ def _on_disconnected():
...
-
         op = pipeline_ops_base.InitializePipelineOperation(callback=callback)
-
         self._pipeline.run_op(op)
         callback.wait_for_completion()
 
+        # Set the running flag
+        self._running = True
+
+    def _ensure_running(self):
```
BertKleewein, line 173: >_ensure_running [](start = 8, length = 15)

This is different from all the other times I've seen/used "ensure" as a verb.  I would assume that "ensure running" means "if it's not running, make it run" and you're using it to mean "if it's not running, fail".  (This is the same issue as the "make sure the transport is disconnected" comment that I made elsewhere.)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -155,26 +156,66 @@ def _on_disconnected():
...
+        There is currently no way to resume pipeline functionality once shutdown has occurred.
+
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.PipelineNotRunning` if the
+            pipeline has already been shut down
+
+        The shutdown process itself is not expected to fail under any normal condition, but if it
+        does, exceptions are not "raised", but rather, returned via the "error" parameter when
+        invoking "callback".
+        """
+        self._ensure_running()
```
BertKleewein, line 194: >self._ensure_running() [](start = 8, length = 22)

I would assume that `shutdown` is idempotent.  In other words, if you call `shutdown` and you're already shutdown, nothing needs to happen and the function can complete successfully.  Other functions at this level behave this way -- if you call `connect` and you're already connected, it's not a failure.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -155,26 +156,66 @@ def _on_disconnected():
...
+        logger.debug("Commencing shutdown of pipeline")
+
+        def on_complete(op, error):
+            if not error:
+                # Only set the pipeline to not be running if the op was successful
```
BertKleewein, line 199: ># Only set the pipeline to not be running if the op was successful [](start = 16, length = 66)

So, if you call `shutdown` and it raises an exception, you can still use the pipeline because it's still running?  This doesn't make sense.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -389,13 +475,30 @@ def disable_feature(self, feature_name, callback):
...
         logger.debug("disable_feature {} called".format(feature_name))
         if feature_name not in self.feature_enabled:
             raise ValueError("Invalid feature_name")
-        self.feature_enabled[feature_name] = False
+        # TODO: What about if the feature is already disabled?
 
         def on_complete(op, error):
+            if error:
+                logger.error(
+                    "Unsubscribe for {} failed. Not disabling feature".format(feature_name)
```
BertKleewein, line 498: >Not disabling feature [](start = 48, length = 21)

I don't like this assumption.  If you fail to disable something, it doesn't (necessarily) mean that it's still enabled.  If lower code disables the feature and then raises an exception, the `feature_enabled` flag will be incorrect.  Because the code in sync_clients.py only calls `enable_feature()` if enabled==False, there's no way to recover from this error.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -363,10 +437,22 @@ def enable_feature(self, feature_name, callback):
...
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.ConnectionFailedError`
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.ConnectionDroppedError`
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.UnauthorizedError`
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.ProtocolClientError`
         """
+        self._ensure_running()
         logger.debug("enable_feature {} called".format(feature_name))
         if feature_name not in self.feature_enabled:
             raise ValueError("Invalid feature_name")
+        # TODO: What about if the feature is already enabled?
```
BertKleewein, line 455: ># TODO: What about if the feature is already enabled? [](start = 8, length = 53)

This is an architectural question.  Enabling means subscribing, which means network traffic.  We want to make sure the feature is enabled, but we don't want to subscribe every time we use the feature.  We also don't want to assume that we're enabled when we're not.  

The real question is "what layer of code is responsible for making sure we don't subscribe more than we need to?"  Currently, this happens in sync_client.py, so I would argue that we should re-enable at this layer, even if we're already enabled.

For what it's worth, I'm currently dealing with a github issue that looks like a Python but because it doesn't repro with other languages because other languages will re-subscribe under certain conditions.  Right now, it looks like a bug in edgeHub, but there's some justification for following the rule: "re-subscribe if the user asks us, even if we think that we know better."

BertKleewein: that being said, I think changing the enable/disable code is out of scope and could be it's own little Pandora's box.

---
In reply to: [552299580](https://github.com/Azure/azure-iot-sdk-python/pull/721#discussion_r552299580) [](ancestors = 552299580)

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -389,13 +475,30 @@ def disable_feature(self, feature_name, callback):
...
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.ConnectionDroppedError`
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.UnauthorizedError`
+        :raises: :class:`azure.iot.device.iothub.pipeline.exceptions.ProtocolClientError`
         """
+        self._ensure_running()
         logger.debug("disable_feature {} called".format(feature_name))
         if feature_name not in self.feature_enabled:
             raise ValueError("Invalid feature_name")
-        self.feature_enabled[feature_name] = False
+        # TODO: What about if the feature is already disabled?
```
BertKleewein, line 493: ># TODO: What about if the feature is already disabled? [](start = 8, length = 54)

My opinion from `enable_feature` above applies here as well.

BertKleewein (changes requested):
:clock1:

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Paho cleans up it's threads in the __del__ method, which is called upon garbage collection. This would automatically happen if our client went out of scope of course (with the IoTHubClient gc'd, then Paho would be as well), but if we want it to happen in response to .shutdown() it would have to be manually deleted.

The other concern is that if we wait for gc, then as long as the IoTHubClient is around (let's say shut down, but not out of scope yet), Paho is still running.

My opinion is that we should delete Paho upon shutdown. If we wanted to create some kind of way to restore the client, we would have to run an InitializePipelineOperation again anyway due to needing to reinitialize pipeline resources like SAS timers etc., and we would create a new instance of the transport as well.

But yeah, honestly the whole thing is kinda messy and has created a lot of edge cases. Perhaps there should be some additional logic added to lock down all other operations when a Shutdown is in progress?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_mqtt.py
cartertinney: This is an outdated comment. I have removed it. It referred to the connection in the Transport, but that's been resolved.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
cartertinney: I'll change this to "verfiy_running" instead, there's less room for confusion of meaning there

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
cartertinney: The idea here is that if you try to shutdown once you are already shut down, it will raise an error. While I could have just let it automatically succeed, I felt it made more sense to raise an error, so that there's a consistent behavior - once you shut down, every future call will fail.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
cartertinney: If there's a failure, the pipeline actually didn't completely shut down. Say there were five pipeline stages, and it failed on stage 3, that means stage 3, 4 and 5 have not been correctly shut down.

I implemented it this way in order to give you the ability to try the operation again so the client can't be put in a state where it is impossible to shutdown.

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
cartertinney: Good point. I'll reorganize, and issue a warning instead

azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
cartertinney: I agree. This was a little box I tripped over while doing this PR and didn't dare open, so I left this little TODO instead so it can't be forgotten about.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: I'm not worried about needing to restore the client.   My concern is more about risk.

Don't forget, you're doing this change (the whole PR) because we know that we need to clean up the sas renewal thread.  We knew that this was an example where GC wasn't working.  ie, functionality is broken.   I asked "do you know if this is required" above because it seems like you're cleaning up Paho because "it seems like a good idea" rather than "functionality is broken".  The purpose of shutdown is to "clean up things that we _must_ clean up", not to "clean up everything we can".  We're trying to work around a problem with GC, not "leave everything as clean as possible when we're done."  This is a slippery slope.  

The reason I think this is risky is because of Paho callbacks.  Every time we call into Paho, there's a chance that it will cause Paho to call back into us.  For example (I don't think this can actually happen, but it's an example): 
1. cleanup calls del self._mqtt_client.  The _mqtt_client attribute is removed from our MqttTransport object.  
2. this causes Paho to be collected
3. while being collected, Paho needs to disconnect (I'm not sure if this can happen or not, but theoretically it could.)
4. As part of paho's disconnect logic, it needs to notify us, so it calls mqtt_transport.on_disconnect
5. mqtt_transport.on_disconnect calls _force_transport_disconnect_and_cleanup.
6. _force_transport_disconnect_and_cleanup calls self._mqtt_client.disconnect().

Normally, this circular logic isn't a problem, presumably because Paho doesn't disconnect if it's already disconnected.  

But, because we deleted self._mqtt_client, calling self._mqtt_client.disconnect() raises an AttributeError.  Maybe this is harmless and maybe not.  But, we have an AttributeError that we didn't have before.  This is why I'm concerned.  When you add code to delete the attribute, I then have to assume that every use of self._mqtt_client could cause an AttributeError.  I'm just not smart enough to imagine all the twisted ways that Paho might call back into us.  If I wanted to be extra careful, I would maybe add a call to hasattr before every use of self._mqtt_client, and then slip a little further down that slippery slope.

That being said, you asserted that Paho's __del__ method cleans up its threads.  Maybe I'm missing something, but I don't see where that happens.  The __del__ logic closes sockets, but I don't see it stopping any threads.  

On top of this, `del` doesn't necessarily cause the object to be collected.  It just removes the attribute, which removes the reference, and this may or may not cause the object to be collected.  If anyone else has a reference to the _mqtt_client object (including circular references), it doesn't get cleaned up until the GC decides to clean it up.  

> as long as the IoTHubClient is around (let's say shut down, but not out of scope yet), Paho is still running.
If Paho is still running after shutdown, then there's a bug we need to fix.  _force_transport_disconnect_and_cleanup calls loop_stop, and this should stop the background thread.  Am I reading the code wrong or do you mean something else?

cartertinney: You know what, yeah, I think I'll cut this out (for now at least). I'll keep the force stop of Paho in, with the _force_transport_disconnect_and_cleanup() call, but I won't delete the object. It's trivial to add it later, but keeping it does greatly increase our risk exposure on an already fairly substantial PR.

BertKleewein (approved):
:shipit:

PR #726, opened 2021-01-19
NOT READY FOR REVIEW: replace http_client usage with urllib

BertKleewein:


BertKleewein:
Bug only repro's < 2.7.9, and we can't even run on those older builds because of breaking TLS changes.

PR #728, opened 2021-01-26
fix (azure-iot-device): Patching script adjustments for properties

cartertinney:
* Adjusted patching scripts so client properties (incl. handlers) are dynamically moved to leaf classes from parents, so that online API documentation can correctly ingest
* Adjusted patching scripts for correct ingestion of the .shutdown() method as well
* Fixes GitHub issue #725

### DIFF
--- a/azure-iot-device/azure/iot/device/patch_documentation.py
+++ b/azure-iot-device/azure/iot/device/patch_documentation.py
@@ -10,6 +10,12 @@ Currently we have to do like this so that we don't use exec anywhere"""
 def execute_patch_for_sync():
     from azure.iot.device.iothub.sync_clients import IoTHubDeviceClient as IoTHubDeviceClient
 
+    def shutdown(self):
+        return super(IoTHubDeviceClient, self).shutdown()
+
+    shutdown.__doc__ = IoTHubDeviceClient.shutdown.__doc__
+    setattr(IoTHubDeviceClient, "shutdown", shutdown)
+
     def connect(self):
         return super(IoTHubDeviceClient, self).connect()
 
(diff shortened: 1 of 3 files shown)


azure-iot-device/azure/iot/device/patch_documentation.py
```diff
@@ -216,6 +241,19 @@ def create_from_x509_certificate(cls, x509, hostname, device_id, module_id, **kw
         "create_from_x509_certificate",
         classmethod(create_from_x509_certificate),
     )
+
+    setattr(IoTHubModuleClient, "connected", IoTHubModuleClient.connected)
```
BertKleewein, line 245: >setattr(IoTHubModuleClient, "connected", IoTHubModuleClient.connected) [](start = 4, length = 70)

Why don't you copy the dunder-doc for properties like you do for functions?  This is probably worth documenting in the code.

cartertinney: So I'm doing something different here than I am for functions/methods, because these are properties, not functions/methods.

You may say, "what's the difference? They're defined as methods...", which is true syntactically, but not true behind the scenes of the object. Defining properties using the @property decorator on functions actually creates a "property" object which is a type of implementation of something called "descriptors" which are like the plumbing of Python's OO implementation. This "property" object contains all the relevant calls and data (including docstrings) within it.

So, in contrast to methods, where we dynamically define override methods for the child classes, which then call the parent class implementations of the method with the same name, here we are actually just dynamically overriding the property object inherited from the parent with the exact same property object, but put directly on the child.

tl;dr: For methods we create something new, and then copy the docstring to it. For properties we just move the whole thing to the same place it already was... but now it's tied directly to the child class, rather than implicitly so via inheritance.

BertKleewein: That's right.  It's coming back to me now.  I forgot about the new method that calls super().

I don't think we should change anything, but...

I wonder if the "move the whole thing to the place it already was" approach would work for methods too.  Well, I don't think it would work for class methods, but I think it would work for instance methods.  

I remember talking about this, and I remember you being against it, and I remember class methods being one issue, but I don't remember if there were other reasons.

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/patch_documentation.py
cartertinney: It might work, it just does feel a little weird moving things around (especially since you have to remember it's happening in case you decide to provide class-specific overrides at a later date). If I could have thought of a different way to accomplish it with properties I probably would have, rather than go this route, I just didn't see many other options.

PR #750, opened 2021-04-08
Add parentScopes and deviceScope properties

vishnureddy17:
We shouldn't merge this until we confirm API availability on the public cloud. Marked as draft to avoid merging for now.


anthonyvercolano (approved):
Well it looks good to me.

azure-iot-hub/azure/iot/hub/protocol/models/fault_injection_connection_properties.py
```diff
@@ -1,28 +0,0 @@
...
-# --------------------------------------------------------------------------
-# Code generated by Microsoft (R) AutoRest Code Generator.
-# Changes may cause incorrect behavior and will be lost if the code is
-# regenerated.
-# --------------------------------------------------------------------------
-
-from msrest.serialization import Model
-
-
-class FaultInjectionConnectionProperties(Model):
```
BertKleewein, line 11: What's up with all these deleted files?  I don't see any swagger changes that would explain this.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -74,7 +74,17 @@ def __del__(self):
...
 
-    def create_device_with_sas(self, device_id, primary_key, secondary_key, status, iot_edge=False):
+    def create_device_with_sas(
+        self,
+        device_id,
+        primary_key,
+        secondary_key,
+        status,
+        iot_edge=False,
+        status_reason=None,
```
BertKleewein, line 84: > status_reason=None [](start = 7, length = 19)

You need to update the docstrings in this file for all the parameters you added.  The docstrings get run through a tool and published for end users, so these need extra attention in terms of content and also in terms of format.  I assume it will be lots of copy/paste, but I'm not sure how you deal with parent_scopes having 2 different types.  I'm sure there's syntax for this; hopefully you can find a good example.

BertKleewein (changes requested):
:clock1:

azure-iot-hub/azure/iot/hub/protocol/models/fault_injection_connection_properties.py
vishnureddy17: Not sure. I tried generating the code both on Ubuntu 20.04 through WSL and on Windows and both produce the same result.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
vishnureddy17: Updated the docstrings. @cartertinney , can you confirm the formatting is okay?

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -192,6 +283,11 @@ def update_device_with_x509(
         :param str secondary_thumbprint: Secondary X509 thumbprint.
         :param str status: Initital state of the created device.
             (Possible values: "enabled" or "disabled").
+        :param bool iot_edge: Whether or not the created device is an IoT Edge device.
```
cartertinney, line 286: I see several of these methods have added a param for IoTEdge. Are we sure this is correct? - Edge with X509 seems odd to me, but I don't know the spec.

vishnureddy17: Edge with X509 was already there in the SDK for the _create_ functions, I just added it to the _update_ functions. And Anca Antochi from platform said that updating iotEdge is supported by the service.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -74,14 +74,29 @@ def __del__(self):
...
         """Creates a device identity on IoTHub using SAS authentication.
 
         :param str device_id: The name (Id) of the device.
         :param str primary_key: Primary authentication key.
         :param str secondary_key: Secondary authentication key.
         :param str status: Initital state of the created device.
             (Possible values: "enabled" or "disabled")
+        :param bool iot_edge: Whether or not the created device is an IoT Edge device. Default value: False
+        :param str status_reason: The reason for the device identity status. Default value: None
+        :param str device_scope: The scope of the device. Default value: None
```
drwill-ms, line 97: Do you want to copy the doc comments I crafted in .NET? My opinion was that the comments from swagger were insufficient for end-user guidance.

BertKleewein (approved):
:shipit:

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
vishnureddy17: Done, thanks!

PR #754, opened 2021-04-20
Registry etag fixes

### LINKED ISSUE #613: Unable to update device registered with X509 using method update_device_with_x509 defined in IoTHubRegistryManager
OS : Windows 10
Python version : Python 3.7.4
pip version :  19.0.3 
azure-iot-hub==2.2.1

ISSUE OCCURRED IN 
  Python module : "azure.iot.hub.IoTHubRegistryManager"
  Method Name : update_device_with_x509

## HEADING OF ISSUE
Unable to update device registered with X509 using method update_device_with_x509 defined in IoTHubRegistryManager

## Description of the issue
Using Method update_device_with_x509(self, device_id, etag, primary_thumbprint, secondary_thumbprint, status) to update already registered device (device is registered with X509 certificate).

The method is returning 'Conflict' Exception always.

## Root Cause of Issue
'Conflict' Exception is thrown because the above method was always creating device instead of updating the device.
It found device already created so throwing 'Conflict' Exception always

## Updated following in the method to solve the issue
the return statement in this method was:
return self.protocol.registry_manager.create_or_update_device(device_id, device,)

Changed it to 
return self.protocol.registry_manager.create_or_update_device(device_id, device, "*")

And its start updating the device.

AB#7613969

### LINKED ISSUE #716: Etag ignored when updating module twin.
# Context

- **OS and version used:**  Ubuntu 18.04 )
- **Python version:** 3.6.12
- **pip version:**  20.3
- **list of installed packages:** 

| Name                |    Version          |         Build   |    Channel |
| --------- | ------------------- | --------- | ---|
| _libgcc_mutex       |      0.1            |            main ||
| astroid             |      2.4.2          |          pypi_0 |    pypi|
| attrs               |      20.3.0         |          pypi_0 |    pypi|
| ca-certificates     |      2020.10.14     |               0 ||
| certifi             |      2020.11.8      |  py36h06a4308_0 ||
| dataclasses         |      0.7            |          pypi_0 |    pypi|
| importlib-metadata  |      3.1.1          |          pypi_0 |    pypi|
| iniconfig           |      1.1.1          |          pypi_0 |    pypi|
| isort               |      5.6.4          |          pypi_0 |    pypi|
| jedi                |      0.17.2         |          pypi_0 |    pypi|
| lazy-object-proxy   |      1.4.3          |          pypi_0 |    pypi|
| ld_impl_linux-64    |      2.33.1         |      h53a641e_7 ||
| libedit             |      3.1.20191231   |      h14c3975_1 ||
| libffi              |      3.3            |      he6710b0_2 ||
| libgcc-ng           |      9.1.0          |      hdf63c60_0 ||
| libstdcxx-ng        |      9.1.0          |      hdf63c60_0 ||
| mccabe              |      0.6.1          |          pypi_0 |    pypi|
| ncurses             |      6.2            |      he6710b0_1 ||
| openssl             |      1.1.1h         |      h7b6447c_0 ||
| packaging           |      20.7           |          pypi_0 |    pypi|
| parso               |      0.7.1          |          pypi_0 |    pypi|
| pip                 |      20.3           |  py36h06a4308_0 ||
| pluggy              |      0.13.1         |          pypi_0 |    pypi|
| py                  |      1.9.0          |          pypi_0 |    pypi|
| pylint              |      2.6.0          |          pypi_0 |    pypi|
| pyparsing           |      2.4.7          |          pypi_0 |    pypi|
| pytest              |      6.1.2          |          pypi_0 |    pypi|
| pytest-mock         |      3.3.1          |          pypi_0 |    pypi|
| python              |      3.6.12         |      hcff3b4d_2 ||
| readline            |      8.0            |      h7b6447c_0 ||
| setuptools          |      50.3.2         |  py36h06a4308_2 ||
| six                 |      1.15.0         |          pypi_0 |    pypi|
| sqlite              |      3.33.0         |      h62c20be_0 ||
| tk                  |      8.6.10         |      hbc83047_0 ||
| toml                |      0.10.2         |          pypi_0 |    pypi|
| typed-ast           |      1.4.1          |          pypi_0 |    pypi|
| wheel               |      0.36.0         |    pyhd3eb1b0_0 ||
| wrapt               |      1.12.1         |          pypi_0 |    pypi|
| xz                  |      5.2.5          |      h7b6447c_0 ||
| zipp                |      3.4.0          |          pypi_0 |    pypi|
| zlib                |      1.2.11         |      h7b6447c_3 ||     

## Description of the issue
It is to my understanding that the etag is a readonly element which can be used for caching and data consistency. I am seeing that the etag property does nothing at all.

Steps:
1. Grab the same twin from two separate registry managers.
2. Have the two registry managers update the same key/value pair in the twin with different values using the etag from the retrieved twins.

Outcome:
No error. Last registry manager to write is the final value.

Expected:
Second registry manager throws an error when updating the twin.

## Code sample exhibiting the issue

rm_1 = IoTHubRegistryManager("conn_string")
rm_2 = IoTHubRegistryManager("conn_string")
one = rm_1.get_module_twin("device", "module")
two = rm_2.get_module_twin("device", "module")
one.properties.desired["test"] = "one"
two.properties.desired["test"] = "two"
rm_1.update_module_twin("device", "module", one, one.etag)
rm_2.update_module_twin("device", "module", two, two.etag)

vishnureddy17:
close #716 
close #613

### DIFF
--- a/azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
+++ b/azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
@@ -22,6 +22,12 @@ from .protocol.models import (
 )
 
 
+def _ensure_quoted(etag):
+    if not isinstance(etag, str) or (len(etag) > 1 and etag[0] == '"' and etag[-1] == '"'):
+        return etag
+    return '"' + etag + '"'
+
+
 class QueryResult(object):
     """The query result.
     :param type: The query result type. Possible values include: 'unknown', 'twin', 'deviceJob', 'jobResponse', 'raw', 'enrollment', 'enrollmentGroup', 'deviceRegistration'
@@ -73,11 +79,6 @@ class IoTHubRegistryManager(object):
         Deinitializer for a Registry Manager Service client.
         """
         self.amqp_svc_client.disconnect_sync()
-
-    def __ensure_quoted(self, etag):
-        if not isinstance(etag, str) or (len(etag) > 1 and etag[0] == '"' and etag[-1] == '"'):
-            return etag
-        return '"' + etag + '"'
 
     def create_device_with_sas(
         self,
@@ -281,7 +282,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.devices.create_or_update_identity(
-            device_id, device, self.__ensure_quoted(etag)
+            device_id, device, _ensure_quoted(etag)
         )
 
     def update_device_with_x509(
@@ -342,7 +343,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.devices.create_or_update_identity(
-            device_id, device, self.__ensure_quoted(etag)
+            device_id, device, _ensure_quoted(etag)
         )
 
     def update_device_with_certificate_authority(
@@ -393,7 +394,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.devices.create_or_update_identity(
-            device_id, device, self.__ensure_quoted(etag)
+            device_id, device, _ensure_quoted(etag)
         )
 
     def get_device(self, device_id):
@@ -422,7 +423,7 @@ class IoTHubRegistryManager(object):
         if etag is None:
             etag = "*"
 
-        self.protocol.devices.delete_identity(device_id, self.__ensure_quoted(etag))
+        self.protocol.devices.delete_identity(device_id, _ensure_quoted(etag))
 
     def create_module_with_sas(self, device_id, module_id, managed_by, primary_key, secondary_key):
         """Creates a module identity for a device on IoTHub using SAS authentication.
@@ -535,7 +536,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.modules.create_or_update_identity(
-            device_id, module_id, module, self.__ensure_quoted(etag)
+            device_id, module_id, module, _ensure_quoted(etag)
         )
 
     def update_module_with_x509(
@@ -573,7 +574,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.modules.create_or_update_identity(
-            device_id, module_id, module, self.__ensure_quoted(etag)
+            device_id, module_id, module, _ensure_quoted(etag)
         )
 
     def update_module_with_certificate_authority(self, device_id, module_id, managed_by, etag):
@@ -602,7 +603,7 @@ class IoTHubRegistryManager(object):
             etag = "*"
 
         return self.protocol.modules.create_or_update_identity(
-            device_id, module_id, module, self.__ensure_quoted(etag)
+            device_id, module_id, module, _ensure_quoted(etag)
         )
 
     def get_module(self, device_id, module_id):
@@ -645,7 +646,7 @@ class IoTHubRegistryManager(object):
         if etag is None:
             etag = "*"
 
-        self.protocol.modules.delete_identity(device_id, module_id, self.__ensure_quoted(etag))
+        self.protocol.modules.delete_identity(device_id, module_id, _ensure_quoted(etag))
 
     def get_service_statistics(self):
         """Retrieves the IoTHub service statistics.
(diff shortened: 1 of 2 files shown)


azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -74,6 +74,11 @@ def __del__(self):
         """
         self.amqp_svc_client.disconnect_sync()
 
+    def __ensure_quoted(self, etag):
+        if not isinstance(etag, str) or (len(etag) > 1 and etag[0] == '"' and etag[-1] == '"'):
+            return etag
+        return '"' + etag + '"'
+
```
vishnureddy17, line 81: The service expects etags to be wrapped in double quotes. To avoid breaking users who have figured this out, I only wrap etags that haven't already been wrapped by the user. A [similar fix was made in Node](https://github.com/Azure/azure-iot-sdk-node/pull/613).

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -264,7 +269,6 @@ def update_device_with_sas(
         kwargs = {
             "device_id": device_id,
             "status": status,
-            "etag": etag,
```
vishnureddy17, line 267: etags have been removed from any request bodies. From my testing, putting etags in the body have no effect and they should only be in headers. I assume the etags were added to the request body due to confusing documentation [like this](https://docs.microsoft.com/en-us/rest/api/iothub/service/devices/createorupdateidentity) which shows etags as part of the request body. It seems like the reason why that property shows up in the documentation is because it was in a schema in the swagger.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -273,7 +277,12 @@ def update_device_with_sas(
         }
         device = Device(**kwargs)
 
-        return self.protocol.devices.create_or_update_identity(device_id, device, "*")
+        if etag is None:
+            etag = "*"
+
+        return self.protocol.devices.create_or_update_identity(
+            device_id, device, self.__ensure_quoted(etag)
+        )
```
vishnureddy17, line 285: I add etags to the headers wherever they are missing.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -718,20 +748,26 @@ def get_twin(self, device_id):
         """
         return self.protocol.devices.get_twin(device_id)
 
-    def replace_twin(self, device_id, device_twin):
+    def replace_twin(self, device_id, device_twin, etag=None):
```
vishnureddy17, line 751: The API supports etags, but we did not expose it to the user before for some reason.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -74,6 +74,11 @@ def __del__(self):
         """
         self.amqp_svc_client.disconnect_sync()
 
+    def __ensure_quoted(self, etag):
```
cartertinney, line 77: Don't do double underscores like this in Python. Many people will say this forces the function or method to be private (as opposed to merely convention-private with a single underscore), but that's incorrect. Python has no concept of private. This use of double underscore instead enables name-mangling, which can appear as making something private, but is actually more complex, and should be avoided unless you really need to be name-mangling.

A single underscore to make this convention-private should suffice.

Also, since this method doesn't use the instance of the object, it probably shouldn't be an instance method. You could make this a classmethod instead, or better yet (IMO), probably just a utility helper function defined outside of the class

cartertinney (changes requested):

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
vishnureddy17: Done

cartertinney (approved):

anthonyvercolano (approved):

anthonyvercolano (approved):

PR #755, opened 2021-04-26
Update job manager sample for user assigned identities

vishnureddy17:
Only changing a sample here. No need to change any implementation for user assigned identities. We got this for free when regenerating the code for parentScopes.


BertKleewein (approved):
:shipit:

cartertinney (approved):

PR #756, opened 2021-04-27
feature(azure-iot-device): DisconnectOperation now cancels in-flight operations in MQTTTransport

cartertinney:
* DisconnectOperations now will clear other pending operations from the OperationManager in the MQTTTransport
* The cleared operations will be completed with an OperationCancelled exception
* This does NOT change ReauthorizeConnectionOperation behavior

### DIFF
--- a/azure-iot-device/azure/iot/device/common/mqtt_transport.py
+++ b/azure-iot-device/azure/iot/device/common/mqtt_transport.py
@@ -436,7 +436,7 @@ class MQTTTransport(object):
             raise _create_error_from_rc_code(rc)
         self._mqtt_client.loop_start()
 
-    def disconnect(self):
+    def disconnect(self, clear_pending=False):
         """
         Disconnect from the MQTT broker.
 
@@ -464,9 +464,12 @@ class MQTTTransport(object):
             err = _create_error_from_rc_code(rc)
             raise err
         else:
-            # Clear all pending items in operation manager
-            # TODO: should this be here, or in on_disconnect handler
-            self._op_manager.cancel_all_operations()
+            # Clear pending ops if instructed, but only if the disconnect was successful.
+            # Technically the disconnect could still fail upon response, however that would then
+            # cause a force disconnect via the on_disconnect handler, thus it is safe to clear
+            # ops here and now.
+            if clear_pending:
+                self._op_manager.cancel_all_operations()
 
     def subscribe(self, topic, qos=1, callback=None):
         """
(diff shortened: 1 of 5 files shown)


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -463,6 +463,13 @@ def disconnect(self):
             # to this error.
             err = _create_error_from_rc_code(rc)
             raise err
+        else:
+            # Clear pending ops if instructed, but only if the disconnect was successful.
```
BertKleewein, line 467: >but only if the disconnect was successful. [](http://example.com/codeflow?start=47&length=42)

I don't agree with the "don't cancel ops if disconnect fails" behavior.

This is because I have an ICM where an internal Paho socket is getting broken and it's causing mqtt_c;lient.disconnect to fail.  

We need to stop thinking about "disconnect failed" as "we don't know thte state.  We might still be connected" and instead treat it as "Paho is completely messed up and we need to clean up as best as possible."

This ICM goes beyond the change that you're making since the `except Exception as e` line above (448) is where the problem in my ICM manifests.  I'm only mentioning it 
so we can start changing our thinking.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -654,3 +661,34 @@ def complete_operation(self, mid):
...
                 logger.debug("No callback set for MID: {}".format(mid))
+
+    def cancel_all_operations(self):
+        """Complete all pending operations with cancellation, removing MID tracking"""
+        logger.debug("Cancelling all pending operations")
+        with self._lock:
+            # Clear pending operations
+            pending_ops = [
+                (mid, callback) for (mid, callback) in self._pending_operation_callbacks.items()
+            ]
```
BertKleewein, line 672: nit: could also do `pending_op_mids = list(self._pending_operation_callbacks.keys())`

Ths list() constricturo is necessary to convert the `dict_keys` iterator into a proper list.  This is annoying.

BertKleewein: Ah, never mind.  You need the callbacks below.  You could still remove the comprehnsion using the list() construcor -- I would have never expected the [] operator to behave differently than list().

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/common/mqtt_transport.py
cartertinney: Okay, I'll leave this for now, but this is good to consider.

PR #760, opened 2021-04-30
Update readme for PnP samples

dominicbetts:
These updates are part of the changes needed to onboard these samples to the Azure Samples browser experience.

### DIFF
--- a/azure-iot-device/samples/pnp/README.md
+++ b/azure-iot-device/samples/pnp/README.md
@@ -1,14 +1,38 @@
-# Samples to demonstrate Azure IoT Plug and Play
+---
+page_type: sample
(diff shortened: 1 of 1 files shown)


azure-iot-device/samples/pnp/README.md
```diff
@@ -1,14 +1,38 @@
-# Samples to demonstrate Azure IoT Plug and Play
+---
+page_type: sample
+description: "A set of Python samples that show how a device that uses the IoT Plug and Play conventions interacts with either IoT Hub or IoT Central."
+languages:
+- python
+products:
+- azure-iot-hub
```
cartertinney, line 7: This could be confusing since they are formatted as package names. `azure-iot-hub` is a real package in Python, the other two are not. I assume however, this is not referring to specific packages?

cartertinney (approved):

PR #769, opened 2021-05-14
update unit tests for parentscopes feature

vishnureddy17:


YoDaMa (approved):
lgtm

azure-iot-hub/tests/test_iothub_registry_manager.py
```diff
@@ -470,12 +524,24 @@ def test_initializes_device_with_kwargs_for_certificate_authority(
         self, mock_device_constructor, iothub_registry_manager
     ):
         iothub_registry_manager.update_device_with_certificate_authority(
-            device_id=fake_device_id, status=fake_status, etag=fake_etag
+            device_id=fake_device_id,
```
cartertinney, line 527: Probably should be changing the description of the test in the pytest.mark.it since it's tightly coupled to only a handful of the items now being tested. Perhaps something like "Initializes device with the provided parameters and a CA auth" in this specific case

This issue is here and everywhere through this test file

vishnureddy17: done

PR #774, opened 2021-05-21
feature(azure-iot-device): Connection State Change Handler

cartertinney:
* Refactored HandlerManagers to differentiate between the existing handlers for received data (now called Receiver Handlers) and the new Client Event Handlers
* Added infrastructure to support Client Event Handlers
* Implemented the `.on_connection_state_change` handler
* Contains partial implementation of some infrastructure for future handlers which was used for development testing

NB: Unittests should be failing right now. I have not written tests yet, and significant changes have been made to handler flow. I want to get some initial feedback on the design before I commit to tests.


azure-iot-device/azure/iot/device/iothub/client_event.py
```diff
@@ -0,0 +1,15 @@
...
+# license information.
+# --------------------------------------------------------------------------
+
+NEW_SASTOKEN_REQUIRED = "NEW_SASTOKEN_REQUIRED"
+CONNECTION_STATE_CHANGE = "CONNECTION_STATE_CHANGE"
+BACKGROUND_EXCEPTION = "BACKGROUND_EXCEPTION"
+
+
+class ClientEvent(object):
+    def __init__(self, name, values_for_user=[]):
```
BertKleewein, line 13: >values_for_user [](http://example.com/codeflow?start=29&length=15)

not sure what this is.  It needs a better name.  If these are args that get passed to the user, maybe call it args or args_for_event_callback or something like that.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -403,6 +416,30 @@ def on_method_request_received(self):
...
+    def on_connection_state_change(self):
+        return self._handler_manager.on_connection_state_change
+
+    @on_connection_state_change.setter
+    def on_connection_state_change(self, value):
+        self._handler_manager.on_connection_state_change = value
+
+    # @property
+    # def on_new_sastoken_required(self):
+    #     return self._handler_manager.on_new_sastoken_required
```
BertKleewein, line 429: Just curious, what's the plan for these handlers?

azure-iot-device/azure/iot/device/iothub/client_event.py
```diff
@@ -0,0 +1,15 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+NEW_SASTOKEN_REQUIRED = "NEW_SASTOKEN_REQUIRED"
+CONNECTION_STATE_CHANGE = "CONNECTION_STATE_CHANGE"
+BACKGROUND_EXCEPTION = "BACKGROUND_EXCEPTION"
+
+
+class ClientEvent(object):
```
BertKleewein, line 12: >ClientEvent [](http://example.com/codeflow?start=6&length=11)

doc please, especially if users receive one of these. (I'm not through the code far enough to know how these flow yet)

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
cartertinney: Add them in the next PR(s). I built some of their infrastructure in this PR because I was testing faked versions of them at client level to ensure that the infrastructure could handle these new "client event" handlers working at the same time.

In order to implement these, I'd just uncomment them at client level (where they're already functional), and then just hook them up to a trigger that comes out of the pipeline when the event occurs.

azure-iot-device/azure/iot/device/iothub/client_event.py
cartertinney: They don't, but that's a good idea to document.

These are essentially just glorified tuples matching an event name with any associated values that may exist for the event and need to be given to the user.

Since they are created and used in different modules, it felt weird to just use tuples with semantics that were a mere convention. I figured a little structure would be a good idea for readability and maintenance going forth.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -88,36 +66,138 @@ def _handler_callback(future):
...
-        # Schedule a coroutine with the correct type of handler runner
-        inbox = self._get_inbox_for_handler(handler_name)
-        if inbox:
-            coro = self._inbox_handler_runner(inbox, handler_name)
-        else:
-            coro = self._event_handler_runner(handler_name)
+    def _start_handler_runner(self, handler_name):
+        """Create, and store a task for running a handler"""
         # Run the handler runner on a dedicated event loop for handler runners so as to be
         # isolated from all other client activities
```
BertKleewein, line 121: The behavior behind this specific text ("so as to be isolated from all other client activities") bothers me.  Having a dedicated event loop here violates some of the implied contract of async programming.  Our code has been like this for a while, so I don't think we should change it right now, but I think it is a problem.

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
```diff
@@ -88,36 +66,138 @@ def _handler_callback(future):
                 fut = tpe.submit(handler, handler_arg)
                 fut.add_done_callback(_handler_callback)
 
-    async def _event_handler_runner(self, handler_name):
-        # TODO: implement
-        logger.error("._event_handler_runner() not yet implemented")
-
-    def _start_handler_runner(self, handler_name):
-        """Create, and store a task for running a handler
+    async def _client_event_handler_runner(self):
```
BertKleewein, line 69: >_client_event_handler_runner [](http://example.com/codeflow?start=14&length=28)

Not a comment on the quality of your code or this change in particular, but I can't get over the thought that maybe we over-designed this part of the code (the handler managers and runners).  We did have legacy receive APIs to contend with, but I feel like this _should_ be simpler.  I couldn't say how though  It's just a gut feel.  (No change requested.  Just pontificating.)

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/iothub/aio/async_handler_manager.py
cartertinney: I agree - there's some things we might be able to improve in the future that I have ideas on as well.

azure-iot-device/tests/iothub/test_sync_handler_manager.py
```diff
@@ -359,20 +463,179 @@ def handler(arg):
...
+            inbox.put(event)
+        time.sleep(0.2)
+
+        # Handler has been called 5 times
+        assert mock_handler.call_count == 5
+
+    @pytest.mark.it(
+        "Sends a HandlerManagerException to the background exception handler if any exception is raised during its invocation"
+    )
+    def test_exception_in_handler(
```
cartertinney, line 591: This functionality will of course have to change once the background exception handler is actually implemented

PR #782, opened 2021-06-02
feat(azure-iot-hub): support aad auth for amqp operation

vishnureddy17:


azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -56,7 +56,7 @@ class IoTHubRegistryManager(object):
     based on top of the auto generated IotHub REST APIs
     """
 
-    def __init__(self, connection_string=None, host=None, auth=None):
+    def __init__(self, connection_string=None, host=None, token_credential=None):
```
vishnureddy17, line 59: This change would break users who are using AAD auth and use the `IoTHubRegistryManager` constructor directly. I think this is fine because:

1. The AAD feature has not been out for very long, so it would only affect very early adopters of this feature
2. Users should not be using the constructor directly anyway, they should be using the `from_token_credential()` function. It's not even that easy to do this by mistake because it would require the user to manually pass in an `AzureIdentityCredentialAdapter` instance. 

Thoughts? @BertKleewein

BertKleewein: I agree.   Since these parameters have only just been added and hubs that support them aren't even fully online, it's safe to assume that nobody uses them.  I would add 2 things to the convesrsation:

1. This does not break existing customers who use the IoTHubRegistryManager constructor with a connection string since the name and position of this argument hasn't changed.

2. We should also document that customers shouldn't call this directly and should use a factory method instead.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -45,13 +52,36 @@ def _build_amqp_endpoint(self, hostname, shared_access_key_name, shared_access_k
...
-        target = "amqps://" + self.endpoint + operation
-        self.amqp_client = uamqp.SendClient(target)
+    def __init__(
+        self,
+        hostname,
+        shared_access_key_name=None,
+        shared_access_key=None,
+        token_credential=None,
+        token_scope="https://iothubs.azure.net/.default",
+    ):
```
BertKleewein, line 62: The way you have these parameters and the mutual exclusion between them, you should consider making a base class with 2 child classes -- like `IoTHubAmqpClientBase` with `IoTHubAmqpClientConnectionString` and `IoTHubAmqpClientTokenAuth` or something like that.

I don't think you _must_ change it, but using 2 derived classes just feels a little warmer and a little fuzzier to me.

vishnureddy17: Yeah, that does sound better. I'll look into it.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
vishnureddy17: It seems like most of our samples call the constructor directly instead of using a factory method. Maybe we should fix that.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
vishnureddy17: Done

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -100,3 +74,57 @@ def send_message_to_device(self, device_id, message, app_props):
...
+
+        auth = uamqp.authentication.JWTTokenAuth(
+            audience=token_scope,
+            uri="https://" + hostname,
+            get_token=get_token,
+            token_type=b"bearer",
+        )
+        auth.update_token()
+        target = "amqps://" + hostname + "/messages/devicebound"
+        self.amqp_client = uamqp.SendClient(target=target, auth=auth)
```
anthonyvercolano, line 130: Just so we are clear, I doesn't look like to me that this is doing any auto renewal here.  Let me know if I am wrong.  I will let others perform the actual approval, but it lgtm.

vishnureddy17: uamqp should be doing the refresh for us.

BertKleewein: How is that possible?  The user is passing in a token with an expiration time.  How is uamqp renewing it?

vishnureddy17: We pass in the `get_token` function that uamqp can call. My understanding from looking at the uamqp source code is that `get_token` will be called by uamqp when a renewal is needed.

azure-iot-hub/tests/test_iothub_amqp_client.py
```diff
@@ -82,10 +95,69 @@ def test_raise_exception_on_send_fail(self, mocker, mock_uamqp_SendClient):
...
         )
         amqp_client_obj = mock_uamqp_SendClient.return_value
         iothub_amqp_client.disconnect_sync()
 
         assert amqp_client_obj.close.call_count == 1
+
+
+@pytest.mark.describe("IoTHubAmqpClientTokenAuth - Amqp Client Connections")
+class TestIoTHubAmqpClientTokenAuth(object):
+    @pytest.mark.it("Get Token From TokenCredential Object")
```
cartertinney, line 109: These should be written as if the sentence begins with "it", i.e. "Gets token from TokenCredential object".

There's a whole theory we have to writing tests where the description for a testcase is a sentence fragment, in indicative-mood third-person present tense, where the subject is implied as the noun defined with pytest.mark.describe. Which sounds complicated, but like I said, it really just means write the test as if the sentence begins with "it".

tl;dr is if you can't write a testname with that grammatical structure, it's a red flag that the test is poorly defined. 

A problem here of course is that the nouns in the pytest.mark.describes are also kind of flawed, but again, that's how they've always been. So IDK, you can probably also just leave all this stuff alone for now so this doesn't turn into a complete refactor of the testfile.

Nontheless, ping me sometime and we can go over some of these conventions for future reference.

cartertinney (commented):
Looks generally good, waiting to see what happens with the renewal discussion

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -113,16 +119,15 @@ def from_token_credential(cls, url, token_credential):
 
         :rtype: :class:`azure.iot.hub.IoTHubRegistryManager`
         """
-        host = url
-        auth = AzureIdentityCredentialAdapter(token_credential)
-        return cls(host=host, auth=auth)
+        # host = url
```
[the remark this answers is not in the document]
vishnureddy17: Yep, removed.

azure-iot-hub/azure/iot/hub/__init__.py
```diff
@@ -7,14 +7,20 @@
...
+)
 from .digital_twin_client import DigitalTwinClient
 
 __all__ = [
     "IoTHubRegistryManager",
     "IoTHubConfigurationManager",
     "IoTHubJobManager",
     "IoTHubHttpRuntimeManager",
-    "IoTHubAmqpClient",
+    "IoTHubAmqpClientTokenAuth",
```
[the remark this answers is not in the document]
vishnureddy17: Correct, though we did expose the Amqp class to the user before. Do you think we should remove it?

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
vishnureddy17: Adam Ling, who works on uamqp, says that the token is refreshed for us.

azure-iot-hub/tests/test_iothub_amqp_client.py
vishnureddy17: Yeah, the test file had these issues before, and I simply stuck with the (flawed) format that was already there. I'll probably stick with it the way it is.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
```diff
@@ -113,16 +119,15 @@ def from_token_credential(cls, url, token_credential):
...
+        # auth = AzureIdentityCredentialAdapter(token_credential)
+        return cls(host=url, token_credential=token_credential)
 
     def __del__(self):
         """
         Deinitializer for a Registry Manager Service client.
         """
-        if self.amqp_svc_client:
-            self.amqp_svc_client.disconnect_sync()
+        self.amqp_svc_client.disconnect_sync()
```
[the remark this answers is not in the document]
vishnureddy17: Yeah, probably best not to touch it.

azure-iot-hub/azure/iot/hub/__init__.py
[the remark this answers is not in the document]
cartertinney: Yes, although should probably run it by the others first just to make 100% sure it's safe to do.

azure-iot-hub/azure/iot/hub/iothub_registry_manager.py
vishnureddy17: Done

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -100,3 +74,57 @@ def send_message_to_device(self, device_id, message, app_props):
...
+        hub_name = hostname.split(".")[0]
+        endpoint = "{}@sas.root.{}".format(shared_access_key_name, hub_name)
+        endpoint = quote_plus(endpoint)
+        sas_token = self._generate_auth_token(
+            hostname, shared_access_key_name, shared_access_key + "="
+        )
+        endpoint = endpoint + ":{}@{}".format(quote_plus(sas_token), hostname)
+        return endpoint
+
+    def __init__(self, hostname, shared_access_key_name, shared_access_key):
```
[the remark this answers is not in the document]
vishnureddy17: Fixed.

azure-iot-hub/azure/iot/hub/__init__.py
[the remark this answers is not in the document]
vishnureddy17: Done

azure-iot-hub/tests/test_iothub_amqp_client.py
```diff
@@ -34,58 +23,109 @@
...
-        iothub_amqp_client = IoTHubAmqpClient(
-            fake_hostname, fake_shared_access_key_name, fake_shared_access_key
-        )
-        iothub_amqp_client.send_message_to_device(fake_device_id, fake_message, fake_app_prop)
+    @pytest.mark.it("Sends messages with application properties using the uamqp SendClient")
+    def test_send_message_to_device_app_prop(self, client, mock_uamqp_SendClient):
+        client.send_message_to_device(fake_device_id, fake_message, fake_app_prop)
         amqp_client_obj = mock_uamqp_SendClient.return_value
 
         assert amqp_client_obj.queue_message.call_count == 1
```
cartertinney, line 45: Also, checking that the mocked methods were called itself doesn't mean much unless you show that the specific message was also passed in, and that the properties were set on that message. This test doesn't actually really test anything other than that you can pass in a message and some properties - it doesn't tell you what happens to them.

For this test you'd probably want something like the following asserts:

```python
# Message was queued
assert amqp_client_obj.queue_message.call_count == 1
msg_obj = amqp_client_obj.queue_message.call_args[0][0]
# Message was configured with properties and destination
assert msg_obj.message_content = fake_message    # .message_content might be the wrong name for this data, I don't know uamqp offhand
assert msg_obj.properties.to == "/devices/{}/messages/devicebound".format(fake_device_id)   # again, not sure where this is actually stored
assert msg_obj.application_properties == fake_app_prop
# Message was sent
assert amqp_client_obj.send_all_messages.call_count == 1
assert amqp_client_obj.send_all_messages.call_args == mocker.call(close_on_done=False)
```

You miiiiiiight even want to take it a bit farther and show that the message send doesn't happen until AFTER the message is queued to keep things really airtight

azure-iot-hub/tests/test_iothub_amqp_client.py
```diff
@@ -34,58 +23,109 @@
...
         assert amqp_client_obj.send_all_messages.call_count == 1
 
-    @pytest.mark.it("Send Message To Device system prop")
-    def test_send_message_to_device_sys_props(self, mocker, mock_uamqp_SendClient):
-        iothub_amqp_client = IoTHubAmqpClient(
-            fake_hostname, fake_shared_access_key_name, fake_shared_access_key
-        )
-        iothub_amqp_client.send_message_to_device(fake_device_id, fake_message, fake_sys_prop)
+    @pytest.mark.it("Sends messages with system properties using the uamqp SendClient")
```
cartertinney, line 48: As above, there are also some additional assertions you should be making here. But also you probably should have another test case for messages that don't have any properties provided (i.e. `{}`), since that's optional.

azure-iot-hub/tests/test_iothub_amqp_client.py
```diff
@@ -34,58 +23,109 @@
...
+        return {
+            "hostname": fake_hostname,
+            "shared_access_key_name": fake_shared_access_key_name,
+            "shared_access_key": fake_shared_access_key,
+        }
+
+    @pytest.fixture
+    def client(self, required_kwargs):
+        return IoTHubAmqpClientSharedAccessKeyAuth(**required_kwargs)
+
```
cartertinney, line 90: There are no instantiation tests for either of these clients - no validation of any of the auth that makes the classes unique

cartertinney (changes requested):

BertKleewein (approved):
:shipit:

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -100,3 +74,57 @@ def send_message_to_device(self, device_id, message, app_props):
...
+class IoTHubAmqpClientSharedAccessKeyAuth(IoTHubAmqpClientBase):
+    def __init__(self, hostname, shared_access_key_name, shared_access_key):
+        self.endpoint = self._build_amqp_endpoint(
+            hostname, shared_access_key_name, shared_access_key
+        )
+        operation = "/messages/devicebound"
+        target = "amqps://" + self.endpoint + operation
+        self.amqp_client = uamqp.SendClient(target)
+
+    def _generate_auth_token(self, uri, sas_name, sas_value):
```
cartertinney, line 88: Could this not just leverage the sastoken object from `azure.iot.hub.sastoken`?

Looking at it now, it seems like the logic is the same. If you leveraged it, it would reduce duplication of code + make the tests easier to write.

cartertinney (approved):

PR #785, opened 2021-06-08
DO NOT MERGE: PNP prototype spike

BertKleewein:


azure-iot-device/azure/iot/device/iothub/models/properties.py
```diff
@@ -0,0 +1,65 @@
...
+        super(WritableProperties, self).__init(**kwargs)
+        self.version = 0
+
+    @classmethod
+    def from_dict(cls, src_dict):
+        # call base class and also populate versionA
+        pass
+
+
+value_types = [int, str, float, WritablePropertyResponse]
```
BertKleewein, line 64: also boolean?  Any other types here?

cartertinney: This might get tricky in Python2.7 because `str` doesn't mean the same thing

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -440,6 +442,41 @@ def on_connection_state_change(self, value):
     # def on_background_exception(self, value):
     #     self._handler_manager.on_background_exception = value
 
+    @abc.abstractmethod
+    def send_telemetry(self, telemetry_dict, component_name=None):
+        # (dict, str) -> None
+        pass
+
+    @abc.abstractproperty
+    def on_command_received(self):
```
cartertinney, line 451: I don't see an implementation of these properties anywhere

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -57,6 +58,7 @@ def _get_config_kwargs(**kwargs):
         "auto_connect",
         "connection_retry",
         "connection_retry_interval",
+        "model_id",
```
cartertinney, line 61: is this truly a "config kwarg"? Stuff like device identification doesn't go in the pipeline config generally

PR #786, opened 2021-06-10
[DON'T REVIEW] fix: support US Government cloud for AAD authentication

vishnureddy17:


vishnureddy17:
Closing because a different token scope is no longer required for US Government cloud. Keeping the branch around in case custom token scopes need to be added in the future.

PR #788, opened 2021-06-16
test(azure-iot-hub): add from_connection_string() and from_token_credential() unit tests for service clients

vishnureddy17:


azure-iot-hub/tests/test_digital_twin_client.py
```diff
@@ -7,8 +7,8 @@
 import pytest
 from azure.iot.hub.protocol.models import AuthenticationMechanism
 from azure.iot.hub.digital_twin_client import DigitalTwinClient
-from azure.iot.hub.auth import ConnectionStringAuthentication
-from azure.iot.hub.protocol.iot_hub_gateway_service_ap_is import IotHubGatewayServiceAPIs
+import azure.iot.hub.auth
```
cartertinney, line 10: It's probably best to import objects directly in test modules (as opposed to implementations) in order to make tests a bit more readable, and since that's how other objects in this test file already work.

Unlike with the implementation, there's no benefit to doing one over the other in temrs of mocking, so just do something consistent.

azure-iot-hub/tests/test_digital_twin_client.py
```diff
@@ -7,8 +7,8 @@
...
+import azure.iot.hub.protocol.iot_hub_gateway_service_ap_is
```
cartertinney, line 11: It's probably best to import objects directly in test modules (as opposed to implementations) in order to make tests a bit more readable, especially since that's how other objects in this test file already work.

Unlike with the implementation, there's no benefit to doing one over the other in temrs of mocking, so just do something consistent.

And certainly don't import a full path like this, if you were really tied to importing the module for whatever reason you'd do

```python
from azure.iot.hub import auth
from azure.iot.hub.protocol import iot_hub_gateway_service_ap_is
```

so that you don't have to write out the full module location every time.

azure-iot-hub/azure/iot/hub/digital_twin_client.py
```diff
@@ -3,8 +3,8 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-from .auth import ConnectionStringAuthentication, AzureIdentityCredentialAdapter
-from .protocol.iot_hub_gateway_service_ap_is import IotHubGatewayServiceAPIs as protocol_client
+from . import auth as hub_auth
+from .protocol import iot_hub_gateway_service_ap_is as protocol_client
```
cartertinney, line 7: This actually changes the semantics of the import. I probably wouldn't change the way IotHubGatewayServiceAPIs is imported. And if I had to, I probably wouldn't keep rolling with "protocol_client" as an alias, as you are now importing a module rather than a class

azure-iot-hub/tests/test_digital_twin_client.py
```diff
@@ -107,6 +112,63 @@ def test_instantiates_with_connection_string_no_shared_access_key(self):
...
+    @pytest.mark.it(
+        "Creates an instance of ConnectionStringAuthentication and passes it to IotHubGatewayServiceAPIs constructor"
+    )
+    def test_connection_string_auth(self, mocker):
+        connection_string_auth_init_mock = mocker.patch.object(
+            azure.iot.hub.auth, "ConnectionStringAuthentication"
+        )
+        connection_string_auth_mock = connection_string_auth_init_mock.return_value
+        connection_string_auth_mock.__getitem__.return_value = fake_hostname
+        protocol_client_init_mock = mocker.patch.object(
```
cartertinney, line 126: I see this mock is probably why you changed the import in the implementation - but is this necessary? I see it's helpful for allowing you to check the params of the instantiation, but could you not just check the value that gets set on the IotHubGatewayServiceAPIs object without mocking?

In general, only mock when necessary. Mock things that are sufficiently complex you can't (or don't want to) instantiate them. Mock things that are strictly necessary for you to demonstrate the requirement in the `pytest.mark.it` statement. Otherwise, mocking is best to be avoided lest you make things more complicated than they need to be.

(I could be wrong about this specific case, maybe it is necessary, I'm just speaking in general here)

azure-iot-hub/tests/test_digital_twin_client.py
```diff
@@ -107,6 +112,63 @@ def test_instantiates_with_connection_string_no_shared_access_key(self):
...
+        "Creates an instance of AzureIdentityCredentialAdapter and passes it to IotHubGatewayServiceAPIs constructor"
+    )
+    def test_token_credential_auth(self, mocker):
+        azure_identity_credential_adapter_init_mock = mocker.patch.object(
+            azure.iot.hub.auth, "AzureIdentityCredentialAdapter"
+        )
+        azure_identity_credential_adapter_mock = (
+            azure_identity_credential_adapter_init_mock.return_value
+        )
+        mock_azure_identity_TokenCredential = mocker.Mock()
```
cartertinney, line 157: Use `mocker.MagicMock` not `mocker.Mock`. MagicMocks are just better.

azure-iot-hub/tests/test_digital_twin_client.py
```diff
@@ -107,6 +112,63 @@ def test_instantiates_with_connection_string_no_shared_access_key(self):
...
+        protocol_client_init_mock = mocker.patch.object(
+            azure.iot.hub.protocol.iot_hub_gateway_service_ap_is, "IotHubGatewayServiceAPIs"
+        )
+
+        DigitalTwinClient.from_token_credential(fake_hostname, mock_azure_identity_TokenCredential)
+
+        assert azure_identity_credential_adapter_init_mock.call_args == mocker.call(
+            mock_azure_identity_TokenCredential
+        )
+        assert protocol_client_init_mock.call_args == mocker.call(
```
cartertinney, line 167: I had said above that perhaps you could avoid mocking this out, but if it turns out you can't, you may want to consider creating a fixture that will always mock it out so you don't have to do it manually.

azure-iot-hub/tests/test_iothub_configuration_manager.py
```diff
@@ -51,6 +53,65 @@ def iothub_configuration_manager():
     return iothub_configuration_manager
 
 
+@pytest.mark.describe("IoTHubConfigurationManager - .from_connection_string()")
+class TestFromConnectionString(object):
```
cartertinney, line 57: Tests for bad connection string values seem to be missing.

azure-iot-hub/tests/test_iothub_configuration_manager.py
```diff
@@ -51,6 +53,65 @@ def iothub_configuration_manager():
...
+    @pytest.mark.it(
+        "Creates an instance of ConnectionStringAuthentication and passes it to IotHubGatewayServiceAPIs constructor"
```
cartertinney, line 59: Are there multiple types of valid connection strings? i.e. connection strings with optional values? Different formats? If so, this test should be parametrized to show that all valid types of strings work.

azure-iot-hub/tests/test_iothub_configuration_manager.py
```diff
@@ -51,6 +53,65 @@ def iothub_configuration_manager():
...
+        IoTHubConfigurationManager.from_connection_string(connection_string=connection_string)
+
+        assert connection_string_auth_init_mock.call_args == mocker.call(connection_string)
+        assert protocol_client_init_mock.call_args == mocker.call(
+            connection_string_auth_mock, "https://" + connection_string_auth_mock["HostName"]
+        )
+
+
+@pytest.mark.describe("IoTHubConfigurationManager - .from_token_credential()")
+class TestFromTokenCredential(object):
```
cartertinney, line 87: Like above, I don't see tests for bad inputs, i.e. invalid hostnames. Presumably there's some kind of error condition there?

azure-iot-hub/tests/test_iothub_registry_manager.py
```diff
@@ -129,14 +130,98 @@ def mock_module_constructor(mocker):
...
+        azure.iot.hub.iothub_amqp_client.IoTHubAmqpClientSharedAccessKeyAuth,
+        "send_message_to_device",
     )
     return mock_uamqp_send
 
 
 @pytest.fixture
 def mock_uamqp_disconnect_sync(mocker):
-    return mocker.patch.object(IoTHubAmqpClientSharedAccessKeyAuth, "disconnect_sync")
+    return mocker.patch.object(
```
cartertinney, line 141: Is this fixture even used?

Again, as usual, I realize this is kind of how the test module currently existed prior to you coming along, but I question when we start having to patch all these methods of an object (in all these fixtures). Perhaps the object itself could be patched (thus making it's methods all implicitly mocks as well)?

I dunno, just spitballing here, probably not worth upending the table. But do make sure that we don't have any dead code that isn't being used at all.

cartertinney (commented):
I pointed out some things in certain places, but they aren't specific to those places, they seem like general issues.

azure-iot-hub/tests/test_iothub_configuration_manager.py
vishnureddy17: Discussed on teams

azure-iot-hub/tests/test_iothub_configuration_manager.py
```diff
@@ -47,10 +49,115 @@ def iothub_configuration_manager():
...
+    )
     return iothub_configuration_manager
 
 
+@pytest.mark.describe("IoTHubConfigurationManager - .from_connection_string()")
+class TestFromConnectionString(object):
+    @pytest.mark.parametrize(
+        "connection_string",
+        [
+            "HostName={hostname};DeviceId={device_id};SharedAccessKeyName={skn};SharedAccessKey={sk}".format(
```
cartertinney, line 63: Use the pytest.param() syntax here so you can provide an id for each test case that will provide readable output (Just l

cartertinney (approved):

PR #793, opened 2021-06-21
chore(azure-iot-device): Reorganized handler property definitions

cartertinney:
* Reduced duplication of code by moving handler property definitions to the abstract clients
* Reorganized order of definitions for clarity
* Added some missing internal docstrings


azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
```diff
@@ -130,6 +130,38 @@ async def _disable_feature(self, feature_name):
             # This branch shouldn't be reached, but in case it is, log it
             logger.info("Feature ({}) already disabled - skipping".format(feature_name))
 
+    def _generic_receive_handler_setter(self, handler_name, feature_name, new_handler):
```
BertKleewein, line 133: Something to think about:  I think it was maybe overkill to support both coroutines and functions as handlers, but another way to do this would be to wrap the function inside a coroutine inside _generic_receive_handler_setter.  This way, for async, all handlers are coroutines, and some of them internally call a function (on a different thread, of course, like we do.)

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/iothub/aio/async_clients.py
cartertinney: We have to support coroutines because of the handlers that need to invoke client coroutine methods (e.g. `.send_method_response()` )

PR #794, opened 2021-06-21
FEEDBACK ONLY: fix amqp service sas interval, add support for feedback and download …

BertKleewein:
…receivers

This is one way to address some of the AMQP weaknesses that we have in our service SDK.  I'm doing it this way because I don't want to invest in the `send_c2d_message` API and I would rather expose the underlying objects.

1. Default sas renewal bumped up to 1 hour. 
2. Define `amqp_message_send_client` as a public property.  Users can access this directly as a uAMQP `SendClient` object and get all of the functionality that this object provides.
3. Make the uAMQP client objects get created on-demand to save user resources.
4. Add `amqp_feedback_receiver_client` and `amqp_file_notification_recceiver_client` for delivery feedback and file upload notification feedback.

`send_message_to_device` can be kept for back compat and hopefully deprecated.


azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -70,20 +113,24 @@ def send_message_to_device(self, device_id, message, app_props):
         message = uamqp.Message(
             msg_content, properties=msg_props, application_properties=app_properties
         )
-        self.amqp_client.queue_message(message)
-        results = self.amqp_client.send_all_messages(close_on_done=False)
+        self.amqp_message_send_client.queue_message(message)
```
vishnureddy17, line 116: We need to check if the uAMQP object exists and create it if it doesn't.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -70,20 +113,24 @@ def send_message_to_device(self, device_id, message, app_props):
...
+        results = self.amqp_message_send_client.send_all_messages(close_on_done=False)
         if uamqp.constants.MessageState.SendFailed in results:
             raise Exception("C2D message send failure")
```
vishnureddy17, line 119: One thing that @cartertinney pointed out is that it's a bad idea to raise an `Exception` here and we should probably raise something different. This PR might be a good place to do it, though I'm not sure what the best thing to raise would be.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
```diff
@@ -70,20 +113,24 @@ def send_message_to_device(self, device_id, message, app_props):
...
+    def _get_target(self, operation):
+        endpoint = self._build_amqp_endpoint(
+            self.hostname, self.shared_access_key_name, self.shared_access_key
         )
-        operation = "/messages/devicebound"
-        target = "amqps://" + self.endpoint + operation
-        self.amqp_client = uamqp.SendClient(target)
+        return "amqps://" + endpoint + operation
 
     def _generate_auth_token(self, uri, sas_name, sas_value):
```
vishnureddy17, line 135: Another thing that @cartertinney pointed out is that we could reduce code duplication by leveraging the [SasToken class](https://github.com/Azure/azure-iot-sdk-python/blob/master/azure-iot-hub/azure/iot/hub/sastoken.py)

vishnureddy17 (commented):
Is the reason for exposing the `uAMQP` objects to reduce our development and maintenance overhead? One concern I have is that this would make it more difficult for users. Some concerns:

1. Users would have to understand which operations use AMQP and which ones use HTTP
2. Users would have to learn how to use `uAMQP` clients on top of learning our SDK
3. The API surface would be less cohesive between AMQP and HTTP operations

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
BertKleewein: That happens in the `amqp_message_send_client` property above.

azure-iot-hub/azure/iot/hub/iothub_amqp_client.py
BertKleewein: Yep, that's one of about 6 reasons that I"m pushing to deprecate this function.

PR #801, opened 2021-06-22
feature(azure-iot-device): Added internal PNP mode

cartertinney:
* Added a new internal "client_mode" attribute to track whether PNP is in use or not
* Added logic that will lock out certain APIs and properties depending on the mode
* Added (partial) implementation of PNP handlers + test stubs

### DIFF
--- a/azure-iot-device/azure/iot/device/iothub/models/pnp_translation.py
+++ b/azure-iot-device/azure/iot/device/iothub/models/pnp_translation.py
@@ -1,0 +1,26 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module contains utility functions to translate basic IoTHub objects
+into PNP objects
+"""
+
+from .commands import Command
+
+
+def method_request_to_command(method_request):
+    """Given a MethodRequest, returns a Command"""
+    pass
+    # return Command(
+    #     request_id=method_request.request_id,
+    #     component_name="",
+    #     command_name=method_request.name,
+    #     payload=method_request.payload,
+    # )
+
+
+def twin_patch_to_writable_property(twin_patch):
+    """Given a Twin Patch (dict), returns a WritableProperty"""
+    pass
--- a/azure-iot-device/azure/iot/device/iothub/sync_clients.py
+++ b/azure-iot-device/azure/iot/device/iothub/sync_clients.py
@@ -292,6 +292,8 @@ class GenericIoTHubClient(AbstractIoTHubClient):
         If the connection to the service has not previously been opened by a call to connect, this
         function will open the connection before sending the event.
 
+        This method is not compatible with PNP.
+
         :param message: The actual message to send. Anything passed that is not an instance of the
             Message class will be converted to Message object.
         :type message: :class:`azure.iot.device.Message` or str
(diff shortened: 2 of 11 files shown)


azure-iot-device/tests/iothub/aio/test_async_clients.py
```diff
@@ -1337,7 +1434,7 @@ class TestIoTHubDeviceClientDisconnect(IoTHubDeviceClientTestsConfig, SharedClie
 
 @pytest.mark.describe("IoTHubDeviceClient (Asynchronous) - .send_message()")
 class TestIoTHubDeviceClientSendD2CMessage(
-    IoTHubDeviceClientTestsConfig, SharedClientSendD2CMessageTests
+    SharedClientSendD2CMessageTests, IoTHubDeviceClientTestsConfig
```
vishnureddy17, line 1437: Can you explain why the order has to be changed here? (and in a bunch of other places)

vishnureddy17 (commented):
It might not be worth the extra effort to do it this way, but the way we mutually exclude pnp and non-pnp methods makes it seem like it'd make more sense to just have separate pnp and non-pnp clients.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -160,6 +161,22 @@ def _generic_receive_handler_setter(self, handler_name, feature_name, new_handle
...
+        translation function.
+
+        :param handler_to_wrap: Handler function that will be wrapped
+        :param translation_fn: Function that translates a non-PNP object passed by the handler
+            into the PNP equivalent
+        """
+
+        def translation_wrapper(handler_obj):
+            translated_obj = translation_fn(handler_obj)
+            handler_to_wrap(translated_obj)
```
BertKleewein, line 176: >handler_to_wrap(translated_obj) [](http://example.com/codeflow?start=12&length=31)

Do handlers ever return values?  I think the answer is no.  If the answer is yes, then you need to add `return` here.  Or just add it either way because it doesn't hurt.

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -315,6 +332,8 @@ def send_message(self, message):
         If the connection to the service has not previously been opened by a call to connect, this
         function will open the connection before sending the event.
 
+        This method is not compatible with PNP.
```
BertKleewein, line 335: >PNP [](http://example.com/codeflow?start=43&length=3)

This wording may have to change.  We're not using "PNP" in any API surfaces.  I'm not sure about documentation.

azure-iot-device/azure/iot/device/iothub/models/pnp_translation.py
```diff
@@ -0,0 +1,26 @@
...
+    pass
+    # return Command(
+    #     request_id=method_request.request_id,
+    #     component_name="",
+    #     command_name=method_request.name,
+    #     payload=method_request.payload,
+    # )
+
+
+def twin_patch_to_writable_property(twin_patch):
```
BertKleewein, line 24: FWIW, a twin _patch_ will never become what we call a WritableProperty object.  I think what we need here is a function to convert a desired property patch into a ClientProperties object.  Yes, the model names around properties are horrible.  I expect them to change.

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/iothub/sync_clients.py
cartertinney: Handlers never return values in our clients, and unless something massive were to change internally, they never will.

azure-iot-device/azure/iot/device/iothub/models/pnp_translation.py
cartertinney: Yeah, since I've decided to leave this alone for now, consider this is just a placeholder

PR #808, opened 2021-06-29
update properties APIs

BertKleewein:
1. We are not guaranteed to receive the `"__T": "c"` properties, so we can almost never know if something is a component or if it's a complex property.  This is because of the nature of device twin patches and also because we cannot assume that existing service apps set this field consistently. 

2. The only time we absolutely know that something is a component is when the user calls `set_component_property` for a reported value.

3. We need to set `"_T": "c"` if we know something is a component.  Even if the hub doesn't do the same thing for us, we have to do this as a matter of courtesy.

4. Names may change, but only for properties.py and the properties APIs.  The behavior should not change.

5. Think of this translation in your head: 
* desired property patch == `PropertiesCollection` object (with `version` attribute).  
* reported properties patch == `PropertiesCollection` object (without `version` attribute).  
* Full twin (from a get) == `ClientProperties` object

6. For a full twin, we do not merge desired and reported properties.  This means that writable property named `foo` will be in two places:
* The reported property will be in `properties_collection.get_attribute("foo")`.  If the user set the reported value using `WritablePropertyResponse`, the actual value will be in properties_collection.get_property("foo")["value"].
* The desired property will be in `properties_collection.writable_property_requests.get_property("foo")`

7. The JSON for the patch or the twin will be available in `properties_collection.backing_object`.  We expect the user to use this (for example,to iterate through properties).

8. There are all kinds of weird, but arguably understandable, oddities that you will no doubt discover.  We had to make many concessions.  Symmetric it is not.


azure-iot-device/azure/iot/device/iothub/models/pnp_translation.py
```diff
@@ -21,6 +22,20 @@ def method_request_to_command(method_request):
...
+    """Given a Twin Patch (`dict`), returns a `ClientPropertyCollection` object"""
+    obj = ClientPropertyCollection()
+    obj.backing_object = twin_patch
+    return obj
+
+
+def twin_to_client_properties(twin):
+    """Given a `Twin` object, return a `ClientProperties` object"""
+    obj = ClientProperties()
+    obj.backing_object = twin.reported_properties
```
cartertinney, line 35: does a ClientProperties have a `backing_object` attribute? It seems like it doesn't

cartertinney (approved):

PR #827, opened 2021-07-21
add a uuid for message_id in send_telemetry

vishnureddy17:


vishnureddy17:
Closing this, will open a separate PR to add an optional message ID parameter if we decide to do that.

PR #830, opened 2021-07-26
TEST ONLY (azure-iot-device): "Reauthorizing" added to ReconnectStage state machine

cartertinney:
don't review yet


cartertinney:
Went with a different branch - closing

PR #840, opened 2021-08-27
fix : Update dps-e2e.yaml

olivakar:


olivakar:
> IMO, we shouldn't use `PYTHONV6` because there is no such thing as PythonV6 and it might be confusing down the line.

Okay I made the naming change though it did find an existing item of the same name but possibly we do not need that now.
But if in the future this fails we have to change to V6 or some other naming. We have to just delete all other resource groups prior to that. For now I have not deleted the V6 resource group in case we ever need it.

vishnureddy17 (approved):

PR #844, opened 2021-09-08
fix(azure-iot-device): ReconnectStage State Machine Overhaul

cartertinney:
Overhauled the ReconnectStage to use a 5-state state machine in order to fix known issues. Among the changes:

* Auto-reconnect and SAS renewals should no longer cause problems if one or the other is turned off
* Operations now fail if they require a connection, and one cannot be automatically established
* All user-initiated actions should now return errors in case of failure rather than any kind of indefinite retry
* Created completely separate flows for automatic reconnection vs other connections
* Added more explicitness in conditional logic and improved documentation for clarity


BertKleewein:
>                 # above to run.  Just because we just connected, it doesn't mean the connection won't drop

Comment needs updating.

---
Refers to: azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py:483 in e327bce. [](commit_id = e327bce2a2f31298b50b9587cab9681bcd8442a2, deletion_comment = False)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
     """
 
-    WAITING_TO_RECONNECT = "WAITING_TO_RECONNECT"
-    LOGICALLY_CONNECTED = "LOGICALLY_CONNECTED"
-    LOGICALLY_DISCONNECTED = "LOGICALLY_DISCONNECTED"
+    CONNECTED = "CONNECTED"  # Client is connected (as far as it knows)
+    DISCONNECTED = "DISCONNECTED"  # Client is disconnected
+    CONNECTING = "CONNECTING"  # Client is in the process of connecting
+    DISCONNECTING = "DISCONNECTING"  # Client is in the process of disconnecting
+    REAUTHORIZING = "REAUTHORIZING"  # Client is in the process of reauthorizing (disconn->conn)
```
BertKleewein, line 939: >(disconn->conn) [](http://example.com/codeflow?start=81&length=15)

what does disconn->conn mean in this context?  disconnected->connected?  disconnected->connecting? etc.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
 
 
 class ReconnectStage(PipelineStage):
+
+    progress_states = [
```
BertKleewein, line 944: >progress_states [](http://example.com/codeflow?start=4&length=15)

I think I know what a progress_state is based on context, but I'm not sure if I'm right.  A comment migth help.  Would `intermediate_states` be a better name? (Or maybe I _don't_ understand.)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
-                    )
-                )
-                self.state = ReconnectState.LOGICALLY_DISCONNECTED
-                self.send_op_down(op)
+                if isinstance(op, pipeline_ops_base.ConnectOperation):
+                    if self.state is ReconnectState.CONNECTED:
+                        logger.debug(
+                            "{}({}): State is already CONNECTED. Sending op down".format(
+                                self.name, op.name
+                            )
```
BertKleewein, line 994: This is an interesting choice.  Sending the ConnectOperation down when self.state says that you're already connect impies that you don't trust self.state to be correct.  I can see why you migth not, but if you're passing down a ConnectOperation, wouldn't you go into ReconnectState.CONNECTING?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
     def _run_op(self, op):
-        if isinstance(op, pipeline_ops_base.ConnectOperation):
-            if self.state == ReconnectState.WAITING_TO_RECONNECT:
+        # Connection Retry Enabled
+        if self.pipeline_root.pipeline_configuration.connection_retry:
+
+            # NOTE: manual lock is used here rather than with context manager for structural
+            # reasons. We want the lock while checking the conditional, but not for what occurs
+            # as a result of the condition.
+            self.state_lock.acquire()
```
BertKleewein, line 970: >self.state_lock.acquire() [](http://example.com/codeflow?start=12&length=25)

Why do you need a lock at all?  Is there some code in this class that runs on a non-pipeline thread? f  Idon't think you need iffft.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
-                self._clear_reconnect_timer()
-                self._complete_waiting_connect_ops(
-                    pipeline_exceptions.OperationCancelled("Explicit disconnect invoked")
-                )
-                # TODO: There is a bug here. If in a WAITING_TO_RECONNECT state, and an explicit
-                # Disconnect occurs, it is completed here rather than going all the way down. This
-                # means that any in-flight operations are NOT cancelled.
-                op.complete()
+                self.waiting_ops.put_nowait(op)
+                self.state_lock.release()
```
BertKleewein, line 986: >self.state_lock.release() [](http://example.com/codeflow?start=16&length=25)

I think you release it here and you release it again somewhere below this.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+                    elif self.state is ReconnectState.DISCONNECTED:
+                        logger.debug(
+                            "{}({}): State changes DISCONNECTED -> CONNECTING. Sending op down".format(
+                                self.name, op.name
+                            )
+                        )
+                        self.state = ReconnectState.CONNECTING
+                        self._add_connection_op_callback(op)
+                    else:
+                        # This should be impossible to reach
```
BertKleewein, line 1006: ># This should be impossible to reach [](http://example.com/codeflow?start=24&length=36)

I don't think this is true -- what prevents this from happening?  We could be in ReconnectState.CONNECTING or ReconnectState.DISCONNECTING for a long time.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+
+            # NOTE: manual lock is used here rather than with context manager for structural
+            # reasons. We want the lock while checking the conditional, but not for what occurs
+            # as a result of the condition.
+            self.state_lock.acquire()
+
+            # If receiving a connection op while one is already in progress, wait for the current
+            # one to finish. This is kind of like a ConnectionLockStage, but inside this one.
+            # It has to happen here because just relying on a ConnectionLockStage before or after
+            # in the pipeline is insufficient, given that operations can spawn in this stage.
```
BertKleewein, line 975: Does this mean we can get rid of ConnectionLockStage?  If not, why not?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+                        # This should be impossible to reach
+                        logger.warning(
+                            "{}({}): Invalid State - {}".format(self.name, op.name, self.state)
+                        )
+
+                elif isinstance(op, pipeline_ops_base.ShutdownPipelineOperation):
+                    self._clear_reconnect_timer()
+
+                # In all cases the op gets sent down, but make sure to release the lock first
+                self.state_lock.release()
```
BertKleewein, line 1064: >self.state_lock.release() [](http://example.com/codeflow?start=16&length=25)

If anything after acquire and before here raises an exception, the lock remains held.  After that, all future calls to acquire will block, and that will be disaster.  Killing the python process will be the only recovery option.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+                        self.state = ReconnectState.CONNECTED
 
-            # BAD STATE (this block should not be reached)
-            else:
-                logger.warning(
-                    "{}: DisconnectEvent received while in unexpected state - {}, Connected: {}".format(
-                        self.name, self.state, self.pipeline_root.connected
-                    )
-                )
+                    # BAD STATE (this block should not be reached)
```
BertKleewein, line 1099: ># BAD STATE (this block should not be reached) [](http://example.com/codeflow?start=20&length=46)

But what if it _is_ reached?  There's nothing preventing it.  You're right that it _shouldn't_ happen, but it might.

If you're in ReconnectState.CONNECTED and you get another ConnectedEvent, then this code happens to be right.  But what if you're in ReconnectState.DISCONNECTED and you get a ConnectedEvent?  Or DISCONNECTING?  Are you 100% sure that you're not connected even though Paho just told you that you are?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+            # If error, set us back to a DISCONNECTED state. It doesn't matter what kind of
+            # connection op this was, any failure should result in a disconnected state.
+
+            # NOTE: Due to the stage waiting any ops if an ongoing connection op is in-progress
+            # as well as the way that the reconnection process checks if there is an in-progress
+            # connection op (and punts the reconnect if so), there is no risk here of setting
+            # directly to DISCONNECTED - the in-progress state being overwritten is always going
+            # to be due to this op that is now completing, we can be assured of that.
+            if error:
+                with self.state_lock:
```
BertKleewein, line 1186: >self [](http://example.com/codeflow?start=21&length=4)

Using `self` in here defeats the use of the weak reference.   This should `with this.state_lock:`

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+                        # fail, at which point the failure was a result of a manual operation and
+                        # reconnection is not supposed to occur. So either way, we end up where we want
+                        # to be despite the false positive - just be aware that this can happen.
+                        logger.debug(
+                            "{}({}): Not attempting to reconnect (Reauthorization in progress)".format(
+                                self.name, event.name
+                            )
+                        )
+
+                    # BAD STATE (this block should not be reached)
```
BertKleewein, line 1150: ># BAD STATE (this block should not be reached) [](http://example.com/codeflow?start=20&length=46)

Again, nothing guaranting that it won't be reached.  What if you're in ReconnectState.CONNECTING and you get a DisconnectedEvent?  That's not very hard to imagine.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -969,223 +930,336 @@ def do_retry():
...
+
                 if error:
-                    if this.never_connected:
-                        # any error on a first connection is assumed to be permanent error
-                        this.state = ReconnectState.LOGICALLY_DISCONNECTED
-                        this._clear_reconnect_timer()
-                        this._complete_waiting_connect_ops(error)
-                    elif this._should_reconnect(error):
+                    # Set state back to DISCONNECTED so as not to block anything else
+                    with self.state_lock:
```
BertKleewein, line 1226: >self [](http://example.com/codeflow?start=25&length=4)

should be `with this.state_lock:`

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: That's a good idea

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: This is what I get for trying to keep the comment under 100 characters. I'll make this clearer.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I think the ConnectionLockStage has value, but may need some adjustments. However, I didn't want to expand the scope of this PR any further. A little redundancy for now isn't the worst thing.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Fixed here and everywhere

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Fixed here and everywhere

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: This is one of the few changes I had yet to push - we now do change the state to DISCONNECTED, but still consider this to be irregular and log a warning as such.

It should not be possible to receive a DisconnectedEvent while in a CONNECTING state, because a DisconnectedEvent should only fire when a connection has been established, and then lost. Once a connection is established in transport, the first thing that happens is a ConnectedEvent fires up, which will immediately put us in a CONNECTED state. It should not be possible for the ReconnectStage to receive an ensuing DisconnectedEvent prior to receiving the ConnectedEvent.

I say "should", but I'm not going to completely rule out the possibility. However if we come across such possibility, we should seek to eliminate such aberrant behavior.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: As mentioned in your comment below, I have updated this logic to change the state anyway, but still raise the warning. This should minimize any interruption, but still bring an issue to our attention. If there is imprecision in our states, I want to know about it.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: The block above adds any ops to `waiting_ops` in the case where we are in a "progress state", i.e. CONNECTING, DISCONNECTING, REAUTHORIZING.

This entire section is an else block, indicating that we already know we are not in a "progress state"

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: This is one of the trickier points - my overall philosophy here is that I want the operations to go down the pipeline, although the ConnectionLockStage that comes next gets in the way of that, and I didn't want to touch that in this PR. The ConnectionLockStage is going to immediately complete the operation successfully without letting it ever hit the transport, and in such case, it would cause the state to get stuck forever in CONNECTING (because it is only moved to a stable state by a ConnectedEvent, which would never happen). Conversely, if the ConnectionLockStage fails it, because of how aggressive it can be with it's cancellation policy, we would end up in a potentially erroneous DISCONNECTED state as a result of the callback that was applied to the ConnectOperation.

For now this seemed the safest thing to do - keep the state where we think it probably should be, and let Events fired from the transport sort it out. However, when revisiting the ConnectionLockStage, this would need to be revisited as well.

I have added a comment to reflect this.



The issue with putting it into a CONNECTING state is that if that operation were to fail, it would (currently) then change the state, quite possibly erroneously, to DISCONNECTED. Certainly, if the ConnectionLockStage gets altered t

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: duh.  I looked back for that before commenting and missed it.   I recomment at least adding that as a comment.  Making it a warning made me think it was a "should not happen, but no big deal if it does".  It feels more like an assert because hitting this line indicates a funamental flaw in the code.  (I wish we had those).

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: >>  If there is imprecision in our states, I want to know about it, and figure out how we can get rid of it.

I couldn't agree more.  Who would have thought that "are we connected or not" would be such a pain in the butt to figure out.  I feel like maybe the best way to do this (not now, but maybe later) would be to have some object who's whole purpose is to keep track of connection state.  (I have something _like_ this in my track2 prototype, but it doesn't keep track of CONNECTING and DISCONNECTING, so it's not quite there yet.)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: Its a tricky problem.  My biggest area of uncertainty is around queueing of events.  The network stack, or Paho, or our pipeline thread could queue events, so the code isn't 100% guaranteed to handle events in the order that they are received.

For example, in pipeline_stages_mqtt.py, `_on_mqtt_connected` is decorated with `invoke_on_pipeline_thread_nowait`, so the call to `_on_mqtt_connected` can get queued up to run in the pipeline thread the next time it's avaialable, but that might be a while if the pipeline thread is busy doing something (like completing a bunch of pending operatings). 

That being said, getting stuck in an intermediate state is the biggest problem because it can block ops from going down.  I think your code favors leaving intermediate states, and that's a good thing.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: I could switch it to a logger.error rather than a warning?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Removed locks

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: it couldn't hurt.  Or a logger.critical.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -500,28 +461,25 @@ def _do_connect(self, op):
         """
         # Alias to avoid overload within the callback below
         # CT-TODO: remove the need for this with better callback semantics
```
anthonyvercolano, line 463: Did you mean to leave this todo?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -957,235 +915,375 @@ def do_retry():
...
+                    elif self.state is ReconnectState.DISCONNECTED:
+                        logger.debug(
+                            "{}({}): State changes DISCONNECTED -> CONNECTING. Sending op down".format(
+                                self.name, op.name
+                            )
+                        )
+                        self.state = ReconnectState.CONNECTING
+                        self._add_connection_op_callback(op)
+                    else:
+                        # This should be impossible to reach. If the state were intermeidate, it
```
anthonyvercolano, line 1006: spelled intermediate wrong.

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -957,235 +915,375 @@ def do_retry():
...
+                            "{}({}): State is already DISCONNECTED. Sending op down".format(
+                                self.name, op.name
+                            )
+                        )
+                        self._add_connection_op_callback(op)
+                        # NOTE: This is the safest thing to do while the ConnectionLockStage is
+                        # doing autocompletes based on connection status. When it is revisited,
+                        # this logic may need to be updated.
+                    else:
+                        # This should be impossible to reach. If the state were intermeidate, it
```
anthonyvercolano, line 1036: sp intermeidate
consider a spelling check extension.

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
cartertinney: Yeah this is fairly old. I can probably get rid of these, but I don't want to do it as part of this PR.

PR #847, opened 2021-09-13
Users/danhellem/uploadtoblob

danhellem:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-python/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

## Reference/Link to the issue solved with this PR (if any)

- renamed the existing "upload_to_blob.py" file to "upload_to_blob_x509.py" as it uses x509 cert
- added new "upload_to_blob.py" file to use device connection string. This is to keep the sample as simple as possible
-  added `hasattr(upload_result, "error_code")` check in both files. Existing code was failing because "error_code" attribute did not exist on dict object when there was no errors.
- updated readme files

### DIFF
--- a/azure-iot-device/samples/async-hub-scenarios/README.md
+++ b/azure-iot-device/samples/async-hub-scenarios/README.md
@@ -55,6 +55,9 @@ In order to use these samples, you **must** set your Device Connection String in
     az iot hub device-twin show --device-id <your device id> --hub-name <yoru IoT Hub name>
     ```
 
+* [upload_to_blob](upload_to_blob.py) - Upload file into the linked Azure storage account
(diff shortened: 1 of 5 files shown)


azure-iot-device/samples/async-hub-scenarios/README.md
```diff
@@ -55,6 +55,9 @@ In order to use these samples, you **must** set your Device Connection String in
     az iot hub device-twin show --device-id <your device id> --hub-name <yoru IoT Hub name>
```
rconard: Seems like a good opportunity to fix this typo

yoru -> your

cartertinney (approved):

PR #857, opened 2021-09-24
feature(azure-iot-device): Background Exception Handler

cartertinney:
* Pipeline will now send background exceptions to the user via the on_background_exception handler
* Background exceptions from threading modules and the handler manager do not.
* Background exceptions from the pipeline are still also given to the original logging function

I suspect this may need some alterations yet based on feedback. Thoughts on this approach?


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -139,9 +139,9 @@ def handle_pipeline_event(self, event):
         try:
             self._handle_pipeline_event(event)
         except Exception as e:
-            # Do not use exc_info parameter on logger.* calls.  This casuses pytest to save the traceback which saves stack frames which shows up as a leak
+            # Do not use exc_info parameter on logger.* calls.  This causes pytest to save the traceback which saves stack frames which shows up as a leak
             logger.error(msg="Unexpected error in {}._handle_pipeline_event() call".format(self))
-            handle_exceptions.handle_background_exception(e)
```
BertKleewein, line 144: >handle_exceptions.handle_background_exception(e) [](http://example.com/codeflow?start=12&length=48)

Can you remove handle_background_exception from the handle_exceptions module and stop importing it?

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -283,6 +302,17 @@ def _handle_pipeline_event(self, event):
...
+        elif isinstance(event, pipeline_events_base.BackgroundExceptionEvent):
+            logger.debug(
+                "{}: BackgroundExceptionEvent received. Calling on_background_exception_handler".format(
+                    self.name
+                )
+            )
+            if self.on_background_exception_handler:
+                pipeline_thread.invoke_on_callback_thread_nowait(
+                    self.on_background_exception_handler
+                )()
```
BertKleewein, line 314: >)() [](http://example.com/codeflow?start=16&length=3)

Shouldn't you pass the exception to the handler?

azure-iot-device/tests/conftest.py
```diff
@@ -26,7 +26,7 @@
 """
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture  # (scope="function")
```
BertKleewein, line 29: I assume this is a leftover.  function scope is default, so this isn't really necessary anyway (also on line 38 if you're going to remove this)

azure-iot-device/azure/iot/device/common/pipeline/pipeline_ops_base.py
```diff
@@ -100,21 +97,23 @@ def complete(self, error=None):
...
+        :raises: OperationError if an error occurs in resolving any callbacks
         """
         if error:
             logger.debug("{}: completing with error {}".format(self.name, error))
         else:
             logger.debug("{}: completing without error".format(self.name))
 
         if self.completed or self.completing:
-            e = pipeline_exceptions.OperationError(
+            raise pipeline_exceptions.OperationError(
```
BertKleewein, line 114: >raise [](http://example.com/codeflow?start=12&length=5)

All of these raises introduce major flow changes. 
1. who catches them?
2. unlike the old code, which logged the exception and continued, the raise cause execution to stop and the stack to unwind.  This means that any code that might have previously been called will no longer be called. 

For example, look at ConnectionLockStage._unblock.  If one of the ops in old_queue hits this code, the rest of the ops in the queue will be lost.  

Raising here is downright destructive.

BertKleewein (changes requested):
:clock1:

PR #895, opened 2021-11-01
fix(azure-iot-device): Remove double DisconnectEvent being issued in some cases

cartertinney:
* Removed functionality whereby `MQTTTransportStage` would raise a `DisconnectedEvent` on Pub/Sub/Unsub attempts if the `MQTTTransport` raised an exception
* This was unnecessary because the `on_mqtt_disconnected` handler being attached to the `MQTTTransport` would already do this


BertKleewein (approved):
:shipit:

PR #901, opened 2021-11-08
sample(azure-iot-device): Recurring Telemetry Samples

cartertinney:
* Added a "scenario sample" to show a more detailed application in a sample, rather than just demos of API calls


azure-iot-device/samples/sync-samples/recurring_telemetry.py
```diff
@@ -0,0 +1,48 @@
...
+def main():
+    # The connection string for a device should never be stored in code. For the sake of simplicity we're using an environment variable here.
+    conn_str = os.getenv("IOTHUB_DEVICE_CONNECTION_STRING")
+    # The client object is used to interact with your Azure IoT hub.
+    device_client = IoTHubDeviceClient.create_from_connection_string(conn_str)
+
+    print("IoTHub Device Client Recurring Telemetry Sample")
+    print("Press Ctrl+C to exit")
+    try:
+        # Connect the client.
```
cartertinney, line 22: Perhaps, like in the async sample, the contents of this try/except block (i.e. the actual function of the sample) should be factored out into a `send_recurring_telemetry` function.

azure-iot-device/samples/async-hub-scenarios/recurring_telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+        i += 1
+        msg = Message("test wind speed " + str(i))
+        msg.message_id = uuid.uuid4()
+        msg.correlation_id = "correlation-1234"
+        msg.custom_properties["tornado-warning"] = "yes"
+        msg.content_encoding = "utf-8"
+        msg.content_type = "application/json"
+        print("sending message #" + str(i))
+        await device_client.send_message(msg)
+        time.sleep(2)
```
cartertinney, line 30: This has to use `time.sleep()` since `asyncio.sleep()` has a somewhat different behavior, and turns program exit via the `KeyboardInterrupt` into a nightmare (even more than it already is annoying to deal with in the async event loop).

When I get to implementing support for async cancels, we might be able to start using `asyncio.sleep()`, not sure yet.

danhellem (approved):

PR #904, opened 2021-11-10
Test: Simple stress tests

BertKleewein:


device_e2e/aio/test_send_message_stress.py
```diff
@@ -0,0 +1,597 @@
...
+    3. Long haul tests.  These are designed to run for a long time under more realistic
+       scenarios.  Their purpose is to behave in a more realistic manner and find bugs that
+       might not come up in the artificial conditions we use for the first to sets of tests.
+
+This file contains tests from the first set.
+
+"""
+
+
+# Settings that apply to all tests
```
vishnureddy17, line 49:
```suggestion
# Settings that apply to all tests in this module
```

device_e2e/aio/test_send_message_stress.py
```diff
@@ -0,0 +1,597 @@
...
+
+            except (
+                ConnectionFailedError,
+                ConnectionDroppedError,
+                OperationCancelled,
+                NoConnectionError,
+            ) as e:
+                # These are the errors that client.connect or client.send_message can raise
+                # if we're not connected.  Can we limit this list?  Surely, we don't expect
+                # ConnectoinDroppedError if auto_connect==True.  Do we want to fail if we do?
```
vishnureddy17, line 127: If `ConnectionDroppedError` is not supposed to happen when `auto_connect` is `True`, we should fail IMO

device_e2e/aio/stress_results.py
```diff
@@ -0,0 +1,70 @@
...
+class Measurements(object):
+    """Object which defines the measurements we take durring a stress test"""
+
+    def __init__(self):
+        self.lock = threading.Lock()
+        self.peak_reconnect_time = 0
+        self.peak_resident_memory_mb = 0
+        self.peak_telemetry_arrival_time = 0
+        self.queued_telemetry_messages = 0
+        self.peak_queued_telemetry_messages = 0
```
vishnureddy17, line 17: `queued_telemetry_messages` might sound like the number of messages that were _ever_ in the queue rather than the number that is _currently_ in the queue
```suggestion
        self.messages_in_queue= 0
        self.messages_in_queue = 0
```

device_e2e/aio/test_send_message_stress.py
```diff
@@ -0,0 +1,597 @@
...
+        # go until time runs out and our list of futures is empty.
+        while not done_sending or len(futures) > 0:
+
+            # When time runs out, stop sending, and slow down out loop so we call
+            # asyncio.gather much less often.
+            if time.time() >= test_end:
+                done_sending = True
+                sleep_interval = 5
+
+            # if the test is still running, send another burst of messages
```
vishnureddy17, line 203:
```suggestion
            # if the test is still running, send another message
```

device_e2e/aio/stress_results.py
```diff
@@ -0,0 +1,70 @@
...
+    round up if the decimal part of x is >= .5, otherwise round down
+    """
+    return math.floor(x + 0.5)
+
+
+def two_decimal_places(x):
+    """
+    round x to two decimal places
+    """
+    return round(x * 100) / 100
```
vishnureddy17, line 32: Why not use the [python built-in `round()` function](https://docs.python.org/3/library/functions.html#round)?

device_e2e/aio/test_send_message_stress.py
```diff
@@ -0,0 +1,597 @@
...
+                    stress_measurements.peak_resident_memory_mb, current_resident_memory_mb
+                )
+                logger.info(
+                    "memory use = {}, new peak = {}".format(
+                        current_resident_memory_mb,
+                        stress_measurements.peak_resident_memory_mb,
+                    )
+                )
+
+            await asyncio.sleep(5)
```
vishnureddy17, line 391: May want to note somewhere that the memory use measurements are based on a sample taken every 5 seconds

vishnureddy17 (approved):

vishnureddy17:
I'm curious why you decided not to do a test with periodic bursts of messages.

device_e2e/aio/stress_results.py
BertKleewein: Because I only looked in the math module. :)

PR #905, opened 2021-11-12
sample(azure-iot-device): Removed the need for methods in exiting Edge sample

cartertinney:
* A bunch of the other sample work I've done has taught me a better way to exit Edge scenarios - it was time to port it back over to our repo


BertKleewein (approved):
:shipit:

PR #909, opened 2021-11-23
test: improve e2e logging, increase method connect timeout, and limit leak checking on test setup failure.

BertKleewein:


device_e2e/conftest.py
```diff
@@ -153,34 +153,18 @@ def pytest_runtest_setup(item):
...
-        raise
-
-
 @pytest.hookimpl(hookwrapper=True)
 def pytest_exception_interact(node, call, report):
     logger.error("------------------------------------------------------")
     logger.error("EXCEPTION RAISED in {} phase: {}".format(report.when, type(call.excinfo.value)))
     logger.error("------------------------------------------------------")
+
+    if hasattr(node, "leak_tracker"):
```
cartertinney, line 162: is there no way for us to clean up leaks in the case of error?

cartertinney (approved):

PR #912, opened 2021-12-01
fix(azure-iot-device): OverflowError for sufficiently long SAS expiry or retry interval

cartertinney:
Fixed an issue where sufficiently long SAS token expiry or retry interval could cause a OverflowError due to exceeding the max value supported by a waiting thread

### DIFF
--- a/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
+++ b/azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
@@ -23,6 +23,10 @@ from azure.iot.device.common.auth import sastoken as st
 
 logger = logging.getLogger(__name__)
 
+# Python 2 doesn't define this constant, so manually do it
+if sys.version_info <= (2, 7):
+    threading.TIMEOUT_MAX = 4294967.0
+
 
 @six.add_metaclass(abc.ABCMeta)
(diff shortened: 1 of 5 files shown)


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -23,6 +23,11 @@
 
 logger = logging.getLogger(__name__)
 
+# Python 2 doesn't define this constant, so manually do it
+if sys.version_info < (3,):
```
anthonyvercolano, line 27: This is different than the test in config.py?

azure-iot-device/tests/common/pipeline/config_test.py
```diff
@@ -6,10 +6,17 @@
...
 import abc
 import six
+import threading
+import sys
 from azure.iot.device import ProxyOptions
 from azure.iot.device import constant
 from azure.iot.device.common.pipeline.config import DEFAULT_KEEPALIVE
 
+# Python 2 doesn't define this constant, so manually do it
+if sys.version_info < (3,):
```
anthonyvercolano, line 16: This is different than the predicate in config.py

anthonyvercolano (approved):

PR #913, opened 2021-12-06
fix(azure-iot-device): Alarm not offset by sleep

cartertinney:
Alarms will now correctly trigger at the specified time, even if the system goes to sleep.

This was already supposed to have been implemented but apparently was not.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/alarm.py
+++ b/azure-iot-device/azure/iot/device/common/alarm.py
@@ -34,10 +34,8 @@ class Alarm(Thread):
         Overrides the method inherited from Thread.
         Will invoke the Alarm's given function at the given alarm time (accurate within 1 second)
         """
-        curr_time = time.time()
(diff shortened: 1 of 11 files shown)


azure-iot-device/azure/iot/device/common/alarm.py
```diff
@@ -25,10 +32,13 @@ def cancel(self):
     def run(self):
         """Method representing the thread's activity.
         Overrides the method inherited from Thread.
-        Will invoke the Alarm's given function at the given alarm time.
+        Will invoke the Alarm's given function at the given alarm time (accurate within 1 second)
         """
-        interval = self.alarm_time - time.time()
-        self.finished.wait(interval)
+        curr_time = time.time()
```
anthonyvercolano, line 37: Not sure of the utilty of the curr_time variable.

cartertinney: Good point.

PR #918, opened 2021-12-09
chore: Formally added license files to packages

cartertinney:
We only had a license on the root of the repo, and as such, weren't including it in our packages.


azure-iot-device/setup.py
```diff
@@ -53,13 +53,14 @@
     version=version,
     description="Microsoft Azure IoT Device Library",
     license="MIT License",
+    license_files=("LICENSE",),
```
vishnureddy17, line 56: Ah, 1-tuples in Python are so weird

vishnureddy17 (approved):

PR #920, opened 2021-12-10
chore: Updated language classifiers

cartertinney:
Formally indicated support for Python 3.9 and 3.10


cartertinney:
* Formally indicated support for Python 3.9 and 3.10

PR #926, opened 2022-01-05
fix job sample for calling device methods

BertKleewein:
Plus some minor improvements to all the service SDK samples.

### DIFF
--- a/azure-iot-hub/samples/invoke_command_sample.py
+++ b/azure-iot-hub/samples/invoke_command_sample.py
@@ -36,4 +36,6 @@ except msrest.exceptions.HttpOperationError as ex:
 except Exception as exc:
     print("Unexpected error {0}".format(exc))
 except KeyboardInterrupt:
-    print("Sample stopped")
+    print("{} stopped".format(__file__))
+finally:
+    print("{} finished".format(__file__))
--- a/azure-iot-hub/samples/iothub_job_manager_method_sample.py
+++ b/azure-iot-hub/samples/iothub_job_manager_method_sample.py
@@ -4,25 +4,32 @@
 # license information.
 # --------------------------------------------------------------------------
 
+import sys
 import os
-import datetime
-import time
 import uuid
+import time
+import datetime
 import msrest
 import pprint
 from azure.iot.hub import IoTHubJobManager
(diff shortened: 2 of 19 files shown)


azure-iot-hub/samples/iothub_job_manager_method_sample.py
```diff
@@ -0,0 +1,67 @@
...
+def create_method_job_request():
+    job = JobRequest()
+    job.job_id = "methodJob-" + str(uuid.uuid4())
+    job.type = "scheduleDeviceMethod"
+    job.cloud_to_device_method = CloudToDeviceMethod(
+        method_name="test_method",
+        payload={"key": "value"},
+        response_timeout_in_seconds=30,
+        connect_timeout_in_seconds=15,
+    )
```
BertKleewein, line 30: import/export and method call used to be a single sample.  I separated them for clarity.  These 2 fields were the only significant change needed to make the sample actually run.

azure-iot-hub/samples/invoke_command_sample.py
```diff
@@ -36,4 +36,6 @@
 except Exception as exc:
     print("Unexpected error {0}".format(exc))
 except KeyboardInterrupt:
-    print("Sample stopped")
+    print("{} stopped".format(__file__))
```
vishnureddy17, line 39: Since we're dropping Python2, can this be an f-string?

vishnureddy17 (approved):

azure-iot-hub/samples/iothub_registry_manager_module_method_sample.py
```diff
@@ -34,7 +35,11 @@
     iothub_registry_manager.delete_module(device_id, module_id)
```
anthonyvercolano, line 35: I haven't looked at these before.  Could you change the primary and secondary keys to generated uuid strings?  In the middle of the night some DRI is going to get a call because a sev 2 was created because a key is in the source.

anthonyvercolano: Please do it in the other samples also.

anthonyvercolano (approved):
Approved with a request.

PR #927, opened 2022-01-06
iot.device: add proxy options support for HTTP transport

zOrg1331:
This adds proxy options support for HTTP transport in the same manner as it was implemented for MQTT.


cartertinney:
Closing this as it will be included in #930

PR #930, opened 2022-01-24
feature(azure-iot-device): Support for proxies in HTTPTransport (TEST ONLY)

cartertinney:
* Replaced the use of the HTTPSConnection client with the use of the requests library for HTTP operations in the Transport
* Enabled HTTP, SOCKS4 and SOCKS5 proxy support

### DIFF
--- a/azure-iot-device/azure/iot/device/common/http_transport.py
+++ b/azure-iot-device/azure/iot/device/common/http_transport.py
@@ -21,7 +21,7 @@ class HTTPTransport(object):
     A wrapper class that provides an implementation-agnostic HTTP interface.
     """
 
-    def __init__(self, hostname, server_verification_cert=None, x509_cert=None, cipher=None):
+    def __init__(self, hostname, server_verification_cert=None, x509_cert=None, cipher=None, proxy_options=None):
         """
         Constructor to instantiate an HTTP protocol wrapper.
 
@@ -29,11 +29,13 @@ class HTTPTransport(object):
         :param str server_verification_cert: Certificate which can be used to validate a server-side TLS connection (optional).
         :param str cipher: Cipher string in OpenSSL cipher list format (optional)
         :param x509_cert: Certificate which can be used to authenticate connection to a server in lieu of a password (optional).
+        :param proxy_options: Options for sending traffic through proxy servers.
         """
         self._hostname = hostname
         self._server_verification_cert = server_verification_cert
         self._x509_cert = x509_cert
         self._cipher = cipher
+        self._proxy_options = proxy_options
         self._ssl_context = self._create_ssl_context()
 
     def _create_ssl_context(self):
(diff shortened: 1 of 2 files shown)


azure-iot-device/azure/iot/device/common/models/proxy_options.py
```diff
@@ -26,9 +27,11 @@ def __init__(
          If it is not provided, authentication will not be used (servers may accept unauthenticated requests).
         :param proxy_password: (optional) This parameter is valid only for SOCKS5 servers and specifies the respective password for the username provided.
         """
+        if proxy_type not in [socks.HTTP, socks.SOCKS4, socks.SOCKS5]:
```
cartertinney, line 30: I think we should probably not require it to be the `socks` library constants (which translate to 1, 2, and 3) but rather our own defined constants. As is, this forces the customer to import the `socks` library into THEIR code which is very unnecessary.

Furthermore, use of these specific constants is only relevant in the MQTT case, and this is supposed to be implementation agnostic.

Naturally we would have to still support the `socks` constants for backwards compatibility.

Thoughts?

azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -82,46 +107,83 @@ def request(self, method, path, callback, body="", headers={}, query_params=""):
...
-                "https {} request sent to {}, and {} response received.".format(
-                    method, path, status_code
-                )
-            )
-            response_obj = {"status_code": status_code, "reason": reason, "resp": response_string}
-            callback(response=response_obj)
+            # Note that various configuration options are not set here due to them being set
+            # via the HTTPAdapter that was mounted at session level.
+            if method == "GET":
+                response = session.get(url, data=body, headers=headers, proxies=self._proxies)
```
cartertinney, line 131: I wonder if these invocations should be perhaps running on the `http_thread` rather than the whole method. That would make the HTTPTransport work more analogously to MQTT (where Paho has it's own thread, but the MQTTTransport runs on the pipeline thread still).

This difference is why there's currently unittests failing due to timing issues FWIW

azure-iot-device/azure/iot/device/common/models/proxy_options.py
BertKleewein: I'm not especially concerned about this definition leaking up.  Adding a different enumeration for the sake of architectural purity just means that the user has another set of possible values to worry about, and they have to distinguish between `socks.SOCKS4` and, say, `azure.iot.device.common.proxy_types.SOCKS4`, and they both mostly mean the same thing, but one is accepted in one place and the other is accepted in a different place, 

Some folks will be just fine with this, some folks will use them interchangeably and incorrectly, some folks will spend time being confused about the difference, and some folks will define their own enumeration (say `customer.sock_type`) and their own mapping functions (`customer_sock_type_to_pysocks_sock_type()` and `customer_sock_type_to_azure_sock_type()`)

BertKleewein: I'm more concerned about the fact that we take a dependency on `PySocks` just for a few `isinstance` checks. We don't actually call any code inside `PySocks`, but we make all of our users install it.

In my perfect world, we would remove `PySocks` from `setup.py` and rewrite this to be something like 'if "SOCKS4' in str(proxytype) or "SOCKS5" in str(proxytype) or "HTTP" in str(proxytype):`.

FWIW, it's only 78 KB, which is only 2% of our total footprint (3.8 MB), but it's still a dependency that we don't need.

BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/common/models/proxy_options.py
cartertinney: Sounds good. I'll see what I can do.

PR #938, opened 2022-02-03
fix(azure-iot-device): Reauthorization on wrong thread

cartertinney:
Fixed an issue where after a failed reauthorization, the re-attempt of the reauthorization happened on the wrong thread


BertKleewein (approved):
:shipit:

PR #941, opened 2022-02-07
feature(azure-iot-device): Enabled server verification certs on ProvisioningClients

cartertinney:
* Added support for server_verification_certs
* Added a couple missing tests


BertKleewein (approved):
:shipit:

PR #942, opened 2022-02-07
feature(azure-iot-device): Removed use of socks library in API

cartertinney:
* Removed the need for the user to use constants from the socks library in the ProxyOptions API
* Updated samples accordingly


cartertinney (commented):
I didn't go so far as to define new constants on our library (e.g. azure.iot.device.HTTP) and currenly rely on the user to provide the string "HTTP", "SOCKS4" or "SOCKS5". This is specified in documentation for ProxyOptions, but I can add the constant definitions on the package if anyone feels strongly about it.

BertKleewein (approved):
:shipit:

PR #943, opened 2022-02-08
fix(azure-iot-device): stack overflow in connection retry, simplify and improve stress tests

BertKleewein:


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -1349,16 +1349,24 @@ def on_complete(op, error):
...
         if not self.waiting_ops.empty():
-            next_op = self.waiting_ops.get_nowait()
-            logger.debug("{}: Resolving next waiting op: {}".format(self.name, next_op.name))
-            self.run_op(next_op)
+            queuecopy = self.waiting_ops
+            self.waiting_ops = queue.Queue()
+
+            while not queuecopy.empty():
+                next_op = queuecopy.get_nowait()
+                if not next_op.completed:
```
cartertinney, line 1365: Is this condition necessary? I don't think it's possible to be completed and in the queue

cartertinney (approved):

azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
BertKleewein: "don't think it can happen" or "can guarantee that it won't happen"?  Calling `run_op` on a completed op can have unpredictable results, and I'd rather have an extra check here.

PR #944, opened 2022-02-09
fix(azure-iot-device): Prevented thread crash if SasToken refresh fails

cartertinney:
* Added error handling for an error that may occur when refershing a SasToken
* Note that we don't know what to do yet in response, but this will triage the immediate thread crashing issue.


BertKleewein (approved):
:shipit:

PR #945, opened 2022-02-10
fix(azure-iot-device): Addressed missing doc ingestion

cartertinney:
* Regenerated patch_documentation modules

PR #946, opened 2022-02-10
doc(azure-iot-device): README updates

cartertinney:
Minor README updates before we do LTS.

More changes will be required after LTS:
* Listing the LTS branch in the repo root README
* Updating sample links to point to the LTS branch on the azure-iot-device README

Note that we can push additional changes to the PyPi release as long as they are README only, so don't worry about that

NOTE ALSO: The changes from #945 got included on this PR by accident. Please make sure that it is merged first to prevent any issues.


BertKleewein (approved):
:shipit:

PR #948, opened 2022-02-10
test: add conect/disconnect stress tests, increase some timeouts

BertKleewein:


anthonyvercolano (approved):
One wonders what theory you are testing.

The  connect/disconnect tests have an unstated assumption that is being tested.  Perhaps it should be stated.

PR #952, opened 2022-02-17
chore(azure-iot-device): Remove support for Python 2.7 and 3.5

cartertinney:
Officially dropping support for Python 2.7 and 3.5

* Disallow install on those versions of Python
* Removed dependencies that are no longer necessary
* Removed `ChainableExceptions` in favor of using built in `raise from` functionality
* Removed `CallableWeakMethod` as Python 3 garbage collection will be sufficient
* Adjusted error handling where different versions of Python raised different errors
* Simplified `pipeline_thread` logic
* Removed references to 2.7 and 3.5 in documentation and samples
* Moved to Python 3-style `super()` invocations
* Removed custom implementation of `urlencode` in favor of using built-in `urlencode`
* Removed support for universal wheels
* Adjusted unittests that had Python 2.7 or 3.5 specific logic
* Removed Python 2.7 jobs from gate and canary


BertKleewein (approved):
:shipit:

PR #956, opened 2022-02-22
Test: add e2e tests for various valid and invalid payload scenarios

BertKleewein:


device_e2e/aio/test_send_message.py
```diff
@@ -39,6 +40,37 @@ async def test_connect_if_necessary(self, client, random_message, service_helper
...
+    @pytest.mark.it("Raises correct exception for un-serializable payload")
+    async def test_bad_payload_raises(self, client):
+        # There's no way to serialize a function.
+        def thing_that_cant_serialize():
+            pass
+
+        try:
+            await client.send_message(thing_that_cant_serialize)
+        except ClientError as e:
+            assert type(e.__cause__) == TypeError
```
cartertinney, line 52: generally better to use `isinstance()` as a rule

device_e2e/aio/test_send_message.py
```diff
@@ -39,6 +40,37 @@ async def test_connect_if_necessary(self, client, random_message, service_helper
         event = await service_helper.wait_for_eventhub_arrival(random_message.message_id)
         assert json.dumps(event.message_body) == random_message.data
 
+    @pytest.mark.it("Raises correct exception for un-serializable payload")
+    async def test_bad_payload_raises(self, client):
+        # There's no way to serialize a function.
+        def thing_that_cant_serialize():
+            pass
+
+        try:
```
cartertinney, line 49: We can use `with pytest.raises(ClientError) as e_info` instead, which is the preferred way to check errors in pytest. This will automatically raise an Assertion if the ClientError is not raised. The error itself then can be retrieved via `e_info.value`

e.g.

```python
with pytest.raises(ClientError) as e_info:
    await client.send_message(thing_that_can't_serialize)
assert isinstance(e_info.value.__cause__, TypeError)
```

Here and everywhere.

cartertinney (approved):

olivakar (approved):

PR #959, opened 2022-02-24
fix(azure-iot-device): ClientEvent Memory Leak

cartertinney:
* Fixed an issue where if no client event handlers were set (`on_connection_state_change`, `on_new_sastoken_required` or `on_background_exception`) ClientEvent objects could pile up in inboxes without being resolved
* Any ClientEvents that occur before a handler is set will now be lost. This was always the intended design, however.


BertKleewein (approved):
:shipit:

PR #963, opened 2022-02-28
chore(azure-iot-device): ConnectionLockStage -> AutoCompleteStage

cartertinney:
* ConnectionLockStage was mostly unnecessary due to how the ReconnectStage works, so it was removed.
* However, the functionality within it that auto-completed operations based on connection state was spun off into a new stage, since that logic is still quite necessary.


olivakar (commented):
Horton is failing though changes look fine.

cartertinney:
Gonna put this one on ice for now - development here discovered a bug blocking these changes, will implement a series of fixes in different PRs.

PR #967, opened 2022-03-03
test: Update e2e tests to use experimental backoff with jitter

BertKleewein:


cartertinney (approved):

device_e2e/aio/retry_async.py
```diff
@@ -0,0 +1,179 @@
...
+# Largest number of seconds to sleep between retries (before applying jitter)
+MAXIMUM_DELAY = 60
+
+# Number of seconds before an operation is considered "failed". This period starts before
+# the first attempt and includes the elapsed time waiting between any failed attempts.
+FAILURE_TIMEOUT = 5 * 60
+
+# Jitter-up factor. The time, after jitter is applied, can be up this percentage larger than the
+# pre-jittered time.
+JITTER_UP_FACTOR = 0.25
```
olivakar, line 40: is there any reason for the choice of these 2 values to be different ?

olivakar (approved):

device_e2e/aio/retry_async.py
BertKleewein: @olivakar - it's entirely arbitrary. All of our other implementations have 2 different numbers, so I did too.

PR #971, opened 2022-03-16
feat(azure-iot-device) : Client cert issuance for dps cert management

olivakar:


azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
```diff
@@ -79,7 +79,11 @@ async def register(self):
 
         register_async = async_adapter.emulate_async(self._pipeline.register)
         register_complete = async_adapter.AwaitableCallback(return_arg_name="result")
-        await register_async(payload=self._provisioning_payload, callback=register_complete)
+        await register_async(
+            payload=self._provisioning_payload,
+            callback=register_complete,
+            client_csr=self.client_csr,
```
cartertinney, line 85: Should this be an argument on the `register` call that is provided in the operation, or something provided at client creation and stored in the PipelineConfig?

azure-iot-device/azure/iot/device/provisioning/pipeline/pipeline_stages_provisioning.py
```diff
@@ -452,11 +461,12 @@ class DeviceRegistrationPayload(object):
     The class representing the payload that needs to be sent to the service.
     """
 
-    def __init__(self, registration_id, custom_payload=None):
+    def __init__(self, registration_id, custom_payload=None, client_csr=None):
         # This is not a convention to name variables in python but the
         # DPS service spec needs the name to be exact for it to work
         self.registrationId = registration_id
         self.payload = custom_payload
+        self.clientCertificateCsr = client_csr
```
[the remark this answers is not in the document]
olivakar: yes

azure-iot-device/azure/iot/device/constant.py
```diff
@@ -10,7 +10,7 @@
 IOTHUB_IDENTIFIER = "azure-iot-device-iothub-py"
 PROVISIONING_IDENTIFIER = "azure-iot-device-provisioning-py"
 IOTHUB_API_VERSION = "2019-10-01"
-PROVISIONING_API_VERSION = "2019-03-31"
+PROVISIONING_API_VERSION = "2021-11-01-preview"  # "2019-03-31"
```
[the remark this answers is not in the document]
olivakar: The existing DPS e2e tests  will not work unless we create a completely different set of resources which would involve a completely different key-vault and pipeline. So what kind of safety concern are you referring to ?

azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
olivakar: it should not be on the config , coz this has nothing to do with the pipeline . This is almost an optional parameter that the user may choose to send via the register call. If the user sends it then there is a certificate in the response. If not proved register can still happen but there will be no certificate in response.

azure-iot-device/azure/iot/device/constant.py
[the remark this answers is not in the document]
anthonyvercolano: Nit: Not sure why you have the comment of the old api string.

azure-iot-device/azure/iot/device/provisioning/aio/async_provisioning_device_client.py
cartertinney: Okay, so this value is directly tied to an individual registration then? In that case, why are we putting it as a property on the client?

olivakar: This is more like that payload. Hence followed the pattern. Even the payload has a setter on the client, but is only required while sending resgister.

cartertinney (approved):

PR #976, opened 2022-03-21
fix: release script (more)

cartertinney:
`bumpversion.exe` -> `bumpversion`


olivakar (approved):
this used to work before right ?

PR #978, opened 2022-03-23
fix: Enforced Windows usage for building release

cartertinney:
* VSTS seems to default to Linux now instead of Windows, breaking scripts.


BertKleewein (approved):
:shipit:

PR #980, opened 2022-03-24
fix(azure-iot-device): Added timeout to HTTP requests

cartertinney:
* Added the recommended 'timeout' argument to the `HTTPTransport`'s invocation of the `requests` library
* Intentionally did not provide customization in lieu of overarching timeout strategy.

### DIFF
--- a/azure-iot-device/azure/iot/device/common/http_transport.py
+++ b/azure-iot-device/azure/iot/device/common/http_transport.py
@@ -156,6 +156,13 @@ class HTTPTransport(object):
         except ValueError as e:
             # Allow ValueError to propagate
             callback(error=e)
+        except requests.exceptions.Timeout as e:
+            # Allow Timeout to propagate
+            # NOTE: This breaks the convention in transports where we don't expose anything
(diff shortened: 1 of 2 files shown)


azure-iot-device/azure/iot/device/common/http_transport.py
```diff
@@ -13,6 +14,10 @@
 logger = logging.getLogger(__name__)
 
 
+# NOTE: There should probably be a more global timeout configuration, but for now this will do.
```
anthonyvercolano, line 17: 10 seconds seems a bit short.

anthonyvercolano (approved):

azure-iot-device/azure/iot/device/common/http_transport.py
cartertinney: This is the same as a timeout for MQTT Subscribe/Unsubscribe and more than an MQTT Publish (which Paho times out after 5s).

I do agree that there should be a more uniform policy, but as the comment indicates, that would need to be a different (and larger) PR.

PR #981, opened 2022-03-29
fix (azure-iot-device) Add MQTT_ERR_KEEPALIVE mapping to fix connection issues caused by new Paho error code

BertKleewein:
There are three side-effects of this, two actual and one theoretical:
1. If Paho fails to receive a PINGRESP, it will return this RC, and the code in RetryConnectionStage will not try to reconnect.  This _should_ cause a failure to be returned to the client, but it will be a generic `ProtocolClientError` instead of a `ConnectionDroppedError`.
2. The log line printed `Unknown CONNACK rc=16` which is incorrect. This was not a `CONNACK` rc.

Also including a change that was found at the same time.  If transport.disconenct raises inside the connection watchdog, the pending connection op would never complete.


cartertinney (approved):

cartertinney:
Good catch on the disconnection failing in the watchdog.

PR #1010, opened 2022-06-15
refactor(azure-iot-device): Added PipelineNucleus object

cartertinney:
Added a new object called the `PipelineNucleus`. This will allow us to move global pipeline data out of the `PipelineRootStage` so that it doesn't have to serve that role in addition to being the entry/exit point to the pipeline, and moving data between the pipeline layer and the client layer of the SDK.

While the Nucleus isn't much to look at now, this will allow the design space for me to unify our two different connection states in a way that makes sense with the rest of our pipeline design, and allow potential room for additional functionality in the future.

Basically, the main role of the `PipelineRootStage` was getting in the way of it's ability to also manage global data in a more meaningful way - it would be very weird for another stage to directly change the connection state on the `PipelineRootStage` in response to an event, only for that same event to end up at the `PipelineRootStage` later. For clarity, we needed a new object to delegate global pipeline data and state to.


azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_pipeline.py
```diff
@@ -593,11 +594,11 @@ def pipeline_configuration(self):
...
-        return self._pipeline.pipeline_configuration
+        return self.nucleus.pipeline_configuration
 
     @property
     def connected(self):
         """
         Read-only property to indicate if the transport is connected or not.
         """
-        return self._pipeline.connected
+        return self.nucleus.connected
```
BertKleewein, line 604: >nucleus [](http://example.com/codeflow?start=20&length=7)

consistency: sometimes it's self.pipeline_nucleus, other times it's self.nucleus. I personally like the shorter term, but I like consistency even more.

BertKleewein (approved):
:shipit:

anthonyvercolano (approved):

PR #1011, opened 2022-06-16
refacotr(azure-iot-device): .pipeline_nucleus -> .nucleus

cartertinney:
Renamed the attribute referring to a PipelineNucleus object throughout the pipeline. Was working on a different PR, and it rapidly became apparent that `.pipeline_nucleus` was not only too long, but had readability problems and was bordering on collision issues (since the module name is also `pipeline_nucleus`).


lev-i (commented):
Does tests/common/pipeline/test_pipeline_nucleus also need to be updated?

lev-i (approved):

PR #1013, opened 2022-06-17
Add EnsureDesiredPropertiesStage kwarg

lev-i:
- Added ensure_desired_properties kwarg to iothub pipeline config
- Disabling ensure_desired_properties allows disabling sending GetTwinOperations to refresh desired properties
- Updated test_pipeline_stages_iothub to properly mimic the newly added pipeline nucleus


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -264,6 +266,8 @@ def create_from_connection_string(cls, connection_string, **kwargs):
         :param bool connection_retry: Attempt to re-establish a dropped connection (Default: True)
         :param int connection_retry_interval: Interval, in seconds, between attempts to
             re-establish a dropped connection (Default: 10)
+        :param bool ensure_desired_properties: A boolean to determine whether or not to refresh
```
cartertinney, line 269: Redundant, type is already listed in the param. Also needs more information. Should instead be

"Ensure the most recent desired properties patch has been received upon reconnections (Default: True)"

Here and everywhere.

azure-iot-device/azure/iot/device/iothub/pipeline/config.py
```diff
@@ -22,6 +30,7 @@ def __init__(self, hostname, device_id, module_id=None, product_info="", **kwarg
         :param str device_id: The device identity being used with the IoTHub
         :param str module_id: The module identity being used with the IoTHub
         :param str product_info: A custom identification string for the type of device connecting to Azure IoT Hub.
+        :param bool ensure_desired_properties: A boolean to determine whether or not to refresh desired properties in the EnsureDesiredPropertiesStage (Default:True)
```
cartertinney, line 33: Again, don't need to specify type since it's already there. In this case, since this is an internal class you also don't need to mention the default value since it's obvious from the definition.

Adapt this to the phrasing style of connection_retry and auto_connect parameters on the base config.

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -39,7 +39,12 @@ def _run_op(self, op):
             # as a way of enabling this functionality.  If the ConnectedEvent handler
             # sees this -1, it will send a GetTwinOperation to refresh desired properties.
 
-            if op.feature_name == constant.TWIN_PATCHES:
+            # Ensure_desired_properties is set to True by default, but if changed to false,
```
cartertinney, line 42: This is apparent from the code - what does this mean though? Should be a comment here explaining why this matters

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
```diff
@@ -65,6 +69,8 @@ def init_kwargs(self):
     @pytest.fixture
     def stage(self, mocker, cls_type, init_kwargs):
         stage = cls_type(**init_kwargs)
+        stage.nucleus = pipeline_nucleus.PipelineNucleus(pipeline_configuration=mocker.MagicMock())
```
cartertinney, line 72: This change will allow tests to pass, but you now don't have correct coverage. Adding this kwarg introduced functionality differences and not only do we have to write tests to cover these differences, but we have to make sure no previously written tests are too tied to the old paradigm.

This might be a little tricky - we can work on it together on Monday perhaps?

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
lev-i: Fixed in all cases, thanks

azure-iot-device/azure/iot/device/iothub/pipeline/pipeline_stages_iothub.py
```diff
@@ -35,11 +35,16 @@ def __init__(self):
...
-            # sees this -1, it will send a GetTwinOperation to refresh desired properties.
-
-            if op.feature_name == constant.TWIN_PATCHES:
+            # Ensure_desired_properties enables twin patches, when true, by setting last version
+            # to -1. The ConnectedEvent handler sees this and sends a GetTwinOperation to refresh
+            # desired properties. Setting ensure_desired_properties to false causes the GetTwinOp
+            # to not be sent. The rest of the functions in this stage stem from the GetTwinOperation,
+            # so disabling ensure_desired_properties effectively disables this stage.
+
+            if (
```
cartertinney, line 44: I would put this conditional even farther out - wrap the entire content of _run_op() in this condition

cartertinney (approved):

PR #1014, opened 2022-06-21
fix (azure-iot-device): PipelineNucleus connection state is now the single source of truth regarding connection

cartertinney:
* Removed the boolean `.connected` attribute being set on the `PipelineNucleus` by the `PipelineRootStage`
* `.connected` is now a property that is derived from the `connection_state` attribute, which is set by the `ConnectionStateStage`
* This is now the single source of truth for the connection state
* Added additional testing fixtures to allow for easy mocking of the connection state
* Note that changes made to the `ConnectionLockStage` are merely a stopgap until the entire stage can be removed


azure-iot-device/azure/iot/device/common/pipeline/pipeline_nucleus.py
```diff
@@ -4,10 +4,26 @@
...
 
+from enum import Enum
+
 
 class PipelineNucleus(object):
     """Contains data and information shared across the pipeline"""
 
     def __init__(self, pipeline_configuration):
         self.pipeline_configuration = pipeline_configuration
-        self.connected = False
```
BertKleewein, line 13: >connected [](http://example.com/codeflow?start=13&length=9)

Did you miss the MQTTPipeline.connected property?

cartertinney: No. It points to this new property here.

PR #1020, opened 2022-07-05
chore: load e2e environment from same environment variables that node scripts use

BertKleewein:


cartertinney:
Is there a reason for these changes? I was happy with the older variable names as they distinguish E2E variables from other variables which I have in my dev environment

PR #1021, opened 2022-07-08
chore: removed ya fiction references

cartertinney:
* Changed various constants used in tests and samples

### DIFF
--- a/azure-iot-device/tests/provisioning/shared_client_tests.py
+++ b/azure-iot-device/tests/provisioning/shared_client_tests.py
@@ -18,9 +18,9 @@ from azure.iot.device.common.pipeline.config import DEFAULT_KEEPALIVE
 logging.basicConfig(level=logging.DEBUG)
 
 
-fake_provisioning_host = "hogwarts.com"
-fake_registration_id = "MyPensieve"
-fake_id_scope = "Enchanted0000Ceiling7898"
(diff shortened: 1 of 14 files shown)


azure-iot-device/tests/provisioning/shared_client_tests.py
```diff
@@ -18,9 +18,9 @@
 logging.basicConfig(level=logging.DEBUG)
 
 
-fake_provisioning_host = "hogwarts.com"
-fake_registration_id = "MyPensieve"
-fake_id_scope = "Enchanted0000Ceiling7898"
+fake_provisioning_host = "hostname.com"
+fake_registration_id = "MyRegId"
+fake_id_scope = "Some0000Scope7898"
 fake_symmetric_key = "Zm9vYmFy"
```
anthonyvercolano, line 24: In variables that have key in there name I would generally set them with a uuid (by calling a function).
MUCH less likely to get hit by password checked in tools.

anthonyvercolano (approved):

azure-iot-device/tests/provisioning/shared_client_tests.py
cartertinney: We have already set up a custom rule to ignore this specific key variable in our static analysis

PR #1023, opened 2022-07-15
fix: prevent infinite GetTwinOperation loop

lev-i:
- Added connection logic to on_get_twin_complete to ensure GetTwinOperations aren't tried indefinitely when not connected
- Added appropriate unit tests
- Fixed minor typos

### DIFF
--- a/azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
+++ b/azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
@@ -33,9 +33,9 @@ fake_gateway_hostname = "__fake_gateway_hostname__"
 fake_server_verification_cert = "__fake_server_verification_cert__"
 fake_sas_token = "__fake_sas_token__"
 fake_symmetric_key = "Zm9vYmFy"
-fake_x509_cert_file = "fantastic_beasts"
-fake_x509_cert_key_file = "where_to_find_them"
-fake_pass_phrase = "alohomora"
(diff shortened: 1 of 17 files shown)


azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
```diff
@@ -668,6 +683,86 @@ def test_does_not_change_last_version_seen_if_same_version(
...
+        mocker.spy(stage, "report_background_exception")
+        return stage
+
+    @pytest.fixture
+    def get_twin_op(self, stage):
+        stage.last_version_seen = -1
+        stage.handle_pipeline_event(pipeline_events_base.ConnectedEvent())
+
+        get_twin_op = stage.send_op_down.call_args[0][0]
+        assert isinstance(get_twin_op, pipeline_ops_iothub.GetTwinOperation)
```
BertKleewein, line 708: Yes, perfect. For extra paranoia, I would assert the value of  `send_op_down.call_count` before and after the `ConnectedEvent` to make sure we're working with the `GetTwinOperation` that came from the `ConnectedEvent`

BertKleewein (approved):
:shipit:

azure-iot-device/tests/iothub/pipeline/test_pipeline_stages_iothub.py
lev-i: Good suggestion, thanks!

PR #1025, opened 2022-07-22
Tests for gate and for fuzzing

BertKleewein:


sdklab/regressions/regression_github_990.py
```diff
@@ -0,0 +1,284 @@
...
+2. After `PUBLISH` and before `PUBACK`, transport disconnects with `rc=1`
+3. On reconnnect, `transport.connect` raised `TlsExchangeAuthError(None,)` caused by SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)')`
+4. `TlsExchangeAuthError` is not transient, so `reconnect` fails. Since the reconnect is not user-initiated, no error is returned to the caller.
+5. `PUBACK` never received, so `send_message` never completes.
+
+The fix is to make `TlsExchangeAuthError` into a retryable error, so a reconnect gets initiated after step 3 above.
+
+In the interest of completeness, this test was expanded to 8 different scenarios based on 3 flags
+
+* Half of them simulate a retryable error (`TlsExchangeAuthError`), and half of them simulate a non-retryable error (bare `Exception`).
```
cartertinney, line 34: I would prefer if you used an exception you defined instead, since bare Exception carries some risk. Significantly less while not in a pytest context, but the same issue can arise anytime you use bare Excepts - there's always the potential that something is catching it and acting unexpectedly as a result.

sdklab/regressions/regression_github_990.py
```diff
@@ -0,0 +1,284 @@
...
+from test_utils import test_env, logging_hook
+from azure.iot.device.aio import IoTHubDeviceClient
+import azure.iot.device.common
+
+
+logging.basicConfig(level=logging.WARNING)
+logging.getLogger("azure.iot").setLevel(level=logging.DEBUG)
+
+"""
+This app simulates was a bug that was reported where a websockets connection was raising a
```
cartertinney, line 20: re: filenaming here - I'm not sure "regression_github_990" is a particularly great name, given that it requires knowing what GH990 was in order to know what to expect. I appreciate this long block comment here explaining the issue, that's fantastic, but you still have to open the file to see it, and as we add more these could get hard to identfy

On the other hand, I realize that it's a highly specific problem that might be hard to encapsulate in a filename.

Not requesting a change here because I don't actually have a better idea. Perhaps this is ideal after all, as having long verbose titles doesn't do much for a quick lookup either.

sdklab/regressions/regression_github_990.py
```diff
@@ -0,0 +1,284 @@
...
+async def main():
+    workaround_github_990()
+
+    # test with retryable errors.
+    tls_auth_error = ssl.SSLError()
+    tls_auth_error.strerror = "CERTIFICATE_VERIFY_FAILED"
+
+    # If `connection_retry` is `True`, the client should re-connect and
+    # `send_message` should succeed
+    await run_test(
```
cartertinney, line 224: Doesn't running a sequence of these tests encounter the same problem you wished to solve by going outside of pytest? sys exit will only happen at the very end of all tests, not once per test.

sdklab/regressions/regression_github_990.py
```diff
@@ -0,0 +1,284 @@
...
+
+        # Connect the device client.
+        await device_client.connect()
+
+        try:
+            print("Sending message...")
+            await device_client.send_message("This is a message that is being sent")
+            print("Message successfully sent!")
+            assert send_should_succeed
+        except Exception as e:
```
cartertinney, line 195: Again, let's try and use a custom defined exception here, this makes me very nervous about false positives

sdklab/regressions/regression_github_990.py
```diff
@@ -0,0 +1,284 @@
...
+
+    @functools.wraps(old_publish)
+    def new_publish(*args, **kwargs):
+        """
+        If this is the first call to `publish`, we set a 3 flags:
+        1. `block_next_publish` tells so `sock_recv` drops the next incoming packet, thus simulating a blocked `PUBACK`.
+        2. `fail_sock_send_until_reconnect` tells `sock_send` to stop sending outgoing packets, thus simulating a broken socket.
+        3. `raise_on_next_reconnect` tells `reconnect` to fail by raising `exception_to_raise_on_reconnect`
+        """
+        nonlocal fail_sock_send_until_reconnect, first_publish, raise_on_next_reconnect, block_next_puback
```
cartertinney, line 83: <3 python 3

sdklab/fuzzing/fuzz_send_message.py
```diff
@@ -0,0 +1,118 @@
...
+
+    finally:
+        print("Shutting down service helper")
+        await service_helper.shutdown()
+
+        print("Shutting down device client")
+        await device_client.shutdown()
+
+
+if __name__ == "__main__":
```
cartertinney, line 114: If we're accepting user arguments via command line, I'd like to see that documented

sdklab/fuzzing/fuzz_send_message.py
```diff
@@ -0,0 +1,118 @@
...
+
+    paho_fuzz_hook.add_paho_logging_hook(device_client)
+
+    try:
+
+        # Connect the device client.
+        print("connecting")
+        await device_client.connect()
+
+        # Start fuzzing after the client is connected
```
cartertinney, line 72: I feel as though this module needs a bit more documentation explaining what's going on here, what's being tested, and why it matters

test_utils/test_utils/paho_fuzz_hook.py
```diff
@@ -0,0 +1,249 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from . import logging_hook
+import random
+import functools
+
+
```
cartertinney, line 10: As we discussed, I'm not sure this belongs in test_utils as opposed to the more specific fuzzing tests in sdklab, unless you see this being useful in other contexts as well

test_utils/test_utils/paho_fuzz_hook.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    paho = logging_hook.get_paho_from_device_client(device_client)
+
+    old_sock_send = paho._sock_send
+
+    @functools.wraps(old_sock_send)
+    def new_sock_send(buf):
+        if random.random() < failure_probability:
+            log_func("---------- RAISING EXCEPTION")
+            raise Exception("Forced Send Failure")
```
cartertinney, line 221: Once again, here and everywhere I'm concerned about the use of Exception as opposed to a custom defined exception that inherits from it

cartertinney (commented):
Left a lot of comments and thoughts and opinions. None of them are necessary to address before merge, but hopefully some of them are useful

cartertinney (approved):

test_utils/test_utils/service_helper.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for
+# full license information.
+from .service_helper_sync import ServiceHelperSync
+import asyncio
+import concurrent.futures
+
+
+class ServiceHelper:
```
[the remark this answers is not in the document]
BertKleewein: That's a good idea. I'll keep it in mind, but I don't want to change that now.

sdklab/regressions/regression_github_990.py
BertKleewein: That's true, but PR 1023 was for a shutdown-specific bug, and I don't want to fall into the trap of "we found one shutdown bug that could only repro if you shutdown and then exited, so we should make all of our test paths shutdown and then exit."

sdklab/run_gate_tests.py
```diff
@@ -0,0 +1,65 @@
...
+    print()
+    print("FINAL RESULT: {}".format("FAILED" if len(failures) else "SUCCEEDED"))
+    if len(failures):
+        print()
+        for failure in failures:
+            print(failure)
+    print()
+    print("+-" * 66)
+
+    sys.exit(len(failures))
```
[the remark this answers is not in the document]
BertKleewein: The docker idea is interesting. I assume this specific app is temporary. The intention with this line is that individual test _apps_ can fail, and since those are in their own processes, those failures shouldn't affect subsequent tests. We'd rather run  as many as possible, especially since they're less likely to affect each other.

PR #1030, opened 2022-07-29
chore: Restructure repository for single project

cartertinney:
* Moved from monorepo style to src style project root
* All scripts moved to "scripts" directory
* All tests moved to "tests" directory
* E2E tests must be manually triggered


BertKleewein:
This really feels unnecessary. What is the benefit?
I appreciate code that is organized, but this is a merge nightmare and causes incompatibility problems and headaches, especially anyone who has WIP.

cartertinney:
> This really feels unnecessary. What is the benefit? I appreciate code that is organized, but this is a merge nightmare and causes incompatibility problems and headaches, especially anyone who has WIP.

re: flattening

The reasoning here would be that with our repo now only for a single product, the organization simply doesn't make much sense or follow any standardized conventions, which is a maintenance issue - it relies on institutional knowledge of the history of the repo to understand why anything is organized the way it is, which doesn't seem to me like a particularly good practice in the long term.  This also allows us to remove monorepo-style infrastructure (e.g. build and release scripts don't need to be able to build multiple things anymore, and those options really ought not to exist because they'd be broken) while reducing down to a single README, LICENSE, etc. 

If there are WIPs that are impacted by this, I have no issue delaying until those WIP can be completed, or assisting in the merge.

scripts/env_setup.py
```diff
@@ -35,16 +32,13 @@ def pip_command(command, error_ok=False):
...
 
-    # Install packages
-    for package_name in packages:
-        # Use an eager upgrade strategy to make sure we have all the latest dependencies.
-        # This way we will be running into any dependency-related bugs before customers do.
-        pip_command("install -U --upgrade-strategy eager -e {}".format(package_name))
+    # Install package
+    # Use an eager upgrade strategy to make sure we have all the latest dependencies.
+    # This way we will be running into any dependency-related bugs before customers do.
+    pip_command("install -U --upgrade-strategy eager -e .")
```
BertKleewein, line 41: >pip_command("install -U --upgrade-strategy eager -e .") [](http://example.com/codeflow?start=4&length=55)

This has the side-effect of polluting developer pip namespaces.  After this, I can do things like `from test.common import test_alarm`

setup.py
```diff
@@ -86,13 +86,23 @@
         "janus",
         "PySocks",
     ],
-    python_requires="!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5, <4",
+    python_requires=">=3.6, <4",
     packages=find_packages(
         exclude=[
```
BertKleewein, line 91: >exclude [](http://example.com/codeflow?start=8&length=7)

Is there a way to do this with `includes` instead of `excludes`? I may be wrong, but I assume we have to be careful to update setup.py every time we add a root-level folder, otherwise it will show up in the  package.

cartertinney: It seems there is, but I wanted to err on the side of caution, given that we've only tested the use of "excludes" before. I also don't love this, so I can give "includes" a try, I'll just need to verify the behavior.

scripts/env_setup.py
cartertinney: Good catch, I'll see about fixing this

cartertinney: Apparently this was ALWAYS true, this isn't a new issue. Very strange. Perhaps related to the `setup.py`.

setup.py
cartertinney: Update: decided to go with a modified `src` pattern where the `src` folder is just named `azure-iot-device`. This allows me to more easily distribute only the directories we want.

Additionally, it's more similar to the existing structure, so less mess when merging with WIPs

scripts/env_setup.py
BertKleewein: Of course, it would have always been true.  Going to a src folder should fix this if I'm understanding correctly.

cartertinney: Yes, moving to the src model did fix this

.flake8
```diff
@@ -6,6 +6,4 @@ ignore = E501,W503,E203
 exclude =
     .git,
     __pycache__,
-    ./azure-iot-provisioning-servicesdk/azure/iot/provisioning/servicesdk/protocol/models
-    ./azure-iot-hub/azure/iot/hub/protocol
-    ./azure_provisioning_e2e/iothubservice20180630/
+    tests/e2e/provisioning_e2e/iothubservice20180630/*
```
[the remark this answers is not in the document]
cartertinney: I just had to update the path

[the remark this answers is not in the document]
BertKleewein: iothubservice20180630 is generated code -- I assume flake8 wasn't passing because of that. For the other ones, I don't know.

BertKleewein (approved):
:shipit:

.flake8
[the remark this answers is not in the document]
cartertinney: The other ones no longer exist as `azure-iot-hub` code has been extracted

PR #1045, opened 2022-08-24
chore: Restructure testing directory for future testing

cartertinney:
* Moved unittests to their own directory
* Added a basic README outlining the 5 types of testing we intend to implement
* Fixed the asyncio warnings being raised during testing of Python 3.7+ (!!!!!!)


tests/README.md
```diff
@@ -0,0 +1,12 @@
+# Five types of tests
+
+## Unit Tests
+Test functionality of individual units, generally an object or a method (or even more specific). Generally whitebox style testing, with lots of mocking.
+## Integration Tests
+Test the intersections of distinct units. Limited mocking, lean more toward greybox or blackbox testing.
+## E2E Tests
+Test end to end functionality with no mocking, a live network, and real IoTHubs
+## Stress Tests
+- Coming Soon
```
lev-i, line 10: Should we put "To Be Implemented - Will test ..." so that intended functionality we discussed earlier this week is documented?

Maybe something like -
Stress Tests: "To Be Implemented - Will test outer limits of normal operation as well as edge cases that are generally not expected to occur."
Scenario Tests: "To Be Implemented - Will cover specific test cases to meet required objectives or vital functionality."

lev-i (approved):
Just the one potential suggestion, otherwise looks good

BertKleewein (approved):
:shipit:

tests/README.md
cartertinney: I figured I would add those when we have more clarity on exactly how they'll work - I didn't want to speak for Bert

PR #1046, opened 2022-08-25
Increase test timeouts

BertKleewein:


tests/e2e/iothub_e2e/sync/test_sync_send_message.py
```diff
@@ -19,6 +19,10 @@ class TestSendMessage(object):
     def test_sync_send_message_simple(self, client, random_message, service_helper, leak_tracker):
         leak_tracker.set_initial_object_list()
 
+        # Arbitrary sleep to ensure that EventHub listener is completely ready.  Only necessary
+        # on Windows machines.
+        time.sleep(5)
```
cartertinney, line 24: nit: if it's only necessary for Windows, you could check the OS with the `sys` module and do this conditionally.

Although I don't really care much either way

cartertinney (approved):

PR #1047, opened 2022-08-30
Test: Use different consumer groups for different E2E and nightly tests

BertKleewein:


cartertinney (approved):
If you could add a description to this PR for future reference on why this is necessary, it would be appreciated

PR #1048, opened 2022-09-01
[V3] feat: remove deprecated "receive" APIs

cartertinney:
- Removed all "receive" methods
- Removed all internal support for "receive" methods


azure-iot-device/azure/iot/device/iothub/inbox_manager.py
```diff
@@ -25,50 +25,16 @@ def __init__(self, inbox_type):
...
-        # compatibility
-        self.c2d_message_inbox = self._create_inbox()
-        self.input_message_inboxes = {}
-        self.named_method_request_inboxes = {}
-
-        # Set this value to True if want to only use unified message mode
-        self.use_unified_msg_mode = False
-
-    def get_unified_message_inbox(self):
+    def get_message_inbox(self):
```
cartertinney, line 33: "Carter!", you say, "Why does this class need to exist anymore when all it does is provide single line methods??? Doesn't this contribute to the over-complexity of the client??"

Yes. Once the async pipeline is in place and we only have one type of inbox (Async ones), it's probably safe to completely remove the InboxManager as a class. But for now it still provides value in that it allows us to abstract out the sync vs async inboxes

setup.py
```diff
@@ -79,7 +79,6 @@
         # Security issue below 1.26.5
         "urllib3>=1.26.5,<1.27",
         # Actual project dependencies
-        "deprecation>=2.1.0,<3.0.0",
```
BertKleewein, line 82: >deprecation [](http://example.com/codeflow?start=9&length=11)

can we remove janus yet?

BertKleewein: nevermind. figured it out.

BertKleewein (approved):
:shipit:

PR #1060, opened 2022-09-15
fix: atomic handlers

cartertinney:
* Made handlers atomic to better handle enable/disable feature failure case.
* On sync clients, failure here raised the wrong exceptions. Now it will raise the correct ones.


BertKleewein (approved):
:shipit:

azure-iot-device/azure/iot/device/iothub/sync_clients.py
```diff
@@ -154,6 +152,9 @@ def _generic_receive_handler_setter(self, handler_name, feature_name, new_handle
         elif new_handler is None and self._mqtt_pipeline.feature_enabled[feature_name]:
             self._disable_feature(feature_name)
 
+        # Set the handler on the handler manager.
+        setattr(self._handler_manager, handler_name, new_handler)
```
BertKleewein, line 156: >setattr [](http://example.com/codeflow?start=8&length=7)

This was much easier than I thought it would be!

PR #1067, opened 2022-10-18
fix: Removed RequestAndResponseStage republishing requests on connect

cartertinney:
* Manual republish on connect is unnecessary, given that Paho already does it automatically.


BertKleewein (approved):
:shipit:

PR #1075, opened 2022-11-07
Connection retry with recurring telemetry

olivakar:
Working in Windows and Linux so far.


samples/300/connect_retry_with_telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+    try:
+        while True:
+            encountered_no_error = await connect_with_retry(device_client, RETRY_NOS)
+            if not encountered_no_error:
+                print("Fatal error encountered. Will exit the application...")
+                logging.error("Fatal error encountered. Will exit the application...")
+                break
+            await do_work(device_client)
+    except KeyboardInterrupt:
+        print("IoTHubClient sample stopped by user")
```
cartertinney, line 160: You need a finally block with .shutdown(), yeah?

samples/300/connect_retry_with_telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+        print("Message sent...")
+        await asyncio.sleep(TELEMETRY_INTERVAL)
+    except Exception:
+        print("Caught exception while trying to send message....")
+        logging.debug(traceback.format_exc())
+
+
+async def connect_with_retry(device_client, number_of_tries=None):
+    if device_client.connected:
+        print("Client is already connected...")
```
cartertinney, line 113: Because connection retry is enabled in this sample (by default), it seems unnecessary to check the connection and issue a manual connect every pass of the loop in main. This function seems like it should be invoked outside of that loop just for the initial connect.

samples/300/connect_retry_with_telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+    except Exception:
+        # Clean up in the event of failure
+        device_client.shutdown()
+        raise
+
+    return device_client
+
+
+async def do_work(device_client):
+    global connected_event
```
cartertinney, line 94: If the manual connection wasn't done each pass of the main loop as I suggested in another comment, this wouldn't need to be it's own coroutine, and could be just part of that main loop itself.

If you end up keeping this coroutine, I'm not really a fan of the name `do_work` since I feel it isn't very descriptive. Perhaps describing what the work is would be best

samples/300/connect_retry_with_telemetry.py
```diff
@@ -0,0 +1,171 @@
...
+    print("Client is connected {}".format(device_client.connected))
+    if not device_client.connected:
```
cartertinney, line 96: Is this a desirable pattern? The vast majority of the time the client will be connected here - it might make more sense to respond to the `NoConnectionError` edge case (EAFP style) than check the connection before every send (LBYL style). Might help readability too, and potentially eliminate the need for the global `Event` variables

samples/300/connect_retry_with_telemetry.py
olivakar: i had once had the initial connect outside but i had to still catch exception and retry even at that time - so does not make sense to put it outside. I just bypass the connection logic if it is already connected.

samples/300/connect_retry_with_telemetry.py
olivakar: not sure what you mean completely. I am retrying all connection in case of transient errors. In case it is not connected - i am waiting for the connected event to fire before sending telemetry. I understand that is LBYL style - but do u mean to fire the disconnected event during NoConnectionError ?

samples/300/connect_retry_with_telemetry.py
```diff
@@ -0,0 +1,168 @@
...
+                    "Failed to connect the device client due to retryable error.Sleeping and retrying after some time..."
+                )
+                await asyncio.sleep(ATTEMPT_INTERVAL)
+            else:
+                print("Failed to connect the device client due to not-retryable error....")
+                return False
+
+
+async def run_sample(device_client):
+    encountered_no_error = await connect_with_retry(device_client, RETRY_NOS)
```
cartertinney, line 123: Why is there a max retry on the initial connect anyway? It seems inconsistent - if the application fails too many times while trying to connect, it throws a RuntimeError, but once that first connect happens, the network can drop indefinitely without application exit.

cartertinney (approved):

PR #1076, opened 2022-11-08
[V3] DPS API changes

cartertinney:
* Added an explicit `.shutdown()` API
* Register now disconnects after any result instead of shutting down
* Enhanced the error pathway on `.register` to ensure the client cannot enter a broken state


BertKleewein (approved):
:shipit:

PR #1078, opened 2022-11-15
[V3] fix: re-enabled (some) memory leak testing in E2E

cartertinney:
* Enabled some memory leak tests that had previously been disabled
* Added a `flush_messages` fixture to the E2E infrastructure in order to avoid spurious leaks
* Other small enhancements to logging

### DIFF
--- a/azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
+++ b/azure-iot-device/azure/iot/device/provisioning/abstract_provisioning_device_client.py
@@ -224,6 +224,11 @@ class AbstractProvisioningDeviceClient(abc.ABC):
         return cls(mqtt_provisioning_pipeline)
 
     @abc.abstractmethod
+    def shutdown(self):
+        """Shut down the client."""
+        pass
+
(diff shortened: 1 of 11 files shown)


azure-iot-device/azure/iot/device/common/pipeline/pipeline_stages_base.py
```diff
@@ -401,7 +401,7 @@ def _cancel_token_update_alarm(self):
         old_alarm = self._token_update_alarm
         self._token_update_alarm = None
         if old_alarm:
-            logger.debug("Cancelling SAS Token update alarm")
+            logger.debug("{}: Cancelling SAS Token update alarm".format(self.name))
```
BertKleewein, line 404: >logger.debug("{}: Cancelling SAS Token update alarm".format(self.name)) [](http://example.com/codeflow?start=12&length=71)

Just out of curiosity, did you add this for consistency or does self.name give you some extra info that was useful in debugging this?

BertKleewein (approved):
:shipit:

PR #1080, opened 2022-11-17
[V3] Aligned Publish, Subscribe and Unsubscribe with Paho behavior

cartertinney:
* Publishes no longer fail with a `NoConnectionError`, since Paho retains the messages and will publish them when reconnection occurs. APIs that send MQTT publishes will now hang until completion, either successful or with failure.

* Subscribes and Unsubscribes will still fail from `NoConnectionError`, but if one somehow gets through to Paho before it realizes the connection has dropped, we will now fail it with a `NoConnectionError` once aware of the connection drop, rather than letting it hang around in the transport indefinitely.

* Accordingly, setting `connection_retry` to False will no longer zealously cancel operations upon a connection drop. Instead, normal rules will apply - the transport will be responsible for failing or keeping alive an operation.

* `OperationTimeoutStage` has been removed as it no longer serves a purpose - Subscribe and Unsubscribe will either fail at first, or fail when the connection drops - no need for timers waiting for an arbitrary limit. Additionally, it better reflects Paho behavior, since a Subscribe could always be successful until Paho realizes the connection drops, so cancelling prematurely is a deviation from the true behavior.

* Together all these changes also fix the memory leak for timed out subscribes leaving references to their callbacks in the transport.

* With the memory leak fixed, all leak tracking has now been re-enabled in the E2E tests


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -652,15 +674,38 @@ def complete_operation(self, mid):
                 # fully expected.  QOS=1 means we might get 2 PUBACKs
                 logger.debug("No callback set for MID: {}".format(mid))
 
+    def fail_all_subs_and_unsubs(self, error_cls, error_msg):
```
cartertinney, line 677: This solution is a little over-fitted to solving the problem in question, I admit, but when I make the transport async I'm going to have to re-implement a lot of this anyway. For now, this implementation produces the desired behavior, and that's good enough IMO.

BertKleewein (approved):
:shipit:

PR #1082, opened 2022-11-23
Feedback requested: Python SDK config and status object for debugging

BertKleewein:


azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -49,6 +51,104 @@
...
+    shut_down: bool = False
+
+    time_since_last_paho_traffic_in: str = ""
+    time_since_last_paho_traffic_out: str = ""
+    client_object_id: int = 0
+    thread_name: str = ""
+    thread_is_alive: bool = False
+    len_out_mesage_queue: int = 0
+    len_in_message_queue: int = 0
+    len_out_pakcet_queue: int = 0
```
anthonyvercolano, line 112: Spelling?

anthonyvercolano (approved):
Well I like it.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -565,6 +693,59 @@ def publish(self, topic, payload, qos=1, callback=None):
...
+
+        return self._paho_status
+
+    def get_debug_config(self):
+        """
+        Return an infomrational status object that describes the configuration of the Paho transport
+        """
+        config = PahoConfig()
+
+        config.transport = self._mqtt_client._transport
```
cartertinney, line 729: I feel as though accessing convention-private attributes of Paho is dangerous. There is no guarantee they don't change in an update to Paho, which could break anyone using our library until we make an update. Furthermore, that risk doesn't really seem worth it for a logging utility.

dev_utils/dev_utils/iptables.py
```diff
@@ -64,28 +64,39 @@ def transport_to_port(transport):
         )
 
 
-def disconnect_output_port(disconnect_type, transport, host):
+def disconnect_output_port(disconnect_type, transport, host=None):
```
cartertinney, line 67: Great change. Was thinking of doing this myself.

cartertinney (commented):
Comments

azure-iot-device/azure/iot/device/common/mqtt_transport.py
```diff
@@ -49,6 +51,104 @@
 }
 
 
+@dataclasses.dataclass
+class PahoStatus(object):
+    connect_rc_codes: dict = dataclasses.field(default_factory=dict)
+    """Value->count dictionary of rc codes returned from the `connect` method"""
```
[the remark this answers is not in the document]
BertKleewein: No, not nitpicky. These string literals end up taking space at runtime, so the distinction is important. I did it this way on purpose.  Even though docstrings on attributes are technically not part of the Python standard, some tools will pick these up (if you put them _after_ the attribute definition). I think PyCharm will pick these up, and maybe some versions of Sphynx.

azure-iot-device/azure/iot/device/common/mqtt_transport.py
BertKleewein: I agree that it's dangerous. I also think it's super useful -- at least some of these are. Using getattr() with a default would be much safer.

PR #1085, opened 2022-12-06
Sample with items in queue

olivakar:
The `connect_retry_with_telemetry` was already reviewed. Simply transferring it to this branch.

The other file has been running for 3 days on Windows successfully.


samples/solutions/producer_consumer.py
```diff
@@ -0,0 +1,244 @@
...
+        self.sleep_time_between_conns = ATTEMPT_INTERVAL
+        self.try_number = 1
+
+    async def create_client(self, conn_str):
+        try:
+            # Create a Device Client
+            self.device_client = IoTHubDeviceClient.create_from_connection_string(conn_str)
+            # Attach the connection state handler
+            self.device_client.on_connection_state_change = self.handle_on_connection_state_change
+        except Exception as e:
```
BertKleewein, line 93: >except Exception as e: [](http://example.com/codeflow?start=8&length=22)

I'm surprised that you just swallow exceptions and continue here.

samples/solutions/producer_consumer.py
```diff
@@ -0,0 +1,244 @@
...
+            "handle_on_connection_state_change fired. Connected status : {}".format(
+                self.device_client.connected
+            )
+        )
+        if self.device_client.connected:
+            self.log_info_and_print("Connected connected_event is set...")
+            self.disconnected_event.clear()
+            self.connected_event.set()
+            self.retry_increase_factor = 1
+            self.sleep_time_between_conns = ATTEMPT_INTERVAL
```
BertKleewein, line 109: >ATTEMPT_INTERVAL [](http://example.com/codeflow?start=44&length=16)

instead of `ATTEMPT_INTERVAL`, maybe `INITIAL_SLEEP_TIME_BETWEEN_CONNS` ?

samples/solutions/producer_consumer.py
```diff
@@ -0,0 +1,244 @@
...
+            if self.exit_app_event.is_set():
+                return
+
+    async def wait_for_connect_and_send_telemetry(self):
+        while True:
+            if not self.device_client.connected:
+                self.log_info_and_print("waiting for connection ...")
+                await self.connected_event.wait()
+            else:
+                msg = await self.message_queue.get()
```
BertKleewein, line 136: >msg = await self.message_queue.get() [](http://example.com/codeflow?start=16&length=36)

`asyncio.queue.get()` will wait until an item is in the queue. What happens if `self.exit_app_event` gets set while this all is waiting for an item to come into the queue?

samples/solutions/producer_consumer.py
```diff
@@ -0,0 +1,244 @@
...
+                [
+                    self.disconnected_event.wait(),
+                    self.exit_app_event.wait(),
+                ],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            await asyncio.gather(*done)
+            [x.cancel() for x in pending]
+            if self.exit_app_event.is_set():
+                self.log_info_and_print("Exiting while connected")
```
BertKleewein, line 163: >Exiting while connected" [](http://example.com/codeflow?start=41&length=24)

Shouldn't you also return here?

samples/solutions/producer_consumer.py
```diff
@@ -0,0 +1,244 @@
...
+            done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
+            await asyncio.gather(*done)
+        except KeyboardInterrupt:
+            self.log_error_and_print("IoTHubClient sample stopped by user")
+        except Exception as e:
+            self.log_error_and_print("Exception in main loop: {}".format(get_type_name(e)))
+        finally:
+            self.log_info_and_print("Exiting app")
+            self.exit_app_event.set()
+            self.log_info_and_print("Waiting for all coroutines to exit")
```
BertKleewein, line 237: >self.log_info_and_print("Waiting for all coroutines to exit") [](http://example.com/codeflow?start=12&length=61)

You should decide how you want to shut down these tasks. You set `app_exit_event`, which should eventually cause the tasks to complete, but you don't wait for the tasks to actually exit. If you're just going to cancel all the tasks after setting this event, then having this event in the code doesn't really do much.

BertKleewein (approved):
:shipit:

samples/solutions/producer_consumer.py
olivakar: yeah i should probably exit the app

samples/solutions/producer_consumer.py
olivakar: changed it to first completed with the queue and exit app

samples/solutions/producer_consumer.py
olivakar: yes

samples/solutions/producer_consumer.py
olivakar: actually i did both types one to wait and one to cancel it did not really change the behavior - i expected it would. But i have deleted the cancel for now.

PR #1086, opened 2022-12-10
[V3] Added initial MQTTClient

cartertinney:
* Added an MQTTClient class (fka MQTTTransport)
* Implemented connect, disconnect and reconnect
* Implemented test infrastructure


v3_async_wip/v3_async_wip/transport_helper.py
```diff
@@ -0,0 +1,101 @@
+# -------------------------------------------------------------------------
```
cartertinney, line 1: Just a helper for manual testing until E2E testing can be implemented. You can ignore this.

v3_async_wip/v3_async_wip/exceptions.py
```diff
@@ -0,0 +1,28 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""This module defines exceptions used by other modules"""
```
cartertinney, line 6: Very much work in progress - recoverable vs non-recoverable hasn't been implemented meaningfully (yet)

v3_async_wip/tests/test_mqtt_client.py
```diff
@@ -0,0 +1,1368 @@
...
+        assert not client.is_connected()
+
+        # Second disconnect was completed without invoking disconnect on Paho because it is
+        # already disconnected
+        await disconnect_task2
+        assert mock_paho.disconnect.call_count == 1
+        assert not client.is_connected()
+
+
+@pytest.mark.skipif(sys.version_info < (3, 8), reason="AsyncMock not supported by Python 3.7")
```
cartertinney, line 1276: Once I get out of this MQTT layer I'll find a solution for backporting AsyncMocks to 3.7 since we'll need them for testing other layers/modules and we can't drop 3.7 until December 2023

v3_async_wip/v3_async_wip/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+# Non-recoverable
+class NonRecoverableException(Exception):
+    """A exception that is not recoverable"""
+
+    pass
+
+
+class AuthorizationError(NonRecoverableException):
```
BertKleewein, line 27: This is tricky, because Auth errors are sometimes recoverable, depending on the type of auth and the level you're operating at. 

For example:
* An x509 connect might fail because the clock is wrong. If some other process fixes the clock and that x509 connect is able to succeed.
* Or the x509 cert has a "not valid until" field, and the cert is invalid until some time passes and then it becomes valid.
* Or an operator might enable (or re-enable) some credentials on the Azure portal and suddenly the client is able to connect. 

You might call thees "external actions", but, if the customer wants a rock-solid device that sits in a closet and never needs maintenence, they should consider an auth error to be recoverable. Even if it tries to reconnect every 30 seconds for 2 months, that might be better (and cheaper) than sending a tech to replace the device.

I suspect the number of "non-recoverable" errors is much smaller than we think. If we define "non-recoverable" as "no matter what happens, this call will never succeed", then a JSON parse error might be non-recoverable. If it fails because the format of the string is wrong it could never succeed. If JSON parse fails because it can't allocate a buffer, it's a little more murky because that parse _could_ succeed if there was more memory.

Likewise, "ConnectionFailed" might happen because the wifi router is unplugged. Try again and it might succeed.

I wonder if "fatal exception" and "non-fatal exception" is a better split, with a very high bar for "fatal", like a fatal error would require a reboot or hardware replacement. I don't know if this makes sense or not, but I think it doesn't make sense.  I don't think "how should the customer respond to this exception" should be part of the exception. Or, if it is part of the exception, it should maybe be a property.

Putting it as a base class makes "Failed connections are never recoverable" part of the API. If we realize that AuthorizationError is actually recoverable, we need to change it's base class, and that's a breaking change.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+
+                f = asyncio.run_coroutine_threadsafe(set_connected(), this._loop)
+                # Need to wait for this one to finish since we don't want to let another
+                # Paho handler invoke until we know the connection state has been set.
+                f.result()
+            else:
+                # Notify tasks waiting on failed connection
+                async def notify_connection_fail():
+                    async with this._connect_failed_cond:
+                        this._connect_failed_cond.notify_all()
```
BertKleewein, line 163: Do we want to include the RC as part of teh notification?

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+                f = asyncio.run_coroutine_threadsafe(set_disconnected(), this._loop)
+                # Need to wait for this one to finish since we don't want to let another
+                # Paho handler invoke until we know the connection state has been set.
+                f.result()
+
+            if do_stop_network_loop:
+
+                async def stop_network_loop():
+                    logger.debug("Stopping Paho network loop due to unexpected disconnect")
+                    client.loop_stop()
```
BertKleewein, line 221: It may not ever be necessary to stop the loop. IIRC, this was added to our code before the `reconnect_on_failure` flag was added to Paho.

v3_async_wip/v3_async_wip/exceptions.py
cartertinney: Yeah, I'm not tied to this at all. It's more or less just a placeholder for now, pending future discussions.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: It is necessary. You can't reconnect until you stop, and then restart the loop as I recall.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+        try:
+            while True:
+                async with self.disconnected_cond:
+                    await self.disconnected_cond.wait_for(
+                        lambda: not self.is_connected() and self._desire_connection
+                    )
+                try:
+                    logger.debug("Reconnect Daemon attempting to reconnect...")
+                    current_attempt = asyncio.create_task(self.connect())
+                    await current_attempt
```
BertKleewein, line 254: Why not just do `await self.connect()`?

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+                if self._auto_reconnect and not self._reconnect_daemon:
+                    self._reconnect_daemon = asyncio.create_task(self._reconnect_loop())
+
+                # Make sure the tasks are running
+                await asyncio.sleep(0)
+
+                # Paho Connect
+                logger.debug("Attempting connect using port {}".format(self._port))
+                try:
+                    rc = await self._loop.run_in_executor(
```
BertKleewein, line 334: Too bad we can't use `asyncio.to_thread`. The syntax is much cleaner.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: So that I can cancel the connect task if the reconnect daemon task is cancelled. Not completely sure that it's necessary (need to investigate cancel more), but it seemed safer for now.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: I could backport it if you feel strongly about it - the source code required is very short.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+                    success.cancel()
+                    failure.cancel()
+                    raise exceptions.ConnectionFailedError(
+                        "Connect returned rc {} - {}".format(rc, message)
+                    )
+
+                # Start Paho network loop. If already started, this will return a fail code,
+                # but we don't really care - no harm, no foul.
+                logger.debug("Starting Paho network loop")
+                self._mqtt_client.loop_start()
```
BertKleewein, line 363: I stronly recomment asserting `self._mqtt_client._thread` and `self._mqtt_client._thread.is_alive()`. There are paths through Paho that stop the thread but don't set `_thread` to `None`, and `loop_start` only checks that the `_thread` object is not `None`. It doesn't check if the thread is actually alive.  This assert will hopefully never fire, but when it does, it will save many hours of debugging.

cartertinney: My understanding is that Paho only can't null the thread if loop_stop() is called from within the loop itself, and I've explicitly avoided doing that as a result. Are you saying there's another situation where that can happen?

v3_async_wip/v3_async_wip/mqtt_client.py
BertKleewein: Fair enough. I really hope that it's not necessary.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,455 @@
...
+                    interval = self._reconnect_interval
+                    logger.debug(
+                        "Reconnect Daemon reconnect attempt failed. Trying again in {} seconds".format(
+                            interval
+                        )
+                    )
+                    await asyncio.sleep(interval)
+        except asyncio.CancelledError:
+            logger.debug("Reconnect Daemon was cancelled")
+            current_attempt.cancel()
```
BertKleewein, line 267: `AttributeError: 'NoneType' object has no attribute 'cancel'`

v3_async_wip/v3_async_wip/mqtt_client.py
BertKleewein: No, I don't think it's worth it. It's not that much cleaner.

v3_async_wip/v3_async_wip/mqtt_client.py
BertKleewein: simple is good.

v3_async_wip/v3_async_wip/mqtt_client.py
BertKleewein: Any time `loop_forever` exits for any reason besides `loop_stop` being called. This is why you have to call `loop_stop` when there's an unexpected disconnect. Maybe it would be more robust to check `_thread.is_alive()` here and null the thread before calling `loop_start`. This way you don't have to rely on `on_disconnect` being called. This can happen if there is an exception inside `loop_forever`. Now that I think about it, we probably have some "client is bricked" bugs in v2 that could be fixed this way.

BertKleewein (approved):

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Should have seen this one coming tbh

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Fixed

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Fixed

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Added a check to the rc of loop_start

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Updated to use a Future instead of a Condition

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -0,0 +1,465 @@
...
+            raise ValueError("rc and message are mutually exclusive")
+        self.rc = rc
+        self.fatal = fatal
+        if rc:
+            message = mqtt.connack_string(rc)
+        super().__init__(message)
+
+
+class ConnectionLock(asyncio.Lock):
+    """Async Lock with additional attributes corresponding to the operation.
```
BertKleewein, line 48: > """Async Lock with additional attributes corresponding to the operation. [](http://example.com/codeflow?start=3&length=73)

I'm having trouble reasoning about the behavior of this object. It's like a critical section or a mutex, but it also contains the result of that operation (the future), but the result is only valid until the operation is complete (when the lock is released). So it's like a lock and an "on connection completed" object combined in one., except the "on connection completed" object is really limited in when you can use it.

PR #1088, opened 2022-12-17
[V3] Subscribe, Unsubscribe, Publish, Receive (MQTTClient)

cartertinney:
* Added subscribe, unsubscribe and publish APIs
* Added incoming message functionality
* Improved testing coverage


v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -68,8 +94,12 @@ def release(self):
         return super().release()
 
 
-class MQTTClient(object):
-    """Provides an async MQTT message broker interface."""
+class MQTTClient:
+    """
+    Provides an async MQTT message broker interface
+
+    This client currently only supports operations at a QoS (Quality of Service) of 1
```
anthonyvercolano, line 101: The requiremen of qos 1 is interesting as at least in node we specifically use qos 0 for twin ops.
 /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_004: [The `getTwin` method shall publish the request message with QoS=0, DUP=0 and Retain=0.]*/
    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_014: [The `updateTwinReportedProperties` method shall publish the request message with QoS=0, DUP=0 and Retain=0.]*/
    this._mqtt.publish(topic, body.toString(), { qos: 0, retain: false }, (err, puback) => {

cartertinney: That's very interesting. In V2 we did not meet this requirement then, as we've been using QoS=1 for everything.

PR #1089, opened 2023-01-05
Ok/stingy conn

olivakar:


samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+
+    async def enqueue_message(self):
+        message_id = 0
+        while True:
+            msg = Message("current wind speed ")
+            msg.message_id = message_id
+            msg.content_type = "application/json"
+            self.log_info_and_print("Created a message...")
+            self.message_queue.put_nowait(msg)
+            self.message_list[message_id] = msg
```
BertKleewein, line 165: >elf.message_list[message_id] = msg [](http://example.com/codeflow?start=13&length=34)

this list grows and never empties. What is it for?

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+        self.log_info_and_print("Current qsize is: {}".format(self.message_queue.qsize()))
+        while not self.message_queue.empty():
+            msg = await self.message_queue.get()
+            task = asyncio.create_task(self.device_client.send_message(msg))
+            self.task_list.append(task)
+            self.message_queue.task_done()
+        try:
+            if self.device_client.connected:
+                self.log_info_and_print("sending messages...")
+                await asyncio.gather(*self.task_list)
```
BertKleewein, line 184: >    await asyncio.gather(*self.task_list) [](http://example.com/codeflow?start=12&length=41)

What happens if one of these tasks raises an exception? The tasks all stay in `self.task_list`, but you don't try to re-send anything that failed.  I also think trying to gather the same task twice will raise an exception, so keeping the completed tasks for the next call to gather will cause gather to fail every time (after the first failure)

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+                self.log_info_and_print("sent all messages...")
+                self.task_list = []
+                await self.device_client.disconnect()
```
BertKleewein, line 187: >disconnect [](http://example.com/codeflow?start=41&length=10)

Don't you want to disconnect even when send_message fails?

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+                    try:
+                        self.log_info_and_print(
+                            "Attempting to connect the device client try number {}....".format(i)
+                        )
+                        await self.device_client.connect()
+                        self.log_info_and_print("Successfully connected the device client...")
+                        await self.do_all_tasks_and_disconnect()
+                        break
+                    except Exception as e:
+                        self.log_error_and_print("Caught exception while trying to connect...")
```
BertKleewein, line 208: >elf.log_error_and_print("Caught exception while trying to connect...") [](http://example.com/codeflow?start=25&length=70)

I recommend printing the type of exception.  Otherwise the log might not tell you what actually happened.

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
+# CUSTOMER PERSONA
```
BertKleewein, line 1: >CUSTOMER [](http://example.com/codeflow?start=2&length=8)

Don't forget copyright header.

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+                            self.log_error_and_print(
+                                "Failed to connect the device client due to retryable error."
+                                "Sleeping and retrying after some time..."
+                            )
+                            await asyncio.sleep(CONNECTION_ATTEMPT_INTERVAL)
+                        else:
+                            self.log_error_and_print(
+                                "Failed to connect the device client due to not-retryable error...."
+                            )
+                            raise
```
BertKleewein, line 224: >raise [](http://example.com/codeflow?start=28&length=5)

This task fails as soon as there is a non-retryable error. Is this what we want? I don't think customers would want the app to exit in this case.

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+
+        tasks = [
+            asyncio.create_task(self.connect_with_retry_send_all_and_disconnect()),
+            asyncio.create_task(self.enqueue_message()),
+            asyncio.create_task(self.memory_stats()),
+        ]
+
+        try:
+            done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
+            await asyncio.gather(*done)
```
BertKleewein, line 264: >await asyncio.gather(*done) [](http://example.com/codeflow?start=12&length=27)

What happens to pending tasks? From what I can tell, they keep running forever.

samples/solutions/stingy_connection.py
```diff
@@ -0,0 +1,275 @@
...
+                        await self.do_all_tasks_and_disconnect()
+                        break
+                    except Exception as e:
+                        self.log_error_and_print("Caught exception while trying to connect...")
+                        if type(e) is exceptions.CredentialError:
+                            self.log_error_and_print(
+                                "Failed to connect the device client due to incorrect or badly formatted credentials..."
+                            )
+                            raise
+                        if self.is_retryable(e):
```
BertKleewein, line 214: >f self.is_retryable(e): [](http://example.com/codeflow?start=25&length=23)

Does the customer want to retry connections? If it's expensive to be connected, should the code be doing this?

BertKleewein (changes requested):
:clock1:

samples/solutions/stingy_connection.py
olivakar: left over code

samples/solutions/stingy_connection.py
olivakar: not resending just clearing task list in case sending message runs into an exception

samples/solutions/stingy_connection.py
olivakar: yes i added that..i now disconnect regardless of whether message sending was successful or not

samples/solutions/stingy_connection.py
olivakar: added..left over from last sample

samples/solutions/stingy_connection.py
olivakar: i think i would have to discuss for this coz this is pretty much the same behavior i did for the other sample. If the error is not retry-able the customer probably has to change some configurational thing which could be credentials or something else. In that scenario I thought the best case is to exit the app and start again. In case our error situation become more fine tuned then we can reduce this list of non-retryable errors.

samples/solutions/stingy_connection.py
olivakar: I included wait for the pending tasks.

samples/solutions/stingy_connection.py
olivakar: I have thought about this. Every time the customer tries connection either the first time or after a period of disconnection the sample tries that connection only for 10 times and only for errors worth trying like a some temporary network issues. This can be changed but I did only include it for only 10 times which is few in number.

BertKleewein (approved):
:shipit:

PR #1091, opened 2023-01-11
[V3] Subscribe, Unsubscribe, Publish, Receive, Cancel (MQTTClient)

cartertinney:
* Added subscribe, unsubscribe and publish APIs
* Added incoming message functionality
* Improved testing coverage
* Added support for cancel to connect and disconnect


v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -329,73 +472,117 @@ async def connect(self):
...
+                functools.partial(
+                    self._mqtt_client.connect,
+                    host=self._hostname,
+                    port=self._port,
+                    keepalive=self._keep_alive,
+                ),
+            )
+            rc_msg = mqtt.error_string(rc)
+            logger.debug("Connect returned rc {} - {}".format(rc, rc_msg))
+        # TODO: more specialization of errors to indicate which are/aren't retryable
```
cartertinney, line 543: This can be done later - the infrastructure to support retryable vs nonretryable errors is implemented, I just haven't done an inventory of possible exceptions. I figure it will be quicker to do this while writing E2E tests rather than trying to do it now.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -404,32 +591,12 @@ async def disconnect(self):
         """
         # Wait for permission to alter the connection
         async with self._connection_lock:
-            loop = asyncio.get_running_loop()
-
-            if self._desire_connection:
-                # Record the type of operation in progress
-                self._connection_lock.connection_type = OP_TYPE_DISCONNECT
 
+            # TODO: why can't this be simplified to 'if self._network_loop'
```
cartertinney, line 595: It isn't immediately apparent, but these SHOULD be identical conditions, yet they aren't in testing. I will not check in until I get to the bottom of it (I suspect it's due to something with the test infrastructure)

cartertinney:
A few implementation notes:

1) Bert suggested something a while back about having an abstracted object that combined several related details (pending operations, state, etc.) that would make for simpler code. I agree that this is something I would like to implement, but for now I approached this implementation with a goal of avoiding anything that could end up being over-engineering, and kept things as simple as possible, even if it makes the code a bit more tricky to follow. The idea here is to keep maximum flexibility for now, and once things solidify a bit more and we have testing, I'll try and simplify/abstract

2) Tony pointed out that the spec for IoTHub says we should use QoS 0 for subscribe and unsubscribe. We never did this on V2, it completely ran on QoS 1, and for simplicity's sake, I kept that the same. There would be a bunch more research and experimentation required, plus some of the async logic would probably have to change to support QoS 0. Right now it just doesn't seem like a good use of development time. That said, I would love to get support for the full range of QoS at some point down the road.

3) I switched over to using `loop_forever` instead of `loop_start` and `loop_stop`. This change has some trade-offs. It makes certain things more complicated. Connect edge cases surrounding failure and cancellation, some of which slightly bleeds over to disconnect. It also definitely complicates the test infrastructure. But some things get a lot simpler - specifically, the complexity of state is massively reduced by not having to keep track of what's going on with the network loop, and managing it from several different places in the code. I suspect that over time I may want to switch back to `loop_start` and `loop_stop` once the changes from point 1 above are implemented, but for now I think that loop_forever is simpler and more flexible.

4) I used generators to receive messages because callbacks really ought not to be combined with async await. These generators kind of resemble the very old idea we had for V2 with inboxes, but are much, much simpler, and have more linguistic support. It's not demonstrated in the unit-tests due to the syntax making the assertions difficult, but you can iterate indefinitely over async generators which will be quite convenient when implementing the layer above this.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -13,11 +13,30 @@
...
 logger = logging.getLogger(__name__)
 
-OP_TYPE_CONNECT = "CONNECT"
-OP_TYPE_DISCONNECT = "DISCONNECT"
 
-RECONNECT_MODE_DROP = "RECONNECT_DROP"
-RECONNECT_MODE_ALL = "RECONNECT_ALL"
+# NOTE: Paho can return a lot of rc values. However, most of them shouldn't happen.
+# Here are the ones that we can expect for each method.
+expected_connect_rc = [mqtt.MQTT_ERR_SUCCESS]
```
BertKleewein, line 19: >expected_connect_rc [](http://example.com/codeflow?start=0&length=19)

Are these here for documentation only or do they serve a purpose? I would be worried about relying too much on this -- the word **shouldn't** in here is especially appropriate.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -121,13 +123,22 @@ def __init__(
         # Synchronization
         self.connected_cond = asyncio.Condition()
         self.disconnected_cond = asyncio.Condition()
-        self._connection_lock = ConnectionLock()
+        self._connection_lock = asyncio.Lock()
+        self._mid_tracker_lock = asyncio.Lock()
 
-        # Tasks
+        # Tasks/Futures
+        self._network_loop = None
```
BertKleewein, line 130: >_network_loop [](http://example.com/codeflow?start=13&length=13)

This might need a better name because of confusion with asyncio event loops.

BertKleewein:
> spec for IoTHub says we should use QoS 0 for subscribe and unsubscribe. 

Be careful with this one -- it might not mean what you think it means.  The QOS that you pass into `subscribe` is different than the QOS you pass into `publish`. `subscribve(topic, qos=1)` affects how the server publishes on `topic`. It does not cause SUBACK to be sent. There is no such thing as "using QOS 0 for unsubscribe".  

---
In reply to: [1379320762](https://github.com/Azure/azure-iot-sdk-python/pull/1091#issuecomment-1379320762) [](http://example.com/codeflow?ancestors=1379320762,1379320762)

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: I use these for tests, so that I can run the full range of expected values in addition to an unexpected value case

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -121,13 +123,22 @@ def __init__(
...
+        self._pending_connect = None
+        self._pending_subs = {}  # Map mid -> Future
+        self._pending_unsubs = {}  # Map mid -> Future
+        self._pending_pubs = {}  # Map mid -> Future
 
-        # Event Loop
-        self._loop = asyncio.get_running_loop()
+        # Incoming Data
+        self._incoming_messages = asyncio.Queue()
+        self._incoming_filtered_messages = {}  # Map topic -> asyncio.Queue
```
BertKleewein, line 141: > # Map topic -> asyncio.Queue [](http://example.com/codeflow?start=46&length=29)

We should consider using typings -- I think there are ways to slowly add them to an existing codebase and this seems like the right time to start adding them.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: How about instead of changing the name of the network loop (which is the correct name), I always refer to the event loop as the event loop instead?

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -329,73 +472,123 @@ async def connect(self):
...
+            # So chain a regular mqtt exception into a connection mqtt exception.
+            try:
+                raise MQTTError(rc=rc)
+            except MQTTError as e:
+                raise MQTTConnectionFailedError(message="Unexpected Paho .connect() rc") from e
+
+        # Start Paho network loop, and store the task. This task will complete upon disconnect
+        # whether due to invocation of .disconnect(), or an unexpected network drop.
+        #
+        # NOTE: If the connect attempt is cancelled, the network loop cannot be stopped.
```
BertKleewein, line 565: >NOTE: If the connect attempt is cancelled, the network loop cannot be stopped. [](http://example.com/codeflow?start=10&length=78)

"cannot be stopped" needs more detail.  Why can't it be stopped? Paho won't let it stop? We don't provide an API? We don't have a good place/time to stop it?  Some other reason?

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -329,73 +472,123 @@ async def connect(self):
...
+            logger.debug("Starting Paho network loop")
+            self._network_loop = self._loop.run_in_executor(None, self._mqtt_client.loop_forever)
+        else:
+            logger.debug(
+                "Paho network loop was already running. Likely due to a previous cancellation."
+            )
+
+        # The result of the CONNACK is received via this future stored on the lock
+        logger.debug("Waiting for connect response...")
+        rc = await self._pending_connect
```
BertKleewein, line 583: >  rc = await self._pending_connect [](http://example.com/codeflow?start=6&length=34)

Returning an RC isn't really very pythonic. Wouldn't it make more sense to have `await _self._pending_connect` raise an error if something goes wrong? 

Also, the comment above on "future stored on this lock" is outdated.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -438,28 +615,191 @@ async def wait_for_disconnect():
...
                     # In this implementation, it should only happen if Paho's inner state
                     # indicates we would like to be connected, but we actually aren't.
                     # We still want to do this disconnect however, because doing so changes
                     # Paho's state to indicate we no longer wish to be connected.
                     logger.debug("Early disconnect return (Already disconnected)")
-                    disconnect_done.cancel()
+                    if self._network_loop_running():
+                        # This block should never execute
+                        logger.warning("Network loop unexpectedly still running. Waiting")
+                        await self._network_loop
```
BertKleewein, line 647: This feels like you're taking something that _shouldn't_ happen and you're turning it into a hang. If the network loop is running, and we don't think it should be running, why do we think it's going to stop now?

cartertinney: I agree. There's defensive coding, and then there's inventing problems to be worried about. I'll remove this

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -239,32 +249,88 @@ async def set_disconnected():
...
+                async with this._mid_tracker_lock:
+                    try:
+                        f = this._pending_subs[mid]
+                        f.set_result(True)
+                    except KeyError:
+                        logger.warning("Unexpected SUBACK received for mid {}".format(mid))
 
-        def on_subscribe(client, userdata, mid):
-            pass
+            asyncio.run_coroutine_threadsafe(complete_sub(), this._loop)
```
BertKleewein, line 284: >asyncio.run_coroutine_threadsafe(complete_sub(), this._loop) [](http://example.com/codeflow?start=12&length=60)

It's super important that you don't call result() on the Future that is returned.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -438,28 +615,191 @@ async def wait_for_disconnect():
...
+        :param str topic: a single string specifying the subscription topic to subscribe to
+
+        :raises: ValueError if topic is None or has zero string length.
+        :raises: MQTTError if there is an error subscribing
+        """
+        try:
+            mid = None
+            logger.debug("Attempting subscribe")
+            # Using this lock allows us to be sure that the ACK won't come in before the
+            # Future can be added to the pending dictionary
```
BertKleewein, line 672: I love how you did this, but this comment is a little misleading. Maybe better to say "This lock postpones the code inside `complete_sub` from setting the subscribe result until after we can add the future to the mid tracker". Or something like that.  The important part is that this code works in concert with `complete_sub` and that's not particularly obvious.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -313,6 +386,76 @@ def set_credentials(self, username, password=None):
...
+        :raises: ValueError if a filter is already applied for the topic
+        """
+        if topic in self._incoming_filtered_messages:
+            raise ValueError("Filter already applied for this topic")
+
+        # Add a Queue for this filter
+        self._incoming_filtered_messages[topic] = asyncio.Queue()
+
+        # Define a callback that uses the filter queue
+        self_weakref = weakref.ref(self)
```
BertKleewein, line 404: I don't think this weakref stuff is still necessary, is it?

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -438,28 +615,191 @@ async def wait_for_disconnect():
...
+            else:
+                logger.debug("Unsubscribe was cancelled before mid was assigned")
+            raise
+        finally:
+            # Delete any pending operation (if it exists)
+            async with self._mid_tracker_lock:
+                if mid and mid in self._pending_unsubs:
+                    del self._pending_unsubs[mid]
+
+    async def publish(self, topic, payload):
```
BertKleewein, line 745: Fun fact: If you call Paho's publish method with Qos=0, `on_publish` will still get called. It just gets called automatically by Paho instead of having to wait for the PUBACK.  

Maybe we should add a Qos parameter here while we're building things up. I've had partners ask to publish as Qos 0, and now might be a good time to add it.

v3_async_wip/tests/simple_e2e.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Do it twice to make sure it's repeatable
+    await conn_disconn()
+    await asyncio.sleep(1)
+    await conn_disconn()
+
+
+@pytest.mark.it("Queued connects and disconnects")
+async def test_queued_connects_and_disconnects(client):
+    # TODO: can this be made to show they all happened in order?
```
BertKleewein, line 110: >can this be made to show they all happened in order? [](http://example.com/codeflow?start=12&length=52)

You could assert that they run _serially_, but anything more specific would mean relying on undefined behavior.

BertKleewein:
> def mock_paho(mocker, paho_threadpool):

My 2 cents: This kind of mock is OK to begin with, but becomes much harder to maintain as time goes by. Since changes to this mock affect every test in here, you end up getting in situations where you can't change the Paho mock for one test without breaking other tests. As you maintain it, you need to put more and more effort into making mock_paho look and act like real paho because you can't let it diverge for fear of over-fitting your code to mock_paho instead of real Paho.  

This is purely preference and I don't recommend changing this. If it were me, I would make the smallest possible Paho mock and then add specific behaviors for tests that require them. 

---
Refers to: v3_async_wip/tests/test_mqtt_client.py:58 in 1d51862. [](commit_id = 1d51862e4ea98b11efefae1f3a85c4afb2a0fd66, deletion_comment = False)

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: It's because when using `loop_forever()`, Paho is in charge of the loop's lifecycle. The loop will exit when a connection is lost, whether due to manual disconnect, drop, or even connection failure. There is no way to manually stop it.

I will add this detail to the inline implementation note.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Good catch on the lock stuff.

As far as the rc, I disagree - we use rc throughout the mqtt_client implementation, I see no reason to inject a spurious layer of abstraction between returning the result from the callback. As you can see above on the .connect() invocation, we are already directly using rcs in this method, so it's more consistent.

My view is that within this class that directly deals with Paho, we should be using rcs, since they are returned by Paho, and we have to just deal with that. But they should never, ever make it outside of the MQTTClient.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: That sounds like it should make a more full QoS support less painful than expected, but nonthless I think that's best for a separate PR.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Last time I checked early in V3 it was still necessary under some circumstances. That said, it's worth revisiting once I have a full E2E suite set up.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Yeah, you're right, I ought to improve the clarity

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: Yeah, I should have documented that, good call.

v3_async_wip/v3_async_wip/mqtt_client.py
cartertinney: I'll put a separate PR for this out tomorrow

PR #1092, opened 2023-01-14
[V3] Added typing hints

cartertinney:
* Added typings to MQTTClient
* Removed weakref implementation for simplicity. We can add it back if necessary later.
* Added mypy as a dev dependency
* Added pre-commit hook


BertKleewein (approved):
:shipit:

PR #1096, opened 2023-01-30
[V3] Initial IoTHubMQTTClient

cartertinney:
* Implemented all sending and receiving of data
* Partial SasToken renewal implementation

A couple items are not yet finished:
* Full SasToken renewal implementation + tests
* Decoding Message property tests
* Notifications (i.e. on_connect, on_disconnect, etc.)

These leftover  items will be addressed in some smaller, follow-up PRs. I need to move some things around, and add typings to legacy code to make them work correctly, and that's too much for this PR.


migration_guide.md
```diff
@@ -30,6 +30,8 @@ Note also that this change does *not* affect automatic reconnection attempts in
 ## Receiving data from IoTHub
 Similarly to the above, there is an additional explicit step you must now make when trying to receive data. In addition to setting your handler, you must explicitly start/stop receiving. Note also that the above step of manually connecting must also be done before starting to receive data.
 
+Furthermore, note that the content of the message is now referred to by the 'payload' attribute on the message, rather than the 'data' attribute.
```
BertKleewein, line 33: <disapproving glare> This seems somewhat arbitrary. What's the benefit of this change?

dev_utils/dev_utils/custom_mock.py
```diff
@@ -0,0 +1,49 @@
...
+#         await self._stop_hanging.wait()
+
+#     async def wait_for_hang(self):
+
+
+#     def is_hanging(self):
+#         return self._is_hanging
+
+#     def stop_hanging(self):
+#         self._stop_hanging.set()
```
BertKleewein, line 49: remove this?

v3_async_wip/v3_async_wip/custom_typing.py
```diff
@@ -0,0 +1,16 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from typing import Union, Dict, List, Tuple, Any
+
+# typing does not support recursion, so unfortunately, we have to use Any for elements of
```
BertKleewein, line 9: >typing does not support recursion, so [](http://example.com/codeflow?start=2&length=37)

Could you use a forward reference? I don't think it would fix all problems, but it would make the type more accurate.  [https://peps.python.org/pep-0484/#forward-references](https://peps.python.org/pep-0484/#forward-references)

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+            MethodRequest, None
+        ] = _create_method_request_generator(self._mqtt_client)
+        self.incoming_twin_patches: AsyncGenerator[TwinPatch, None] = _create_twin_patch_generator(
+            self._mqtt_client
+        )
+
+        # Internal request/response infrastructure
+        self._request_ledger = rr.RequestLedger()
+        self._twin_responses_enabled = False
+        self._twin_response_listener = asyncio.create_task(self._process_twin_responses())
```
BertKleewein, line 86: >_twin_response_listener [](http://example.com/codeflow?start=13&length=23)

What happens if this task raises an exception?

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+
+        # TODO: do we need to track what features are enabled?
+        # I don't think so, but check what happens on double subscribe
+
+    def _create_token_update_alarm(self) -> alarm.Alarm:
```
BertKleewein, line 91: >_create_token_update_alarm [](http://example.com/codeflow?start=8&length=26)

You can get rid of the update timer if you just put a call to `self._sastoken.refresh()` in the connect code. This is how java and C# do it.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+                "Waiting to receive twin from IoTHub... (rid: {})".format(request.request_id)
+            )
+            response = await request.get_response()
+        except asyncio.CancelledError:
+            logger.debug(
+                "Attempt to get twin from IoTHub was cancelled while waiting for a response. If the response arrives, it will be discarded (rid: {})".format(
+                    request.request_id
+                )
+            )
+            await self._request_ledger.delete_request(request.request_id)
```
BertKleewein, line 341: >wait self._request_ledger.delete_request(request.request_id) [](http://example.com/codeflow?start=13&length=60)

I think you want to call `delete_request` on other exceptions too.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+                )
+            )
+            response = await request.get_response()
+        except asyncio.CancelledError:
+            logger.debug(
+                "Attempt to send twin patch to IoTHub was cancelled while waiting for response. If the response arrives, it will be discarded (rid: {})".format(
+                    request.request_id
+                )
+            )
+            await self._request_ledger.delete_request(request.request_id)
```
BertKleewein, line 277: > await self._request_ledger.delete_request(request.request_id) [](http://example.com/codeflow?start=11&length=62)

Same comment as above. You should probably remove the request on all exceptions.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+) -> AsyncGenerator[Message, None]:
+    c2d_msg_topic = mqtt_topic.get_c2d_topic_for_subscribe(device_id)
+    mqtt_msg_generator = mqtt_client.get_incoming_message_generator(c2d_msg_topic)
+
+    async def c2d_message_generator(
+        incoming_mqtt_messages: AsyncGenerator[mqtt.MQTTMessage, None]
+    ) -> AsyncGenerator[Message, None]:
+        async for mqtt_message in incoming_mqtt_messages:
+            # TODO: decode differently depending on encoding type from topic
+            c2d_message = Message(mqtt_message.payload.decode("utf-8"))
```
BertKleewein, line 565: >c2d_message = Message(mqtt_message.payload.decode("utf-8")) [](http://example.com/codeflow?start=12&length=59)

If payload.decode raises an exception, what happens with this generator? Does it continue to function or do we need to somehow make a new one?

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -0,0 +1,627 @@
...
+        client_id=client_id,
+        query_params=urllib.parse.urlencode(query_param_seq, quote_via=urllib.parse.quote),
+    )
+    return username
+
+
+def _create_c2d_message_generator(
+    device_id: str, mqtt_client: mqtt.MQTTClient
+) -> AsyncGenerator[Message, None]:
+    c2d_msg_topic = mqtt_topic.get_c2d_topic_for_subscribe(device_id)
```
BertKleewein, line 557: >2d_msg_topic = mqtt_topic.get_c2d_topic_for_subscribe(device_id) [](http://example.com/codeflow?start=5&length=64)

Input messages and C2D are mutually exclusive. You can't do input messages without a module_id, an you can't do c2d _with_ a module_id.

BertKleewein (approved):
:shipit:

migration_guide.md
cartertinney: Consistency. We call it a 'payload' on other objects. Ryan was on board with this change, but we can add some backwards compatibility support to it as well.

requirements_test.txt
```diff
@@ -1,9 +1,11 @@
 pytest
-pytest-mock
+pytest-mock>=3.10.0
 pytest-asyncio>=0.17
 pytest-testdox>=1.1.1
 pytest-cov
 pytest-timeout
+pytest-lazy-fixture
+mock                    # Need to use instead of builtin for backports to 3.7
```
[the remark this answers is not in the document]
cartertinney: No, it's probably just a good idea in general to use the independent mock library so we have a more consistent set of tools across versions of Python.

dev_utils/dev_utils/custom_mock.py
cartertinney: Whoops

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: It will leverage the not-yet-existing infrastructure for background error reporting (a couple PRs from now). Will add a TODO comment to reflect this so it doesn't get forgotten.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Not all sastokens are refresh-able, i.e. NonRenewableSasToken.

Potentially with a different implementation of that area we could leverage this though. I'll think on it.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: If we add error handling logic in there, it'll continue to function

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Will add this pending further discussion

PR #1098, opened 2023-02-07
[V3] Enhanced reused modules for V3

cartertinney:
* Ported user_agent.py and added typings for V3 
* Ported, added typings and simplified mqtt_topic_iothub.py for V3
* Enhanced Message class to make properties and topics easier to use internally.
* Enhanced Message-related logic to be more rigid about expected typings
* Added full support for JSON payloads when sending/receiving messages
* APIs related to Device/Module-specific features now throw exceptions if not using the correct configuration

### DIFF
--- a/v3_async_wip/tests/test_iothub_mqtt_client.py
+++ b/v3_async_wip/tests/test_iothub_mqtt_client.py
@@ -731,7 +731,7 @@ class TestIoTHubMQTTClientSendMessage:
         return models.Message("some payload")
 
     @pytest.mark.it(
-        "Awaits a publish to the telemetry topic using the MQTTClient, sending the given Message's JSON-converted payload"
+        "Awaits a publish to the telemetry topic using the MQTTClient, sending the given Message's payload, formatted according to its content type and content encoding properties"
     )
     @pytest.mark.parametrize(
         "device_id, module_id",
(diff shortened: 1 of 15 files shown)


v3_async_wip/v3_async_wip/models.py
```diff
@@ -74,26 +86,72 @@ def set_as_security_message(self) -> None:
...
+            d["$.exp"] = self.expiry_time_utc
+        if self.user_id:
+            d["$.uid"] = self.user_id
+        if self.correlation_id:
+            d["$.cid"] = self.correlation_id
+        return d
 
-class MethodRequest(object):
+    @classmethod
+    # TODO: should this just replace the __init__?
```
cartertinney, line 118: Thoughts?

v3_async_wip/v3_async_wip/config.py
```diff
@@ -12,6 +12,7 @@
 
 # TODO: add typings for imports
 # TODO: update docs to ensure types are correct
+# TODO: can these just be TypeDicts?
```
BertKleewein, line 15: >TypeDicts [](http://example.com/codeflow?start=26&length=9)

This would be awesome.  Or DataClass objects

v3_async_wip/v3_async_wip/mqtt_topic_iothub.py
```diff
@@ -0,0 +1,284 @@
...
+# NOTE (Oct 2020): URL encoding policy is currently inconsistent in this module due to restrictions
+# with the Hub, as Hub does not do URL decoding on most values.
+# (see: https://github.com/Azure/azure-iot-sdk-python/wiki/URL-Encoding-(MQTT)).
+# Currently, as much as possible is URL encoded while keeping in line with the policy outlined
+# in the above linked wiki article. This is to say that Device ID and Module ID are never
+# encoded, however other values are. By convention, it's probably fine to be encoding/decoding most
+# values that are not Device ID or Module ID, since it won't make a difference in production as
+# the narrow range of acceptable values for, say, status code, or request ID don't contain any
+# characters that require URL encoding/decoding in the first place. Thus it doesn't break on Hub,
+# but it's still done here as a client-side best practice - Hub will eventually be doing a new API
```
BertKleewein, line 33: >will [](http://example.com/codeflow?start=64&length=4)

nit: _may_ be doing a new API. ;)

v3_async_wip/v3_async_wip/models.py
BertKleewein: Yeah, maybe there's an advantage, but it feels like there are bigger fish to fry.

v3_async_wip/tests/test_iothub_mqtt_client.py
```diff
@@ -731,8 +729,89 @@ def message(self):
...
+            pytest.param(FAKE_DEVICE_ID, FAKE_MODULE_ID, id="Module Configuration"),
+        ],
     )
+    @pytest.mark.parametrize("content_encoding", ["utf-8", "utf-16", "utf-32"])
+    @pytest.mark.parametrize(
+        "content_type, payload, expected_str_payload",
+        [
+            pytest.param("text/plain", "some_text", "some_text", id="text/plain"),
+            pytest.param(
+                "application/json", {"some": "json"}, '{"some": "json"}', id="application/json"
```
BertKleewein, line 779: >"application/json", [](http://example.com/codeflow?start=16&length=19)

dumb corner case that is easy to screw up/break. You can json encode a string, and you get a string with that string in quotes.

cartertinney: Not actually a corner case - that's desired behavior and is covered a little bit farther down in a separate test

BertKleewein (approved):
:shipit:

PR #1103, opened 2023-02-17
[V3] SAS Auth Revisions

cartertinney:
* Restructured SAS-related objects and their responsibilities
* Removed references to "renewing" a SAS token due to bad semantics
* Keeping a valid SAS token is no longer the domain of the IoTHubMQTTClient
* Made signing data with a SigningMechanism asynchronous (in signature) - actual async implementation will come later.


v3_async_wip/v3_async_wip/sastoken.py
```diff
@@ -5,173 +5,222 @@
...
-        key_name: Optional[str] = None,
-        ttl: int = 3600,
-    ) -> None:
-        """
-        :param str uri: URI of the resource to be accessed
-        :param signing_mechanism: The signing mechanism to use in the SasToken
-        :type signing_mechanism: Child classes of :class:`azure.iot.common.SigningMechanism`
-        :param str key_name: Symmetric Key Name (optional)
-        :param int ttl: Time to live for the token, in seconds (default 3600)
+class SasTokenGenerator:
```
cartertinney, line 58: Ultimately, I did end up deciding to go with an object instead of a function for this, even though it's a little more bloated. The reasoning was that I just felt the functional programming style implementation clashed with everything else being more object oriented. I think having a consistent style is worth the tradeoff of being slightly more bulky than is strictly necessary

v3_async_wip/v3_async_wip/sastoken.py
```diff
@@ -5,173 +5,222 @@
...
+            expiry=str(expiry_time),
+        )
+        return SasToken(token_str)
 
-    @property
-    def expiry_time(self) -> int:
-        """Expiry Time is READ ONLY"""
-        return self._expiry_time
 
+class ExternalSasTokenGenerator(SasTokenGenerator):
```
cartertinney, line 94: I'm not thrilled with this "inheriting" from SasTokenGenerator, and then overriding everything. It's mostly for typing - ideally SasTokenGenerator and ExternalSasTokenGenerator would both inherit from an abstract class, which ideally would be called SasTokenGenerator for good typehint semantics, but then what does that leave from us to call the basic child generator class that uses a signing mechanism? SigningMechanismSasTokenGenerator? That's bad.

Open to ideas.

v3_async_wip/v3_async_wip/sastoken.py
```diff
@@ -5,173 +5,222 @@
...
+        except Exception as e:
+            raise SasTokenError("Unable to generate SasToken") from e
 
-    Data Attributes:
-    expiry_time (int): Time that token will expire (in UTC, since epoch)
-    resource_uri (str): URI for the resource the Token provides authentication to access
-    """
 
-    def __init__(self, sastoken_string) -> None:
+class SasTokenProvider:
```
cartertinney, line 118: I had talked about calling this a SasTokenWarden, but ultimately, I stuck with SasTokenProvider due to semantics - while I think Warden is a better name for what it does, from the perspective of something using this class, Warden is meaningless, and reduces clarity. A user of this class uses it to provide the token - whatever the wholistic picture of what is happening inside is irrelevant.

v3_async_wip/v3_async_wip/edge_hsm.py
```diff
@@ -47,7 +47,8 @@ def __init__(
         self.generation_id = generation_id
         self.workload_uri = _format_socket_uri(workload_uri)
 
-    def get_certificate(self) -> str:
+    # TODO: Use async http to make use of this being a coroutine
```
BertKleewein, line 50: >   # TODO: Use async http to make use of this being a coroutine [](http://example.com/codeflow?start=1&length=63)

It might be an extra hassle because of the unix sockets code we use for edge.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -153,19 +153,19 @@ async def shutdown(self) -> None:
...
+            logger.debug("Cancelling 'keep_credentials_fresh' background task")
+            self._keep_credentials_fresh_bg_task.cancel()
+            cancelled_tasks.append(self._keep_credentials_fresh_bg_task)
+
         # Wait for the cancellation to complete before returning
-        try:
-            await self._twin_response_listener
-        except asyncio.CancelledError:
-            pass
+        await asyncio.gather(*cancelled_tasks, return_exceptions=True)
```
BertKleewein, line 168: > await asyncio.gather(*cancelled_tasks, return_exceptions=True) [](http://example.com/codeflow?start=7&length=63)

suggest adding a timeout. The cancel should be almost guaranteed, but I'm not sure if it is.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -153,19 +153,19 @@ async def shutdown(self) -> None:
...
         await self.disconnect()
-        # Cancel the SAS token update alarm. Note that this is not a task, it's a threaded Alarm.
-        # No need to wait for the result.
-        if self._sastoken_update_alarm:
-            logger.debug("Cancelling SAS Token update alarm")
-            self._sastoken_update_alarm.cancel()
-        # Cancel and wait for the completion of the twin response task
-        logger.debug("Cancelling twin response listener")
-        self._twin_response_listener.cancel()
+        cancelled_tasks = []
```
BertKleewein, line 156: >     cancelled_tasks = [] [](http://example.com/codeflow?start=3&length=25)

Recommend wrapping self.disconnect in a try/catch block.

v3_async_wip/v3_async_wip/sastoken.py
BertKleewein: I've seen the convention where SasTokenGenerator is the interface and SasTokenGeneratorImpl is the implementation of that interface. We don't have to use this everywhere but maybe here it makes sense.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Timeouts aren't really a thing with coroutines or gather, and trying to implement one on top would result in additional complexity and potentially problematic states. As long as the coroutine isn't hanging on a synchronous function and doesn't catch the CancelledError without re-raising (which we can establish via testing), there's no way for cancel to fail.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Disconnect can't fail. I could add one for safety, I suppose, but it would be purely defensive.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Timeouts aren't properly supported with async until Python 3.11 - the idea really is that coroutines are supposed to just be cancellable. The only reasons a cancel wouldn't work is if the CancelledError is suppressed, or it's hanging on a synchronous call.

We can test to ensure that CancelledError doesn't get suppressed. Hanging on a synchronous call could be possible I suppose if the customer is using custom SAS token injection via synchronous callback, but we could even get around that by running their callbacks in a separate thread. Although that all being said, if a synchronous call is hanging, it's eating the entire event loop anyway, and you wouldn't even be able to invoke .shutdown() in the first place. So I suppose that's a moot point entirely.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -79,40 +82,21 @@ def __init__(
...
-                try:
-                    self._sastoken.refresh()
-                    logger.debug("SAS Token renewal succeeded")
-                except st.SasTokenError:
-                    logger.error("SAS Token renewal failed")
-                    # TODO: background exception?
-                # With the token renewed, now set a new Alarm
-                self._sastoken_update_alarm = self._create_token_update_alarm()
 
+        # Background Tasks
```
BertKleewein, line 86: ># Background Tasks [](http://example.com/codeflow?start=8&length=18)

I thought more about background exceptions. I don't think we need to invent our own mechanism for this. If we catch it and send it to our own "background exception handler", two things _don't_ happen:

1. Developers can't use "break on unhandled exceptions" in the debugger. Yes, if they know about the background exception handler, they can put code there, but what if they don't know about this? In V2, background exceptions were almost impossible to debug even if you knew about the background exception handler. All of the context was lost at that point -- it's way better to have the debugger break at the point of failure rather than inside some handler. 

2. `sys.execpthook` doesn't get called.
We had some github issue where the customer was talking about "uncatchable exceptions", and I think this is what they were talking about. I'd rather not have to explain why our exceptions are somehow different than all other exceptions because we have a special way for customers to find out about them.

I think the better way to architect things like this is to let the task fail and then recover afterwards. This way, the background exception is treated like all other exceptions, but we still have the benefit of being able to continue working.  Maybe `add_done_callback` can be used here?

v3_async_wip/v3_async_wip/edge_hsm.py
cartertinney: We'd have to approach it differently, I don't think the monkeypatch would work (although it might). But manually using a unixsocket Session object isn't a big hassle IMO. That said, to make sure this doesn't break, it's probably a good idea to wait to try and change this until the E2E tests are hooked up.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
BertKleewein: but it can fail, it calls logger.debug and it calls into Paho, and either of those could raise an exception for any number of reasons that are beyond our control.  Yes, unlikely, but this function gets called when things are getting torn down or things are broken, so it's hard to say what can fail and what can't. 

Normally I wouldn't worry about being defensive like this, but `shutdown` is special.  It's an "I need to get out of here" function, and we don't want it to fail. It's like the "power off" button on your phone. In normal situations, nothing should go wrong with powering off your phone, but you don't want the "power off" button to fail and leave your phone partially powered on with no recourse.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Yean we could probably do something like that. I'd have to explore it a bit more, this isn't something I've thought too much about yet, but resetting the task in a callback is an interesting idea.

That said, without piping them up, there's no way to actually catch them. A debugger could break in a background task, but we have no means to throw the exception without some kind of background exception handler. Then again, I'm really not sure what utility there really is in getting the exceptions directly to the user - what are they really supposed to do with that information dynamically at runtime? You can't realistically "respond" to them, because they didn't come from your inputs or actions. Either you injected bad code that is causing failure (that's not a dynamic change you can address) or you're receiving bad data from IoTHub/Edge/some other Module (again, not something you can address dynamically at runtime)

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Fair enough on the point about logging, I didn't consider the implications of that. So would you like to see it suppress the error completely, or cache the error, continue trying to shut other things down, and then raise the error at the end?

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
BertKleewein: I don't know if I would go so far as to say "aren't really a thing".  `asyncio.wait_for` is a thing, and `asyncio.wait` has a timeout parameter. 3.11 added `asyncio.timeout` as a context manager object,. 

Your comment about "hanging on a synchronous function" is really important here because shutdown is a "last resort" function when things are going wrong, like powering off your phone or a three-fingered-salute. You don't want it to hang and you don't want it to partially complete.

If these tasks haven't completed after, say, 10 seconds, you don't have to worry about problematic states. You are already in a problematic state. Your best bet is to return and get out of the way and hope things don't get worse.

v3_async_wip/v3_async_wip/sastoken.py
BertKleewein: It wasn't officially added until 3.8, but if it's mostly for Typing, then maybe it should be a Protocol instead of an Interface.

BertKleewein (approved):
:shipit:

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
BertKleewein: Good question. I was thinking we should suppress, but re-raising at the end is an interesting idea. Not sure which is better though.

PR #1108, opened 2023-03-09
[V3] Generator and background task exception handling

cartertinney:
* Formalized incoming data generators as properties and refactored implementation
* Formalized C2D messages as being Device exclusive
* Added suppression for errors in incoming data generators
* Added suppression for errors in background tasks
* Added robust error handling tests for those generators and tasks

### DIFF
--- a/v3_async_wip/tests/test_iothub_mqtt_client.py
+++ b/v3_async_wip/tests/test_iothub_mqtt_client.py
@@ -665,7 +665,7 @@ class TestIoTHubMQTTClientInstantiation:
 
         await client.shutdown()
 
-    # NOTE: For testing the functionality of this task, see the corresponding test suite (TestIoTHubMQTTClientSasTokenUpdate)
+    # NOTE: For testing the functionality of this task, see the corresponding test suite (TestIoTHubMQTTClientKeepCredentialsFresh)
     @pytest.mark.it(
         "Begins running the ._keep_credentials_fresh() coroutine method as a background task, storing it as an attribute, if using SAS authentication"
     )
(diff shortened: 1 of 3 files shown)


v3_async_wip/tests/test_iothub_mqtt_client.py
```diff
@@ -2478,35 +2699,189 @@ async def test_message_property_defaults(self, client):
...
+        mqtt_msg2.payload = '{"json": "in", "a": {"different": {"string": "format"}}}'.encode(
+            "utf-8"
+        )
+
+        # Load the MQTTMessages into the MQTTClient's filtered message queue
+        await client._mqtt_client._incoming_filtered_messages[generic_topic].put(mqtt_msg1)
+        await client._mqtt_client._incoming_filtered_messages[generic_topic].put(mqtt_msg2)
+
+        # Get items from generator
+        mreq1 = await client.incoming_direct_method_requests.__anext__()
```
olivakar, line 2884: will this not work ?
mreq1 = await anext(client.incoming_direct_method_requests)

olivakar (approved):

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -101,41 +136,78 @@ async def _enable_twin_responses(self) -> None:
...
-                new_sastoken = await self._sastoken_provider.wait_for_new_sastoken()
-                logger.debug("New SAS Token available, updating MQTTClient credentials")
-                self._mqtt_client.set_credentials(self._username, str(new_sastoken))
-                # TODO: should we reconnect here? Or just wait for drop?
+                try:
+                    logger.debug("Waiting for new SAS Token to become available")
+                    new_sastoken = await self._sastoken_provider.wait_for_new_sastoken()
+                    logger.debug("New SAS Token available, updating MQTTClient credentials")
+                    self._mqtt_client.set_credentials(self._username, str(new_sastoken))
+                    # TODO: should we reconnect here? Or just wait for drop?
```
lev-i, line 200: Why not reconnect if the new credentials are available and updated?

v3_async_wip/tests/test_iothub_mqtt_client.py
```diff
@@ -2261,8 +2313,156 @@ async def test_message_property_defaults(self, client):
         assert msg.content_type == "text/plain"
         assert msg.content_encoding == "utf-8"
 
+    @pytest.mark.it(
+        "Suppresses any unexpected exceptions raised while extracting the message properties from the MQTTMessage, dropping the MQTTMessage and continuing"
```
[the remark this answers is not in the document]
cartertinney: No, it's a background exception, we suppress it so our background tasks don't crash and brick the client. Same principle as in V2, but different implementation due to async

v3_async_wip/tests/test_iothub_mqtt_client.py
cartertinney: It would in most versions of Python, but not 3.7

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Complexity. I think we probably are going to do a manual reconnect here, but it doesnt' have to be part of the PR

PR #1110, opened 2023-03-09
[V3] Simplified shutdown

cartertinney:
* Simplified `.shutdown()` implementations with use of `asyncio.shield()`
* Marked tests where some more complex mocking is required to truly demonstrate the accuracy of shutdown (low priority for now)
* Added try/finally blocks to tests that could hang upon assertion failure due to the mocking of methods that are required for teardown

### DIFF
--- a/v3_async_wip/v3_async_wip/iothub_client.py
+++ b/v3_async_wip/v3_async_wip/iothub_client.py
@@ -4,10 +4,11 @@
 # license information.
 # --------------------------------------------------------------------------
 import abc
+import asyncio
 import logging
 import os
 import ssl
-from typing import Optional, Union, cast
+from typing import Optional, cast
 from .custom_typing import FunctionOrCoroutine
 from .iot_exceptions import IoTEdgeEnvironmentError
 from . import config, edge_hsm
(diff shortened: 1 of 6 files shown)


v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -173,14 +160,14 @@ async def shutdown(self) -> None:
...
 
-        # Wait for the cancellation to complete before returning
-        # NOTE: If cancelled while awaiting here, all tasks in gather will still be cancelled
-        # because the cancellations have already been issued.
-        # NOTE: Also, cancelling a gather implicitly cancels all the tasks that are gathered anyway
-        await asyncio.gather(*cancelled_tasks, return_exceptions=True)
-
-        if cached_exception:
-            raise cached_exception
+        results = await asyncio.gather(
```
olivakar, line 163: this can be done with asyncio.wait also right ? shield would not have been needed then.
any reason you chose this patten ?

olivakar (approved):

v3_async_wip/v3_async_wip/iothub_client.py
```diff
@@ -50,46 +51,17 @@ async def shutdown(self) -> None:
...
-
-        logger.debug("IoTHubClient shutdown procedure complete")
-        if cached_exception:
-            # NOTE: In the case of multiple failures, only the last one gets raised.
-            # Not much way around it, and besides, this is all an extreme edge case anyway.
-            logger.warning(
-                "Raising previously suppressed error now that shutdown procedure is complete"
-            )
-            raise cached_exception
+            operations.append(asyncio.shield(self._sastoken_provider.shutdown()))
```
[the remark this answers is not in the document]
cartertinney: The granularity isn't really important now that we're relying on built-in language constructs

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: .wait only handles one awaitable at a time. It's also a lower level primitive, so I use a higher level abstraction when possible to benefit from higher level cancellation and error handling

PR #1111, opened 2023-03-15
[V3] Removed iothub_client.py + Object lifecycle improvements

cartertinney:
* Objects no longer implicitly start background tasks upon instantiation, they now implement `.start()` and `.stop()` methods
* Removed `iothub_client.py` and associated tests in advance of the addition of `iothub_session.py` in the next PR


v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -214,23 +199,46 @@ async def _keep_credentials_fresh(self) -> None:
                 logger.error("No SasTokenProvider. Cannot update credentials")
                 break
 
-    async def shutdown(self) -> None:
-        """Shut down the client.
+    async def start(self) -> None:
```
BertKleewein, line 202: >start [](http://example.com/codeflow?start=14&length=5)

start and stop are both idempotent. Is this by coincidence or by design? At the very least, these should be documented as "only call once" or "can be called multiple times". I'm not sure if there's any value in doing more than this (like adding artificial enforcement of non-idempotency).

BertKleewein (approved):
:shipit:

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
cartertinney: Have formalized docs and added tests to illustrate the by-design-ness of this.

PR #1112, opened 2023-03-21
[V3] Initial IoTHubSession

cartertinney:
* Added initial implementation of IoTHubSession
* Blob upload APIs are still pending discussion and have been excluded
* Edge related APIs are still pending discussion and have been excluded
* Auto-reconnect is currently disabled pending investigation


v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -0,0 +1,284 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import asyncio
+import contextlib
+import ssl
+from typing import Optional, Union, AsyncGenerator, Type
+from types import TracebackType
+
+from v3_async_wip import signing_mechanism as sm
```
olivakar, line 12: so i don't always like so short names so something like sign_mech and conn_str but it is upto u

v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
+            self._sastoken_provider = None
+
+        # Create a default SSLContext if not provided
+        if not ssl_context:
+            ssl_context = _default_ssl_context()
+
+        # Instantiate the MQTTClient
+        cfg = config.IoTHubClientConfig(
+            hostname=hostname,
```
BertKleewein, line 80: Nit: inconsistent naming -- you're shortening the variable name instead of the module name.

v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -0,0 +1,284 @@
...
+        traceback: TracebackType,
+    ) -> None:
+        # NOTE: these are not supposed to be able to fail so no error handling required
+        await self._mqtt_client.disconnect()
+        await self._stop_all()
+
+    async def _stop_all(self) -> None:
+        await self._mqtt_client.stop()
+        if self._sastoken_provider:
+            await self._sastoken_provider.stop()
```
BertKleewein, line 121: How about this?
```
try:
  await self._mqtt_client.stop()
finally:
  if self._sastoken_provider:
    await self._sastoken_provider.stop()
```
This way the sastoken provider stops even if there's an exception stopping the client. (Using `try`/`finally` instead of re-raising ensures that the original exception context doesn't get lost.)

v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -0,0 +1,283 @@
...
+            raise ValueError(
+                "Missing authentication - must provide one of 'shared_access_key', 'sastoken_fn' or 'ssl_context'"
+            )
+
+        # Set up SAS auth (if using)
+        generator: Optional[st.SasTokenGenerator]
+        # NOTE: Need to keep a reference to the SasTokenProvider so we can stop it during cleanup
+        self._sastoken_provider: Optional[st.SasTokenProvider]
+        if shared_access_key:
+            uri = _format_sas_uri(hostname=hostname, device_id=device_id, module_id=module_id)
```
BertKleewein, line 64: >hostname=hostname [](http://example.com/codeflow?start=34&length=17)

This is wrong because you're overloading the `hostname` variable to be either the `hostname` or the `gatewayhostname` from the connection string. If connecting to a gateway, the sas token gets built using `hostname` and the TLS socket is opened using `gatewayhostname`

v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -0,0 +1,283 @@
...
+        return "{hostname}/devices/{device_id}/modules/{module_id}".format(
+            hostname=hostname, device_id=device_id, module_id=module_id
+        )
+    else:
+        return "{hostname}/devices/{device_id}".format(hostname=hostname, device_id=device_id)
+
+
+def _default_ssl_context() -> ssl.SSLContext:
+    """Return a default SSLContext"""
+    ssl_context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLS_CLIENT)
```
BertKleewein, line 280: >PROTOCOL_TLS_CLIENT [](http://example.com/codeflow?start=46&length=19)

I think we want to also set `SSLContext.minimum_version` by default. IoTHub should always negotiate to 1.2+, but proxies and other MITMs might not.

v3_async_wip/tests/test_iothub_session.py
```diff
@@ -0,0 +1,1289 @@
...
+
+# Mock out the underlying client in order to not do network operations
+@pytest.fixture(autouse=True)
+def mock_mqtt_iothub_client(mocker):
+    return mocker.patch.object(mqtt, "IoTHubMQTTClient", spec=mqtt.IoTHubMQTTClient).return_value
+
+
+# @pytest.fixture(autouse=True)
+# def mock_http_iothub_client(mocker):
+#     return mocker.patch.object(http, "IoTHubHTTPClient", spec=http.IoTHubHTTPClient).return_value
```
BertKleewein, line 58: nit: dead

v3_async_wip/tests/test_iothub_session.py
```diff
@@ -0,0 +1,1289 @@
...
+# Note that some parametrizations are also defined within the scope of a single test suite if that
+# is the only unit they are relevant to.
+
+
+# Parameters for arguments to the __init__ or factory methods. Represent different types of
+# authentication. Use this parametrization whenever possible on .create() tests.
+# NOTE: Do NOT combine this with the SSL fixtures above. This parametrization contains
+# ssl contexts where necessary
+create_auth_params = [
+    # Provide args in form 'shared_access_key, sastoken_fn, ssl_context'
```
BertKleewein, line 102: ># Provide args in form 'shared_access_key, sastoken_fn, ssl_context' [](http://example.com/codeflow?start=4&length=68)

recommend putting 'shared_access_key, sastoken_fn, ssl_context' into a string variable to use across test definitions,  Also "kwarg_name, kwarg_value"

v3_async_wip/tests/test_iothub_session.py
```diff
@@ -0,0 +1,1289 @@
...
+    pytest.param(
+        None,
+        sastoken_generator_fn,
+        lazy_fixture("custom_ssl_context"),
+        id="User-Provided SAS Token Auth + Custom SSLContext",
+    ),
+    pytest.param(None, None, lazy_fixture("custom_ssl_context"), id="Custom SSLContext Auth"),
+]
+# Just the parameters where SAS auth is used
+create_auth_params_sas = [param for param in create_auth_params if "SAS" in param.id]
```
BertKleewein, line 127: this is slick. I like how you did this.

v3_async_wip/tests/test_iothub_session.py
```diff
@@ -0,0 +1,1289 @@
...
+        # Provide args in the form 'device_id, module_id'
+        pytest.param(FAKE_DEVICE_ID, None, id="Device"),
+        pytest.param(FAKE_DEVICE_ID, FAKE_MODULE_ID, id="Module"),
+    ]
+
+    @pytest.mark.it(
+        "Instantiates and stores a SasTokenProvider that uses symmetric key-based token generation, if `shared_access_key` is provided"
+    )
+    @pytest.mark.parametrize("shared_access_key, sastoken_fn, ssl_context", create_auth_params_sak)
+    @pytest.mark.parametrize("device_id, module_id", create_id_params)
```
BertKleewein, line 169: >"device_id, module_id" [](http://example.com/codeflow?start=29&length=22)

Also recommend putting this into a string var.

v3_async_wip/tests/test_iothub_session.py
```diff
@@ -0,0 +1,1289 @@
...
+        original_ssl_ctx_cls = ssl.SSLContext
+
+        # NOTE: due to implementation of SSLContext, we have to mock, and then unset the mock.
+        # Thus, this side effect has been implemented to do just that.
+        def return_and_reset(*args, **kwargs):
+            ssl.SSLContext = original_ssl_ctx_cls
+            assert kwargs["protocol"] is ssl.PROTOCOL_TLS_CLIENT
+            return my_ssl_context
+
+        mocker.patch.object(ssl, "SSLContext", side_effect=return_and_reset)
```
BertKleewein, line 368: This feels clunky. I've been staring at this for a while and I still have no idea what you're doing here. The comment `due to the implementation of SSLContext` doesn't tell me anything and I think it adds to my confusion.

BertKleewein (approved):
:shipit:

v3_async_wip/v3_async_wip/iothub_session.py
cartertinney: I will double check this again, but I believe this does work - Ryan and I discussed it with the Edge folks while you were out of office.

v3_async_wip/v3_async_wip/iothub_session.py
cartertinney: Good call, this was just me being lazy

v3_async_wip/tests/test_iothub_session.py
cartertinney: It's likely to be back in the next PR, but it's not like it's hard to re-implement, so I'll just remove it now.

v3_async_wip/tests/test_iothub_session.py
cartertinney: I'm not quite clear what you mean here - this list of params is here to be used across test definitions. Could you clarify?

v3_async_wip/tests/test_iothub_session.py
cartertinney: Yeah, I remember thinking that as I wrote it.... I'll improve the explanation in a longer note

v3_async_wip/v3_async_wip/iothub_session.py
cartertinney: Added this and detailed tests

PR #1115, opened 2023-03-31
[V3] Allow connection loss to propagate through APIs and generators

cartertinney:
* Removed SAS renewal logic in the IoTHubMQTTClient.
* Added support for reporting connection drop errors in MQTTClient and IoTHubMQTTClient
* Modified IoTHubSession receiver generators to raise error in the case of connection drop / disconnection
* Modified outgoing data IoTHubSession APIs to raise error in the case of connection drop / disconnection
* Added basic `sastoken_ttl` argument support

NOTE: Auth API adjustments to reflect the lack of SAS renewal will be part of a separate PR

### DIFF
--- a/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
+++ b/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
@@ -239,7 +239,7 @@ class IoTHubMQTTClient:
         await self._mqtt_client.disconnect()
         logger.debug("Disconnect succeeded")
 
-    async def wait_for_connection_drop(self) -> MQTTError:
+    async def report_connection_drop(self) -> MQTTError:
         """Block until the connection is dropped and return the cause"""
         async with self._mqtt_client.disconnected_cond:
             await self._mqtt_client.disconnected_cond.wait_for(
--- a/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
+++ b/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
@@ -362,7 +362,7 @@ class IoTHubMQTTClient:
                 )
             )
             # TODO: should body be logged? Is there useful info there?
-            if response.status != 200:
+            if response.status >= 300:
                 raise IoTHubError(
                     "IoTHub responded to twin patch with a failed status - {}".format(
                         response.status
@@ -430,7 +430,7 @@ class IoTHubMQTTClient:
                 await self._request_ledger.delete_request(request.request_id)
 
         # Interpret response
-        if response.status != 200:
+        if response.status >= 300:
             raise IoTHubError(
                 "IoTHub responded to get twin request with a failed status - {}".format(
                     response.status
--- a/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
+++ b/v3_async_wip/v3_async_wip/iothub_mqtt_client.py
@@ -8,7 +8,7 @@ import asyncio
 import json
 import logging
 import urllib.parse
-from typing import Callable, Optional, AsyncGenerator, TypeVar, cast
+from typing import Callable, Optional, AsyncGenerator, TypeVar
 from .custom_typing import TwinPatch, Twin
 from .iot_exceptions import IoTHubError, IoTHubClientError
 from .mqtt_client import (  # noqa: F401 (Importing directly to re-export)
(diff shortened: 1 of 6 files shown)


v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -234,9 +254,15 @@ async def direct_method_requests(
...
-            yield self._mqtt_client.incoming_direct_method_requests
+            yield self._add_disconnect_interrupt_to_generator(
+                self._mqtt_client.incoming_direct_method_requests
+            )
         finally:
-            await self._mqtt_client.disable_direct_method_request_receive()
+            try:
+                await self._mqtt_client.disable_direct_method_request_receive()
+            except mqtt.MQTTError:
+                # i.e. not connected
```
BertKleewein, line 264: ># i.e. not connected [](http://example.com/codeflow?start=16&length=20)

It might be better to say something like "OK to ignore. We expect MQTTError to be raised here after disconnection." or something like that.

v3_async_wip/v3_async_wip/iothub_session.py
```diff
@@ -245,9 +271,37 @@ async def desired_property_updates(
...
+    ) -> AsyncGenerator[_T, None]:
+        """Wrap a generator in another generator that will either return the next item yielded by
+        the original generator, or raise error in the event of disconnect
+        """
+
+        async def wrapping_generator():
+            while True:
+                new_item_t = asyncio.create_task(generator.__anext__())
+                done, _ = await asyncio.wait(
+                    [new_item_t, self._report_conn_drop_task],
```
BertKleewein, line 295: >_report_conn_drop_task [](http://example.com/codeflow?start=38&length=22)

Does this work if you're cancelling 2 generators? I know that you can't await an already-awaited task, but I don't know how this applies to asyncio.wait.

BertKleewein: If this is a problem, you could use asyncio.Event here (e.g. `done, pending = asyncio.wait([new_item_t,, self.drop_conn_event.wait()], return_when=asyncio.FIRST_COMPLETED)`. 

The hidden trick here is that you have to cancel the `drop_con_event,wait()` task if it's inside the `pending` array.  If you don't do this, you'll leak a task every time this gets called.

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
```diff
@@ -275,6 +239,15 @@ async def disconnect(self) -> None:
         await self._mqtt_client.disconnect()
         logger.debug("Disconnect succeeded")
 
+    async def report_connection_drop(self) -> MQTTError:
```
BertKleewein, line 242: >report_connection_drop [](http://example.com/codeflow?start=14&length=22)

Confusing name. This function does not report a connection drop. It waits for a connection drop.

cartertinney: I wanted to add something about the fact it returns the error. Let me think about some other names

v3_async_wip/v3_async_wip/iothub_session.py
cartertinney: I think it does, but I'll look into it and double check. I had considered doing this (not with an Event, but with a Future), but my implementation seemed simpler.

v3_async_wip/v3_async_wip/mqtt_client.py
```diff
@@ -386,6 +390,14 @@ def is_connected(self) -> bool:
         """
         return self._connected
 
+    def previous_disconnection_cause(self) -> Optional[MQTTError]:
+        """
+        Returns an MQTTError from the previous disconnection if it was unexpected as a result of
+            a connection drop.
+        Returns None if the previous disconnection attempt was intentional
```
BertKleewein, line 397: > Returns None if the previous disconnection attempt was intentional [](http://example.com/codeflow?start=7&length=67)

It looks like generators don't exit if the disconnect was intentional. Don't we want them to exit?

v3_async_wip/v3_async_wip/iothub_mqtt_client.py
BertKleewein: maybe `wait_for_connection_drop` since the waiting is the important part.

v3_async_wip/v3_async_wip/iothub_session.py
cartertinney: Update: This does work fine for cancelling multiple generators

BertKleewein (approved):
:shipit:

PR #1118, opened 2023-04-05
[V3] Revised Migration Guides + README

cartertinney:
* Revised IoTHub Migration Guide
* Revised Provisioning Migration Guide (even though ProvisioningSession is still a work in progress
* Revised README to reflect v3


migration_guide_iothub.md
```diff
@@ -1,237 +1,222 @@
...
+    async with IoTHubSession.from_connection_string("<Your Connection String>") as session:
+        # <do things>
 ```
 
-Note that many people using V2 may already have been doing manual connects, as for some time, this has been our recommended practice.
-
-Note also that this change does *not* affect automatic reconnection attempts in the case of network failure. Once the manual connect has been successful, the client will (under default settings) still attempt to retain that connected state as it did in V2.
-
+When the context manager is entered, a connection will be established before running the block of code inside the context manager. After the block is done executing, a disconnection will occur upon context manager exit. You can consider that all code within the block is written with the expectation of a connection, as the context manager
+represents a connection to the IoT Hub.
```
BertKleewein, line 35: Suggestion: Since markdown concatenates consecutive lines, put each sentence on it's own line. It renders the same, but it's easier to review and easier to track changes.

BertKleewein (approved):
:shipit:

PR #1125, opened 2023-04-13
Notates sample being used in tutorial

lev-i:
Tutorial located here:
https://learn.microsoft.com/en-us/azure/iot-develop/quickstart-send-telemetry-iot-hub?pivots=programming-language-python

Which has an open PR to list one of the new samples being used. This sample will also likely be updated, so we want to keep track of what's being used where.

### DIFF
--- a/samples/iothub_simple_telemetry.py
+++ b/samples/iothub_simple_telemetry.py
@@ -3,7 +3,12 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-"""This sample demonstrates a simple recurring telemetry using an IoTHubSession"""
+"""
+This sample demonstrates a simple recurring telemetry using an IoTHubSession
+
+It's set to be used in the following MS Learn Tutorial:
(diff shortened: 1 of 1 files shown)


timlt:
@lev-i we updated that doc and merged your PR with the updates.  

A user is now reporting an issue that the sample no longer runs. 

Issue:  https://github.com/MicrosoftDocs/azure-docs/issues/108290

TLDR; version:  I reproduced his issue and his suggestion of the fix is correct.  Basically:

> Root cause: incorrect pip command mentioned in the web page:
> pip3 install azure-iot-device
> 
> Correct command is:
> pip install azure-iot-device==3.0.0b2
> 
> Reference:
> https://github.com/Azure/azure-iot-sdk-python#installing-the-library

Should we go ahead and update the "pip3" command in the doc to the inline example he lists that works?   I wasn't clear if we wanted to hardcode that version reference though.

lev-i:
@timlt
Yes that's correct, the command we use now is:
`pip install azure-iot-device==3.0.0b2`

Thanks!

olivakar (approved):

PR #1127, opened 2023-04-18
Python E2E : DPS end to end updated tests

olivakar:
I had to revert back to old `sas_token` and old `connection_string` mechanisms which existed even before the last V2 common package usages as replacing with current V3 signing mechanism and connection string and sas token is not working.

Some weird import adjustments had to be done. I am not exactly sure why they stopped working in the pipeline.

Also deleted the sync test files.


tests/e2e/provisioning_e2e/connection_string.py
```diff
@@ -1,22 +1,91 @@
...
+    # This logic could be expanded to return the category of ConnectionString
+    if host_name and device_id and shared_access_key:
+        pass
+    elif host_name and shared_access_key and shared_access_key_name:
+        pass
+    else:
+        raise ValueError("Invalid Connection String - Incomplete")
+
+
+class ConnectionString(object):
```
BertKleewein, line 61: I know that they're not part of our public API, but couldn't you use the sastoken.py and connection_string.py from `azure-iot-device/azure/iot/device`?

tests/e2e/provisioning_e2e/tests/test_async_symmetric_enrollments.py
```diff
@@ -103,29 +106,27 @@ def assert_device_provisioned(device_id, registration_result):
     :param device_id: The device id
     :param registration_result: The registration result
     """
-    assert registration_result.status == "assigned"
-    assert registration_result.registration_state.device_id == device_id
-    assert registration_result.registration_state.assigned_hub == linked_iot_hub
+    print(registration_result)
+    assert registration_result["status"] == "assigned"
+    assert registration_result["registrationState"]["deviceId"] == device_id
+    assert registration_result["registrationState"]["assignedHub"] == linked_iot_hub
```
BertKleewein, line 113: Why are we changing `registration_result`, etc, from objects to dictionaries?  This seems like a step backwards.

BertKleewein (approved):
:shipit:

tests/e2e/provisioning_e2e/connection_string.py
olivakar: i tried using them actually. it did not work - not with direct replacements.

tests/e2e/provisioning_e2e/tests/test_async_symmetric_enrollments.py
olivakar: that is because of this. we don't have translation from what the service gives us to our special object..if we want to keep things as it is then we ca define RegistrationResult on our own...and translate things like "deviceId" to "device_id"

PR #1139, opened 2023-05-05
Error story enhancements

cartertinney:
* Added `MQTTConnectionDroppedError` for when an established connection is lost. This allows the user to differentiate between reconnect scenarios (connection was lost, `MQTTConnectionDroppedError` was raised) and retry scenarios (operation failed, `MQTTError` was raised)
* Added `SessionError` for errors coming from the scope of a Session object
* Unified errors in a single global module
* Updated tests and samples

NOTE: retry scenarios are still not included in samples, pending doc/sample discussions


azure-iot-device/azure/iot/device/iothub_session.py
```diff
@@ -20,6 +21,28 @@
 # TODO: add tests for sastoken_ttl argument once we settle on a SAS strategy
 
 
+def _requires_connection(f):
+    """Decorator to indicate a method requires the Session to already be connected."""
+
+    def check_connection_wrapper(*args, **kwargs):
```
BertKleewein, line 27: >f check_connection_wrapper(*args, **kwargs): [](http://example.com/codeflow?start=6&length=44)

Please add `functools.wraps`

cartertinney: Good idea. Missed this.

PR #1140, opened 2023-05-18
Support for user-supplied SAS Tokens

cartertinney:
Done:
* Support for user-supplied SAS Tokens now available via the `sastoken` parameter when constructing an `IoTHubSession` or a `ProvisioningSession`
* New `.update_sastoken()` method added to both `IoTHubSession` and `ProvisioningSession` to facilitate user-supplied SAS token replacement
* Removed the provisional `sastoken_fn` callback parameter from constructor of both `IoTHubSession` and `ProvisioningSession` as it is no longer necessary
* Added support for connection strings that contain a SAS token via the `.from_connection_string` factory method
* Introduced a new exception, `CredentialError`, which indicates an expired SAS token
* Removed `ExternalSasTokenGenerator` class as it is no longer necessary because `sastoken_fn` is no longer supported
* Renamed `InternalSasTokenGenerator` class to simply `SasTokenGenerator`
* Moved SAS token TTL configuration to the `.generate_sastoken()` method of `SasTokenGenerator` instead of instantiation
* Removed usage of `SasTokenProvider` from underlying MQTT and HTTP clients, however the implementation still is defined (for now)
* Updated docstrings
* Added unit tests for connection strings containing `SharedAccessSignature` and `GatewayHostName`
* Added previously missing unit tests for `sastoken_ttl`
* Enabled support for SAS auth in E2E tests
* Migration guide changes

NOTE:
Updating a SAS Token on an existing Session, while supported via API, is currently broken when used in some patterns due to a more general connection bug. See Issue #1146 for more details. As such, no sample is currently added for updating a SAS Token.


azure-iot-device/azure/iot/device/connection_string.py
```diff
@@ -93,18 +91,25 @@ def _parse_connection_string(connection_string):
...
+    shared_access_signature = d.get(SHARED_ACCESS_SIGNATURE)
     device_id = d.get(DEVICE_ID)
     x509 = d.get(X509)
 
-    if shared_access_key and x509 and x509.lower() == "true":
+    # Validate only one type of auth included
+    auth_count = 0
+    if shared_access_key:
+        auth_count += 1
+    if x509 and x509.lower() == "true":
```
BertKleewein, line 103: Do we have to worry about x509 being truthy but not "true"

azure-iot-device/azure/iot/device/iothub_session.py
```diff
@@ -84,47 +83,49 @@ def __init__(
...
+            # If using SasToken generation, we cannot generate during this __init__
+            # because .generate_sastoken() is a coroutine. The token will be generated and then
+            # set on the underlying client upon context manager entry
             uri = _format_sas_uri(hostname=hostname, device_id=device_id, module_id=module_id)
             signing_mechanism = sm.SymmetricKeySigningMechanism(shared_access_key)
-            generator = st.InternalSasTokenGenerator(
-                signing_mechanism=signing_mechanism, uri=uri, ttl=sastoken_ttl
+            self._sastoken_generator = st.SasTokenGenerator(
+                signing_mechanism=signing_mechanism, uri=uri
             )
```
BertKleewein, line 115: We still need sastoken_ttl, even if we don't have automatic renewal. The token still needs to expire eventually and we still want the customer to be able to se the expiration interval.

dev_utils/dev_utils/iothub_mqtt_helper.py
```diff
@@ -30,14 +30,14 @@ async def create_client_config(cs_str):
...
-def _create_sastoken_generator(connection_string, ttl=3600):
+def _create_sastoken_generator(connection_string):
     uri = _form_sas_uri(
         hostname=connection_string[cs.HOST_NAME],
         device_id=connection_string[cs.DEVICE_ID],
         module_id=connection_string.get(cs.MODULE_ID),
     )
     signing_mechanism = sm.SymmetricKeySigningMechanism(key=connection_string[cs.SHARED_ACCESS_KEY])
-    sastoken_generator = st.InternalSasTokenGenerator(signing_mechanism, uri, ttl)
+    sastoken_generator = st.SasTokenGenerator(signing_mechanism, uri)
```
BertKleewein, line 40: Maybe add expiration interval back just like I suggested in iothub_session.py.  I think iothub_mqtt_helper is for test code only, so maybe not necessary here.

azure-iot-device/azure/iot/device/connection_string.py
cartertinney: It shouldn't happen, but it's possible by definition of the boolean keyword.

azure-iot-device/azure/iot/device/iothub_session.py
cartertinney: It's added upon each generation

PR #1163, opened 2024-01-10
Typings

olivakar:


azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -232,12 +238,12 @@ def _replace_user_supplied_sastoken(self, sastoken_str):
...
 
     @abc.abstractmethod
-    def _generic_receive_handler_setter(self, handler_name, feature_name, new_handler):
+    def _generic_receive_handler_setter(self, handler_name: str, feature_name: str, new_handler: Optional[Callable[[], Any]]) -> None:
         # Will be implemented differently in child classes, but define here for static analysis
         pass
 
     @classmethod
-    def create_from_connection_string(cls, connection_string, **kwargs):
+    def create_from_connection_string(cls, connection_string: str, **kwargs: Dict[str, Any]) -> Self:
```
cartertinney, line 246: Don't annotate the kwargs. Here and everywhere.

azure-iot-device/azure/iot/device/iothub/abstract_clients.py
```diff
@@ -393,66 +399,66 @@ def create_from_sastoken(cls, sastoken, **kwargs):
...
-    def connected(self):
+    def connected(self) -> bool:
         """
         Read-only property to indicate if the transport is connected or not.
         """
         return self._mqtt_pipeline.connected
 
     @property
-    def on_connection_state_change(self):
+    def on_connection_state_change(self) -> Callable[[Any], Any]:
```
cartertinney, line 449: I'm somewhat surprised this passes static analysis, although I suppose that's due to only annotating the surface.

The issue is that this is too broad. It says you can set any callable that takes any number of arguments and returns any type of value. This is not true, and following this annotation would produce errors.

This should probably be FunctionOrCoroutine[[None], None], however, I'm not quite sure if that is valid for all versions of Python - it used to not be due to a bug, but I believe that has been fixed since I reported it? If it hasn't been, I can live with Callable[[Any], Any], but it's problematic.

Here and everywhere obviously (although the specific type annotations would change with handler definition)

cartertinney (approved):

PR #1176, opened 2024-03-11
bugfix: sample: disconnect when all the messages are sent

ks6088ts:
## Description of the problem

Current implementation of the sync version of `provision_symmetric_key.py` calls disconnect API repeatedly, it seems weird.

I think it would be correct just as the async version [samples/async-hub-scenarios/provision_symmetric_key.py](https://github.com/Azure/azure-iot-sdk-python/blob/main/samples/async-hub-scenarios/provision_symmetric_key.py#L56) does.


ks6088ts:
I'm not sure why the CI fails...
@olivakar Could you take a look at it ?

cartertinney:
> I'm not sure why the CI fails... Mentioning @olivakar , who is checked in the latest update. Could you take a look at it ?

Apologies for the late reply - there was an infrastructure issue that broke CI. It has now been addressed, we will take in this fix.

cartertinney (approved):

Issue #1187, opened 2024-06-19
[feature request] support requests >2.32 via maintained requests-unixsocket fork?


robin-wayve:
**Is your feature request related to a problem? Please describe.**
We would like to upgrade `requests` for https://github.com/advisories/GHSA-9wx4-h78v-vm56

**Describe the solution you'd like**
requests-unixsocket has been forked to address the previous blocker: https://github.com/msabramo/requests-unixsocket/issues/73#issuecomment-2125848213

robin-wayve:
previously discussed @ #1182

cartertinney:
I'll try and get this updated in the next week or so. We have to make sure that the fork of `requests-unixsocket` is safe.

cartertinney:
Version 2.14.0 has been released containing this fix

Issue #1180, opened 2024-04-24
Allow higher version of urllib3


mdczaplicki:
Please allow higher version of `urllib3`. The one required in this package is quite old.

cartertinney:
I'll take a look next time I get a chance to see what options are viable for us. We had heavily restricted the versioning for urllib3 due to a compatibility issue - perhaps with time having passed we can loosen the restriction.

Eramismus:
Are there any updates on relaxing the urllib3 version? Is there still a compatibility issue?

cartertinney:
@Eramismus

Unfortunately, right now it looks like we will not be lifting this version restriction.

cartertinney:
`azure-iot-device` version 2.14.0 has been released containing this fix.

PR #1189, opened 2024-07-09
Updated dependencies to fix breaking changes in requests

cartertinney:
* `requests` had critical security fixes that required updating minimum version
* This fix broke the `requests-unixsocket` dependency which used monkey-patching
* `requests-unixsocket` is abandoned, so switched to use the fork `requests-unixsocket2`
* Also updated `urllib3` dependency to be more flexible.
* Added support for Python 3.11 and 3.12

### DIFF
--- a/tests/unit/iothub/aio/test_async_clients.py
+++ b/tests/unit/iothub/aio/test_async_clients.py
@@ -9,6 +9,7 @@ import pytest
 import asyncio
 import time
 import urllib
+import sys
 from azure.iot.device import exceptions as client_exceptions
 from azure.iot.device.common.auth import sastoken as st
 from azure.iot.device.iothub.aio import IoTHubDeviceClient, IoTHubModuleClient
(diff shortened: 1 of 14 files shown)


ryanwinter (approved):

robin-wayve:
resolves #1187?

cartertinney:
> resolves #1187?

Yes. Also #1180

tests/unit/iothub/aio/test_async_clients.py
```diff
@@ -693,6 +694,10 @@ async def test_raises_error_when_message_size_greater_than_256(self, client, mqt
         assert "256 KB" in e_info.value.args[0]
         assert mqtt_pipeline.send_message.call_count == 0
 
+    @pytest.mark.skipif(
```
ryanwinter, line 697: What are these python 3.12 issues? Does this impact the customer at all? Looks like if they have large messages they might not get an error?

vsts/python-e2e.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
   strategy:
     matrix:
       py310_mqtt:              { pv: '3.10', transport: 'mqtt', consumer_group: 'e2e-consumer-group-3' }
```
ryanwinter, line 42: py312 here?

ryanwinter (changes requested):

tests/unit/iothub/aio/test_async_clients.py
cartertinney: Yeah - but honestly, this max size never actually worked in practice, so it's not an issue.

vsts/python-e2e.yaml
cartertinney: Good catch

ryanwinter (approved):

Issue #1192, opened 2024-09-03
'NoneType' object has no attribute 'on_mqtt_connected_handler'


Siecje:
# Context

- **OS and version used:** Linux version 6.1.21-v8+
- **Python version:** 3.9.2
- **pip version:** 24.2
- **list of installed packages:** 
```
Package             Version
------------------- -----------
async-timeout       4.0.3
atcom               0.4.3
azure-core          1.30.2
azure-iot-device    2.13.0
azure-storage-blob  12.22.0
bcrypt              4.1.3
bleak               0.20.1
certifi             2024.2.2
cffi                1.16.0
charset-normalizer  3.3.2
click               8.1.7
cryptography        3.2.1
dbus-fast           1.95.2
deprecation         2.1.0
exceptiongroup      1.2.2
idna                3.7
iniconfig           2.0.0
iotc                1.1.3
isodate             0.6.1
janus               1.0.0
numpy               1.26.4
packaging           24.0
paho-mqtt           1.6.1
pandas              2.2.2
paramiko            2.7.2
pip                 24.2
pkg_resources       0.0.0
pluggy              1.5.0
psutil              5.9.8
pycparser           2.22
PyNaCl              1.5.0
pyserial            3.5
PySocks             1.7.1
pytest              8.3.2
python-dateutil     2.9.0.post0
pytz                2024.1
PyYAML              6.0.1
pyzmq               25.1.2
requests            2.31.0
requests-unixsocket 0.3.0
RPi.GPIO            0.7.1
sentry-sdk          1.44.1
setuptools          72.1.0
six                 1.16.0
tomli               2.0.1
typing_extensions   4.11.0
tzdata              2024.1
urllib3             1.26.18
wheel               0.44.0
```
- **cloned repo:** <VERSION> ( **If** you are using a cloned sdk repository, in a command prompt: git describe )

## Description of the issue

I get Sentry errors for `'NoneType' object has no attribute 'on_mqtt_connected_handler'`. The traceback doesn't include my code just the code in the package. I'm not creating a thread in my code.

```
  File "threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "paho/mqtt/client.py", line 3591, in _thread_main
    self.loop_forever(retry_first_connection=True)
  File "paho/mqtt/client.py", line 1756, in loop_forever
    rc = self._loop(timeout)
  File "paho/mqtt/client.py", line 1164, in _loop
    rc = self.loop_read()
[...]

PR #1193, opened 2024-09-03
Handle when transport has been garbage collected.

Siecje:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-python/.github/CONTRIBUTING.md).

## Reference/Link to the issue solved with this PR (if any)

#1192

## Description of the problem

`this` can be None if the object has already been garbage collected.

Check if `this` is None and log as in `on_disconnect`.


cartertinney (approved):

Siecje:
Cherry pick the commit into another PR or just make the changes yourself.

cartertinney:
@Siecje  Yeah, we're just bringing some repository infrastructure back online, so once that's sorted I'll look at opening my own PR on this.

However, since you submitted it, can you explain more about the scenario where this happens? My understanding had been that this wasn't something that could happen, which is why it wasn't handled (unlike in `on_disconnect` where it was handled). Do you have a scenario that can reproduce the issue?

I realize it's been a year, so you may not have that info on hand, but if you have anything to go off of, it'd be appreciated.

cartertinney:
Superseded by #1240

PR #1224, opened 2026-04-14
Add a sample that demonstrates TLS 1.3 support

timtay-microsoft:
Please note that this feature is not actually available to most users yet as the service side is still in preview.

### DIFF
--- a/samples/how-to-guides/connect_with_tls13.py
+++ b/samples/how-to-guides/connect_with_tls13.py
@@ -4,12 +4,8 @@
 # license information.
 # --------------------------------------------------------------------------
 
-import random
-import asyncio
-
-from azure.iot.device.aio import IoTHubDeviceClient
-from azure.iot.device import exceptions
 import os
+from azure.iot.device import IoTHubDeviceClient, Message
 
 # The device connection string to authenticate the device with your IoT hub.
 CONNECTION_STRING = os.getenv("IOTHUB_DEVICE_CONNECTION_STRING")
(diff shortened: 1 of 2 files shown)


samples/how-to-guides/connect_with_tls13.py
```diff
@@ -0,0 +1,172 @@
...
+import os
+
+# The device connection string to authenticate the device with your IoT hub.
+CONNECTION_STRING = os.getenv("IOTHUB_DEVICE_CONNECTION_STRING")
+
+if CONNECTION_STRING is None:
+    print(
+        "Must set environment variable 'IOTHUB_DEVICE_CONNECTION_STRING' in the format '<hub name>.device.azure-devices.<dns suffix>' in order to use TLS 1.3"
+    )
+    raise Exception
```
cartertinney: Never `raise Exception` in Python. While there's nothing risky about it in this specific context because it's an application, it's generally dangerous as it gets tricky to catch due to being the parent class of all exceptions, and is an antipattern.

The correct thing to do here I think is:

```python
if ".device.azure-devices." not in CONNECTION_STRING:
    # classic connection strings that look like '<hub-name>.azure-devices.<dns suffix>' only support up to TLS 1.2
    raise ValueError("Device connection string must match the format '<hub name>.device.azure-devices.<dns suffix>' in order to use TLS 1.3")
```

You can forgo the validation of not being None entirely.

cartertinney (changes requested):

cartertinney (approved):

PR #1231, opened 2026-06-11
release: 2.15.0

cartertinney:


cervm:
@cartertinney Great to see a new version in the works! Do you have an estimate for when this version might be released as stable? We’re especially looking forward to the upgrade to Paho MQTT 2.x, as it would unblock other tasks for us

cartertinney:
@cervm As soon as I can, I'll aim to get this out by September, our apologies for the delay

PR #1241, opened 2026-08-21
fix: validate encoded telemetry message size

cartertinney:
## Summary

- replace `sys.getsizeof()` telemetry validation with deterministic IoT Hub message accounting: encoded body bytes, outgoing system-property values, and custom-property names and values
- mirror Paho's payload conversion (`str` as UTF-8, `int`/`float` as ASCII, `None` as empty, and `bytes`/`bytearray` as-is) while excluding MQTT topic and packet overhead
- share system/custom property normalization with MQTT topic encoding and include the module output system property before validation
- replace Python-layout-dependent magic numbers and Python 3.12+ skips with exact below/equal/above 262144-byte tests for Unicode, bytes, custom properties, and system properties across sync/async device/module sends

## Behavior and compatibility

IoT Hub defines message size independently of the transport as body bytes plus system-property values plus application-property names and values. The old implementation measured CPython object allocation instead, so identical messages could validate differently across Python versions.

This change intentionally affects messages near the 256-KB boundary:

- exactly 262144 logical message bytes remain accepted; larger messages are rejected
- UTF-8 strings and raw byte payloads are measured as actually published by Paho
- custom-property keys now contribute to size, while Python object headers no longer do
- outgoing system-property values, including an IoT Edge module output name, contribute to size
- percent-encoding, the base MQTT topic, and MQTT framing do not contribute because they are protocol overhead rather than IoT Hub message content
- unsupported caller-provided payload types now raise `TypeError` directly before a pipeline operation starts, rather than surfacing as `ClientError` after Paho rejects them

The MQTT topic encoding itself is unchanged; sizing and encoding now consume the same normalized property values to prevent drift.

## Validation

- focused message/topic/sync/async send tests passed on Python 3.10, 3.11, 3.12, 3.13, and 3.14
- direct `TypeError` and size-validation tests: **176 passed** on the current merged PR head
- full unit suite on the current merged PR head: **5487 passed** on Python 3.14
- repository-wide Ruff, Python compilation, and `git diff --check` passed
- the previous E2E regression was reproduced and its sync/async assertions now cover the direct `TypeError` contract


azure-iot-device/azure/iot/device/iothub/pipeline/mqtt_topic_iothub.py
```diff
@@ -5,9 +5,13 @@
 # --------------------------------------------------------------------------
 
 import logging
-from datetime import date
 import urllib
 
+from azure.iot.device.iothub.models.message import (
+    _get_custom_properties,
```
ewertons, line 11: Excellent!

ewertons (approved):

